PR #12, opened 2017-08-24
Nturinski/ci fix

nturinski:
Fixes repository and license warnings
Fixes npm install 401 error by changing Linux dist from 'trusty' to 'precise'


.travis.yml
```diff
@@ -13,4 +13,6 @@ notifications:
   email:
     on_success: never
     on_failure: always
-    
\ No newline at end of file
+    
+
+dist: precise
```
EricJizbaMSFT: Since it looks like removing 'package-lock.json' fixed the issue - can you revert this change to .travis.yml?

PR #18, opened 2017-08-29
App node expansion

nturinski:
Will not build on TravisCI due to TS warnings, but runs as an extension.  Does anybody know the official way to acquire the accessToken?  I don't think the way I am currently doing it is the preferred method.

### DIFF
--- a/src/appServiceNodes.ts
+++ b/src/appServiceNodes.ts
@@ -7,6 +7,7 @@ import { TreeDataProvider, TreeItem, TreeItemCollapsibleState, EventEmitter, Eve
 import { AzureAccountWrapper } from './azureAccountWrapper';
 import { SubscriptionClient, SubscriptionModels } from 'azure-arm-resource';
 import WebSiteManagementClient = require('azure-arm-website');
+import request = require('request-promise');
 import * as WebSiteModels from '../node_modules/azure-arm-website/lib/models';
 import * as path from 'path';
 import * as opn from 'opn';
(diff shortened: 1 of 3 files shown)


src/appServiceNodes.ts
```diff
@@ -119,6 +134,54 @@ export class AppServiceNode extends NodeBase {
...
 }
 
+export class FilesNode extends NodeBase {
+    constructor(readonly label: string, readonly href: string) {
+        super(label);
+    }
+
+    getTreeItem(): TreeItem {
+        return {
+            label: `📁` + this.label,
```
whiskeyjay: Let's get a real icon for it.

src/appServiceNodes.ts
```diff
@@ -88,6 +89,20 @@ export class AppServiceNode extends NodeBase {
         }
     }
 
+    async getChildren(azureAccount: AzureAccountWrapper): Promise<NodeBase[]> {
+        if (azureAccount.signInStatus !== 'LoggedIn') {
```
whiskeyjay: Will the status ever be something other than 'LoggedIn' here, since the parent node also requires sign-in?

src/azureAccountWrapper.ts
```diff
@@ -102,4 +102,15 @@ export class AzureAccountWrapper {
     registerFiltersChangedListener(listener: (e: void) => any, thisArg: any): Disposable {
         return this.accountApi.onFiltersChanged(listener, thisArg, this.extensionConext.subscriptions);
     }
+
+    getAccessToken() {
+        if (this.accountApi.sessions.length) {
+            let currentSession = this.accountApi.sessions[0];
+            if (currentSession.credentials.tokenCache._entries.length) {
+                return currentSession.credentials.tokenCache._entries[0].accessToken;
```
whiskeyjay: Let's request the tokens to be exposed by the account extension.

src/appServiceNodes.ts
nturinski: Can signInStatus change after logging in? Maybe during a disconnect or something?  Is that something that we have to worry about?

PR #19, opened 2017-08-30
Prompt to sign-in if create command executed without an account

whiskeyjay:
Also:
Optimized the loading operations in resource group/location step, making two requests run parallel.
Added a QuickPickItemWithData interface to make it easier to get back the data user picked.

### DIFF
--- a/src/azureAccountWrapper.ts
+++ b/src/azureAccountWrapper.ts
@@ -22,7 +22,7 @@ export class AzureAccountWrapper {
     }
 
     getAzureSessions(): AzureSession[] {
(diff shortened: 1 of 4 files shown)


src/azureAccountWrapper.ts
```diff
@@ -22,7 +22,7 @@ export class AzureAccountWrapper {
     }
 
     getAzureSessions(): AzureSession[] {
-        const status = this.accountApi.status;
+        const status = this.signInStatus;
```
nturinski: I think that this was supposed to be a signInStatus() function invocation.

Issue #14, opened 2017-08-28
Zip Deploy


whiskeyjay:

fiveisprime:
Ran into a couple of issues while attempting to deploy (running on Windows).

- [ ] kudu-api isn't included in package.json
- [ ] The archive doesn't appear to have been created, got an ENOENT on package.zip

whiskeyjay:
Solved missing kudu-api reference.
Archive not created because I am still working on it.
As for the bad state issue, deployment slots shall be able to help, something we should support soon.

whiskeyjay:
https://github.com/Microsoft/vscode-azureappservice/pull/21

Issue #3, opened 2017-08-19
Show "Deployment Slots", "Files", "Log Files" and "Web Jobs" under Web Apps nodes.


whiskeyjay:

nturinski:
Done pending adding full story to Files/Log FIles

Issue #2, opened 2017-08-16
Create New App Service


whiskeyjay:
Adds a new command -
Azure App Service: Create

The command will prompt for a bunch of inputs to fill up the necessary parameters for creating a new Web App.

PR #21, opened 2017-09-02
Zip deploy Node.js project to Web App

whiskeyjay:
Issue https://github.com/Microsoft/vscode-azureappservice/issues/14
There are three ways to initiate the deploy process:

1. Right click on a web app node:
image
2. Right click on a workspace folder:
image
3. Create your own Zip package (currently must be named package.zip) put it in your working folder and right click on it:
image

Depending on the context of the command, you will be prompted different things, for example, method #1 will prompt you for the workspace folder (if more than one; if only one then no prompt); #2 and #3 will prompt for subscription and web app, it will chain in the "Create New Web App" wizard steps if the user chose to deploy to a new web app.

The deployment will stop the web app, delete all web site contents (the wwwroot folder) and upload the zip file to be extracted into a new wwwroot folder. In the end the web app will be started.


src/webAppZipPublisher.ts
```diff
@@ -0,0 +1,307 @@
...
+
+class SubscriptionStep extends SubscriptionStepBase {
+    constructor(wizard: WizardBase,
+        azureAccount: AzureAccountWrapper,
+        subscription?: SubscriptionModels.Subscription) {
+        super(wizard, 'Select target Subscription', azureAccount, subscription);
+    }
+
+    async prompt(): Promise<void> {
+        if (!!this.subscription) {
```
nturinski: If there's no subscription, should it prompt the user to login?

src/webAppZipPublisher.ts
```diff
@@ -0,0 +1,307 @@
...
+        this.steps.push(new SubscriptionStep(this, azureAccount, subscription));
+        this.steps.push(new WebAppStep(this, azureAccount, site));
+        this.steps.push(new DeployStep(this, azureAccount));
+    }
+
+    protected onExecuteError(step: WizardStep, stepIndex: number, error: Error) {
+        if (error instanceof UserCancelledError) {
+            return;
+        }
+        this.writeline(`Deployment failed - ${error.message}`);
```
nturinski: This error comes up as Deployment failed - 404: [object Object] for me

src/kuduClient.ts
```diff
@@ -0,0 +1,113 @@
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
whiskeyjay: My reasons for this:

- The kudu-api package doesn't have a TypeScript type def, it is easy to make a mistake calling it directly.
- The package also doesn't check for HTTP response code, caller is responsible for doing the check.
- Since there are more than one places calling this package (explorer, deployment), it benefits the callers if there is a typed wrapper doing some common house keeping tasks (ie check response code) for them.

src/webAppCreator.ts
```diff
@@ -78,57 +83,26 @@ class SubscriptionBasedWizardStep extends WizardStep {
...
-                data: s
-            };
-
-            quickPickItems.push(item);
-        });
+        if (!!this.subscription) {
+            return;
+        }
 
+        const quickPickItems = await this.getSubscriptionsAsQuickPickItems();
```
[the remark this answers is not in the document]
whiskeyjay: Yes, if the user cancels by clicking away or ESC, the underlying method will throw an exception which will be caught by the Wizard and then terminate gracefully.

src/webAppZipPublisher.ts
whiskeyjay: This line is to check if a subscription has been pre-set (for example, if the wizard was launched from the explorer, which has the context of subscription), if there is a pre-set subscription, then skip the prompt.

Issue #41, opened 2017-09-14
use variable in bash script


chrisdias:
the bash script that gets created has a hard coded `cdias-appname-site.git` in the comments:

``` bash
# git remote add azure "https://$USERNAME@$SITENAME.scm.azurewebsites.net/cdias-appname-site.git"
```

we should be able to use $SITENAME instead:

``` bash
# git remote add azure "https://$USERNAME@$SITENAME.scm.azurewebsites.net/$SITENAME.git"
```

chrisdias:
and i realize this is my fault :)

PR #47, opened 2017-09-15
Fix #40 & #41

### LINKED ISSUE #40: give bash script a file name
after creation of a site i get a nice bash script, but it is named `untitled-1`. 

while the mime type is set correctly (bash script), i suggest giving the file a name such as `deploy.sh`, much in the same way as the cosmosdb extension creates a `results.json` file.

whiskeyjay:
#40 - If a workspace is available, give the generated script a filename. If no workspace is open, keep using untitled because we don't have a folder to put the script file in.

#41 - Replace a hard coded value with a variable.

### DIFF
--- a/src/webAppCreator.ts
+++ b/src/webAppCreator.ts
@@ -532,6 +532,8 @@ class ShellScriptStep extends WebAppCreatorStepBase {
                 if (!vscode.workspace.textDocuments.find(doc => doc.uri.fsPath === uri.fsPath) && !fs.existsSync(uri.fsPath)) {
                     uri = uri.with({ scheme: 'untitled' });
                     break;
+                } else {
+                    uri = null;
                 }
                 count++;
             }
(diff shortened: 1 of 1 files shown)


src/webAppCreator.ts
```diff
@@ -519,10 +521,30 @@ class ShellScriptStep extends WebAppCreatorStepBase {
...
+                count++;
+            }
+        }
+
+        if (uri) {
+            const doc = await vscode.workspace.openTextDocument(uri);
+            const editor = await vscode.window.showTextDocument(doc);
+            await editor.edit(editorBuilder => editorBuilder.insert(new vscode.Position(0,0), script));
+        } else {
+            const doc = await vscode.workspace.openTextDocument({ content: script, language: 'shellscript' });
```
EricJizbaMSFT: @PrashanthCorp can you do the same thing for https://github.com/Microsoft/vscode-cosmosdb/issues/1

The language is 'mongo'. The current behavior in the cosmos repo throws an error when I try to save, but this works

Issue #56, opened 2017-09-15
Refresh on a node refreshes the entire explorer


EricJizbaMSFT:
Repro steps:
1. Expand a web app
1. Click Refresh on 'Deployment Slots'

Expected: Just the deployment slots node is refreshed
Actual: The entire explorer is refreshed

whiskeyjay:
Fixed with https://github.com/Microsoft/vscode-azureappservice/pull/58

PR #58, opened 2017-09-15
Fix refresh node bug #56

nturinski:
Add ternary statement to refresh specific node if node context exists
otherwise refres entire tree

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -33,7 +33,7 @@ export function activate(context: vscode.ExtensionContext) {
 
     context.subscriptions.push(vscode.window.registerTreeDataProvider('azureAppService', appServiceDataProvider));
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -33,7 +33,7 @@ export function activate(context: vscode.ExtensionContext) {
 
     context.subscriptions.push(vscode.window.registerTreeDataProvider('azureAppService', appServiceDataProvider));
 
-    initCommand(context, 'appService.Refresh', () => appServiceDataProvider.refresh());
+    initCommand(context, 'appService.Refresh', (node?: NodeBase) => node ? appServiceDataProvider.refresh(node) : appServiceDataProvider.refresh());
```
whiskeyjay: Is it necessary to do the null/undefined check here? Since the node parameter of the refresh method is also optional.

Issue #24, opened 2017-09-07
"Create new Web App" context menu item for subscriptions in Explorer


whiskeyjay:
Right clicking on a subscription should have a context menu item "Create new Web App", and the "create wizard" will start from prompting resource groups instead of subscriptions because of the context.

PR #72, opened 2017-09-20
Able to create apps from subscription node #24

nturinski:
Adds Create a Web App to SubscriptionNode context menu and then
refreshes only that node (rather than collapsing the entire tree)

#24


src/extension.ts
```diff
@@ -44,6 +44,16 @@ export function activate(context: vscode.ExtensionContext) {
             node.openInPortal();
         }
     });
+    initAsyncCommand(context, 'subscription.CreateWebApp', async (node: SubscriptionNode) => {
+        if (node && node.openInPortal) {
+            const wizard = new WebAppCreator(outputChannel, azureAccount, node.subscription);
+            const result = await wizard.run();
+            
+            if (result.status === 'Completed') {
+                vscode.commands.executeCommand('appService.Refresh', node);
```
EricJizbaMSFT: I know the code was like this before your PR, but since you're in the same area, can you change it to `appServiceDataProvider.refresh(node)`?

I don't like the fact that `executeCommand` relies on a hard-coded string (`appService.Refresh`). If we call `appServiceDataProvider.refresh` directly, we get intellisense, automatic renaming, etc.

Issue #30, opened 2017-09-11
View & Edit Web App settings


whiskeyjay:
The extension shall allow the users to view and edit web app settings.

whiskeyjay:
Completed by #80

PR #76, opened 2017-09-22
Add, Edit, Delete settings for web apps and deployment slots

whiskeyjay:
https://github.com/Microsoft/vscode-azureappservice/issues/30

image

image


src/extension.ts
```diff
@@ -101,6 +102,23 @@ export function activate(context: vscode.ExtensionContext) {
             await node.swapDeploymentSlots(outputChannel, azureAccount);
         }
     });
+    initAsyncCommand(context, 'appService.AppSettings', async (context: any) => {
```
nturinski: Do we really need to check which type of NodeBase the context is?

PR #80, opened 2017-09-25
Take 2: Add, Edit, Delete settings for web apps and deployment slots (Explorer)

whiskeyjay:
A complete different design of #30.

image

image

image


whiskeyjay:
@nturinski this PR contains the refactoring to the nodes so now you can easily get a node's parent node as well as the root tree data provider.

src/explorer/appSettingsNodes.ts
```diff
@@ -0,0 +1,206 @@
...
+            contextValue: "applicationSettings",
+            iconPath: { 
+                light: path.join(__filename, '..', '..', '..', '..', 'resources', 'light', 'Settings_16x_vscode.svg'),
+                dark: path.join(__filename, '..', '..', '..', '..', 'resources', 'dark', 'Settings_16x_vscode.svg')
+            }
+        }
+    }
+
+    async getChildren(): Promise<NodeBase[]> {
+        this._settings = await this.getSettings();
```
EricJizbaMSFT: nit: both `this.getSettings()` and `this.generateTreeItems` are only used by this function. I would prefer to have the code directly in getChildren()

src/explorer/nodeBase.ts
```diff
@@ -20,11 +19,17 @@ export class NodeBase {
         };
     }
 
-    async getChildren(azureAccount: AzureAccountWrapper): Promise<NodeBase[]> {
-        return [];
+    getTreeDataProvider<T extends TreeDataProvider<NodeBase>>(): T {
```
EricJizbaMSFT: I think we can get rid of 'T' here. It will always be the same type in this codebase

nturinski:
@whiskeyjay Suh-weeeet

src/explorer/appSettingsNodes.ts
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    async edit(): Promise<void> {
+        const oldKey = this.key;
+        const newKey = await vscode.window.showInputBox({
+            ignoreFocusOut: true,
+            prompt: 'Enter setting key',
+            value: this.key,
+            validateInput: v => (<AppSettingsNode>this.getParentNode()).validateNewKeyInput(v, oldKey)
```
EricJizbaMSFT: I feel like all of these should either be `this.getParentNode<AppSettingsNode>()` or `<AppSettingsNode>this.parentNode`, right?

PR #82, opened 2017-09-27
Add Delete context menu item to the TreeView

### LINKED ISSUE #57: Add Delete context menu item to the TreeView
Deployment Slot (delete with confirmation)
Web App (deep link to portal delete)

nturinski:
Fixes #57 
Delete web apps and web services. 
Still need to implement for Deployment Slots.


src/explorer/appServiceNode.ts
```diff
@@ -85,6 +85,34 @@ export class AppServiceNode extends NodeBase {
         return this.start();
     }
 
+    async delete(azureAccount: AzureAccountWrapper): Promise<boolean> {
+        let servicePlanName = this.site.serverFarmId.substring(this.site.serverFarmId.lastIndexOf('/serverfarms/') + '/serverfarms/'.length);
+        let servicePlanRG = this.site.serverFarmId.substring(this.site.serverFarmId.indexOf('resourceGroups/') + 'resourceGroups/'.length, this.site.serverFarmId.indexOf('/providers/'));
```
EricJizbaMSFT: It's unlikely, but this doesn't work for a resource group named 'providers'. What if you used a regex isntead:
```
const matches = this.site.serverFarmId.match("^\/subscriptions\/.*\/resourceGroups\/(.*)\/providers\/Microsoft\.Web\/serverfarms\/(.*)");
const servicePlanRG = matches[1];
const servicePlanName = matches[2];
```

src/explorer/appServiceNode.ts
```diff
@@ -85,6 +85,34 @@ export class AppServiceNode extends NodeBase {
...
+        let servicePlan = await this.getWebSiteManagementClient(azureAccount).appServicePlans.get(servicePlanRG, servicePlanName);
+        let options = {};
```
EricJizbaMSFT: I don't think options is ever used?

src/explorer/appServiceNode.ts
```diff
@@ -85,6 +85,34 @@ export class AppServiceNode extends NodeBase {
...
+        options['deleteEmptyServerFarms'] = false;
+        let mOptions: MessageOptions = { modal: true };
+        if (servicePlan.numberOfSites < 2) {
+            let input = await window.showInformationMessage(`This is the last web app on the plan, "${servicePlanName}".  Delete the Web App and App Service Plan?`, mOptions, ...['Both', 'Web App ONLY']);
+            if (input) {
+                let deleteServicePlan = false;
```
EricJizbaMSFT: You can simplify this to:
```
const deleteServicePlan = input === 'Both';
```

src/explorer/appServiceNode.ts
```diff
@@ -85,6 +85,34 @@ export class AppServiceNode extends NodeBase {
...
+                if (input === 'Both') {
+                    deleteServicePlan = true;
+                }
+                await this.getWebSiteManagementClient(azureAccount).webApps.deleteMethod(this.site.resourceGroup, this.site.name, { deleteEmptyServerFarm: deleteServicePlan });
+                return true;
+            }
+            return false;
+        } else {
+            let input = await window.showInformationMessage(`Delete the Web App "${this.site.name}"?`, mOptions, ...['Confirm']);
```
EricJizbaMSFT: Instead of `...['Confirm']` you can just do `'Confirm'`

src/explorer/appServiceNode.ts
```diff
@@ -85,6 +85,34 @@ export class AppServiceNode extends NodeBase {
...
 
+    async delete(azureAccount: AzureAccountWrapper): Promise<boolean> {
+        let servicePlanName = this.site.serverFarmId.substring(this.site.serverFarmId.lastIndexOf('/serverfarms/') + '/serverfarms/'.length);
+        let servicePlanRG = this.site.serverFarmId.substring(this.site.serverFarmId.indexOf('resourceGroups/') + 'resourceGroups/'.length, this.site.serverFarmId.indexOf('/providers/'));
+        let servicePlan = await this.getWebSiteManagementClient(azureAccount).appServicePlans.get(servicePlanRG, servicePlanName);
+        let options = {};
+        options['deleteEmptyServerFarms'] = false;
+        let mOptions: MessageOptions = { modal: true };
+        if (servicePlan.numberOfSites < 2) {
+            let input = await window.showInformationMessage(`This is the last web app on the plan, "${servicePlanName}".  Delete the Web App and App Service Plan?`, mOptions, ...['Both', 'Web App ONLY']);
```
EricJizbaMSFT: Instead of `...['Both', 'Web App ONLY']` you can just do `'Both', 'Web App ONLY'`

src/extension.ts
```diff
@@ -64,20 +64,32 @@ export function activate(context: vscode.ExtensionContext) {
             await node.restart().then(() => outputChannel.appendLine(`App "${node.site.name}" has been restarted.`), err => outputChannel.appendLine(err));
         }
     });
+    initAsyncCommand(context, 'appService.Delete', async (node: AppServiceNode) => {
+        if (node) {
+            outputChannel.appendLine(`Deleting App "${node.site.name}"...`);
+            let result = await node.delete(azureAccount);
+            if (result) {
+                outputChannel.appendLine(`App "${node.site.name}" has been deleted.`), err => outputChannel.appendLine(err);
```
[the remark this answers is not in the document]
EricJizbaMSFT: I'm guessing you copied from the functions above, but those aren't implemented the best way. Start, stop, and restart actually return a nested promise. Instead of
```
async restart(): Promise<void> {
        await this.stop();
        return this.start();
    }
```
It should be:
```
async restart(): Promise<void> {
        await this.stop();
        await this.start();
    }
```

And then this file could be changed to:
```
await node.stop();
outputChannel.appendLine(`App "${node.site.name}" has been stopped.`)
```

If you want to fix those while you're here that'd be great :) Otherwise I can do it later

src/explorer/appServiceNode.ts
whiskeyjay: Do you think showWarningMessage is better for operations like delete?

src/explorer/appServiceNode.ts
whiskeyjay: On the portal the message is: Are you sure you want to delete "${this.site.name}"

Would be nice if we can align with the portal.

PR #88, opened 2017-10-02
Nturinski/improved local git

nturinski:
* Adds dirty repo check and asks user if they want to commit changes
* Checks scmType before localGit deploy.  Prompts user if they want to update config.

### DIFF
--- a/src/explorer/siteNodeBase.ts
+++ b/src/explorer/siteNodeBase.ts
@@ -212,25 +212,9 @@ export class SiteNodeBase extends NodeBase {
         let git = require('simple-git/promise')(workspace.rootPath);
 
         try {
-            await git.init();
             let status = await git.status();
             if (status.files.length > 0) {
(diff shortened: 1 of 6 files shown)


src/explorer/siteNodeBase.ts
```diff
@@ -165,37 +165,62 @@ export class SiteNodeBase extends NodeBase {
...
         let git = require('simple-git/promise')(workspace.rootPath);
 
         try {
             await git.init();
+            let status = await git.status();
+            if (status.files.length > 0) {
+                let input = await window.showInformationMessage(`There are uncommitted changes in local repo "${workspace.rootPath}". Commit?`, 'Yes');
+                if (input) {
+                    await git.add('./*');
+                    await git.commit('Deployed through VS Code');
```
whiskeyjay: Is there a way to show the source control UI of VSC so users can have more options for commit (like customize the commit message, double checking the changes, etc.)?

PR #90, opened 2017-10-02
Make app setting name validation case insensitive to match portal behavior

whiskeyjay:


src/explorer/appSettingsNodes.ts
```diff
@@ -111,12 +111,18 @@ export class AppSettingsNode extends NodeBase {
     }
 
     validateNewKeyInput(newKey: string, oldKey?: string): string {
-        if (!newKey || newKey.trim().length === 0) {
+        newKey = newKey ? newKey.trim() : '';
+        oldKey = oldKey ? oldKey.trim().toLocaleLowerCase() : oldKey;
```
EricJizbaMSFT: I think programatic values like an application setting ignore locale. We only need to take locale into account when we're displaying something to the user. I don't even know how you would set the locale for your app service...

PR #121, opened 2017-10-06
Fix #120

### LINKED ISSUE #120: Wrong order of "Disconnect" and "Connect" in log stream output
Steps to repro:

1. Connect to log stream.
2. Immediately connect to log stream again.

The 2nd connect command will disconnect the first connection and start a new one. However, in the output "Connect" is printed first, then "Disconnect".

whiskeyjay:
Fix #120


src/explorer/siteNodeBase.ts
```diff
@@ -74,9 +74,9 @@ export class SiteNodeBase extends NodeBase {
 
         if (util.isSiteDeploymentSlot(this.site)) {
             const slotName = util.extractDeploymentSlotName(this.site);
-            await this.webSiteClient.webApps.stopSlot(rgName, siteName, slotName);
+            await this.webSiteClient.webApps.stopSlotWithHttpOperationResponse(rgName, siteName, slotName);
```
EricJizbaMSFT: Discussed offline and I don't think 'stopSlotWithHttpOperationResponse' is necessary (but also wouldn't hurt)

PR #124, opened 2017-10-09
Move 'generate deployment script' to its own command

### LINKED ISSUE #98: Refactor how deploy script is displayed to the user
We've received some feedback that users are confused about whether or not they need to run the deploy script after they finish the create wizard. They don't realize that we create the web app for them

There's a few options:
1. Don't show the script, and create a separate 'Generate deploy script' action
1. Show the script, but make it more obvious that we are creating a web app for the user
1. Show the script, but add comments at the beginning describing how it should be used

Personally, I prefer the first two options. I don't think people read comments and the file already has quite a few comments

whiskeyjay:
Fix #98 

Instead of popping up the script at the end of the create wizard, use a context menu to create a deployment script using the selected web app as template.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
             },
             {
                 "command": "appService.DeploymentScript",
-                "title": "Generate Deployment Script",
+                "title": "Generate Azure CLI Script",
                 "category": "Azure App Service"
             },
             {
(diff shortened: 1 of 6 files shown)


src/util.ts
```diff
@@ -105,4 +105,33 @@ export function errToString(error: any): string {
...
+
+    for (let i = 0; i < parts.length; i += 2) {
+        const key = parts[i];
+        const value = parts[i + 1];
+
+        if (key === '' || value === '') {
+            throw invalidIdErr;
+        }
+
+        result[parts[i]] = parts[i + 1];
```
EricJizbaMSFT: nit: `result[key] = value;`

package.json
```diff
@@ -94,6 +95,11 @@
                 "category": "Azure App Service"
             },
             {
+                "command": "appService.DeploymentScript",
+                "title": "Generate Deployment Script",
```
EricJizbaMSFT: Can we change this name and/or the grouping? Right now it sounds like you're creating a script to deploy _to_ the app service (rather than deploying _the_ app service)

Maybe `Generate Creation Script` (Since we call the other command `Create Web App`)

PR #126, opened 2017-10-10
Nturinski/enhancement110

nturinski:
Tabling null checks for now-- all other tsconfig rules are enforced.

Did not realize that I did not have the tslint file yet :(


nturinski:
Need to pend merge until Eric's PR is merged otherwise Travis CI build will break.

src/wizard.ts
```diff
@@ -130,7 +130,7 @@ export class WizardBase {
 export interface WizardResult {
     status: WizardStatus;
     step: WizardStep;
-    error: Error;
+    error: Error | null;
```
EricJizbaMSFT: undefined

src/webAppCreator.ts
```diff
@@ -43,13 +43,13 @@ export class WebAppCreator extends WizardBase {
         return (<WebsiteStep>websiteStep).website;
     }
 
-    protected beforeExecute(step: WizardStep, stepIndex: number) {
+    protected beforeExecute(_step: WizardStep, stepIndex: number) {
```
EricJizbaMSFT: Can we just get rid of beforeExecute completely? This is the only place it's used and the output already has enough info:
```
Start creating new Web App...
The new Web App will be created in subscription "" ().
Existing resource group "" will be used.
Existing App Service Plan "" will be used.
Creating new Web App ""...
Web App "" created: 
```

Specifically, the 1st and 5th lines seem redundant to me

nturinski: @whiskeyjay I have no opinion on this-- Jay?

PR #128, opened 2017-10-11
Stephwe/quickcreate

### LINKED ISSUE #31: "Quick Create Web App" alternative
Current "Create New Web App" wizard has too many prompts, we should consider have a "Quick Create" alternative so users can create a new web app after answering two or three prompts.

Or, the extension should have settings that remember certain default settings, for example, the Node.js version, so it won't prompt each time a user wants to create a new web app.

StephenWeatherford:
Remember user selections, fixes #31

### DIFF
--- a/src/webAppCreator.ts
+++ b/src/webAppCreator.ts
@@ -13,12 +13,13 @@ import * as WebSiteModels from '../node_modules/azure-arm-website/lib/models';
 import * as util from './util';
 
 export class WebAppCreator extends WizardBase {
-    constructor(output: vscode.OutputChannel, readonly azureAccount: AzureAccountWrapper, subscription?: SubscriptionModels.Subscription) {
+    constructor(output: vscode.OutputChannel, readonly azureAccount: AzureAccountWrapper, persistence: vscode.Memento, subscription: SubscriptionModels.Subscription) {
         super(output);
-        this.steps.push(new SubscriptionStep(this, azureAccount, subscription));
-        this.steps.push(new ResourceGroupStep(this, azureAccount));
-        this.steps.push(new AppServicePlanStep(this, azureAccount));
-        this.steps.push(new WebsiteStep(this, azureAccount));
+        this.steps.push(new SubscriptionStep(this, azureAccount, subscription, persistence));
+        this.steps.push(new WebsiteNameStep(this, azureAccount, persistence));
+        this.steps.push(new ResourceGroupStep(this, azureAccount, persistence));
+        this.steps.push(new AppServicePlanStep(this, azureAccount, persistence));
+        this.steps.push(new WebsiteStep(this, azureAccount, persistence));
     }
 
     async run(promptOnly = false): Promise<WizardResult> {
@@ -39,8 +40,7 @@ export class WebAppCreator extends WizardBase {
     }
 
     get createdWebSite(): WebSiteModels.Site {
-        const websiteStep = this.steps.find(step => step instanceof WebsiteStep);
-        return (<WebsiteStep>websiteStep).website;
+        return this.findStepOfType(WebsiteStep).website;
     }
 
     protected beforeExecute(step: WizardStep, stepIndex: number) {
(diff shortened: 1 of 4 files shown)


src/webAppCreator.ts
```diff
@@ -130,29 +137,34 @@ class ResourceGroupStep extends WebAppCreatorStepBase {
...
+        const quickPickOptions = { placeHolder: `Select the resource group to create the new Web App in. (${this.stepProgressText})` };
         const subscription = this.getSelectedSubscription();
         const resourceClient = new ResourceManagementClient(this.azureAccount.getCredentialByTenantId(subscription.tenantId), subscription.subscriptionId);
         var resourceGroups: ResourceModels.ResourceGroup[];
-        var locations: SubscriptionModels.Location[];
         const resourceGroupsTask = util.listAll(resourceClient.resourceGroups, resourceClient.resourceGroups.list());
-        const locationsTask = this.azureAccount.getLocationsBySubscription(this.getSelectedSubscription());
-        const quickPickItemsTask = Promise.all([resourceGroupsTask, locationsTask]).then(results => {
+        var locationsTask = this.azureAccount.getLocationsBySubscription(this.getSelectedSubscription());
+        var locations = await locationsTask;
```
whiskeyjay: why await here? Shouldn't line 161 be
```
Promise.all([resourceGroupsTask, locationsTask])
```

src/webAppCreator.ts
```diff
@@ -189,18 +203,19 @@ class ResourceGroupStep extends WebAppCreatorStepBase {
                 return null;
             }
         });
+
         const locationPickItems = locations.map<QuickPickItemWithData<SubscriptionModels.Location>>(location => {
             return {
                 label: location.displayName,
-                description: `(${location.name})`,
+                description: `(${location.name}) `,
```
whiskeyjay: Why the extra space?

src/webAppCreator.ts
```diff
@@ -477,7 +479,7 @@ class WebsiteStep extends WebAppCreatorStepBase {
         return [
             {
                 name: 'node|4.4',
-                displayName: 'Node.js 4.4'
+                displayName: 'Node.js 4.4',
```
whiskeyjay: extra ","?

src/wizard.ts
```diff
@@ -113,6 +118,12 @@ export class WizardBase {
         this.output.appendLine(text);
     }
 
+    protected onRunError(step: WizardStep, stepIndex: number, error: Error) {
+        if (!(error instanceof UserCancelledError)) {
+            vscode.window.showErrorMessage(error.message);
```
whiskeyjay: Sometimes the message can be very long (for example, the Kudu server can return an entire HTML page of 4xx or 5xx error). Not sure in this case the showErrorMessage will be the best experience.

src/wizard.ts
```diff
@@ -147,14 +162,29 @@ export class WizardStep {
...
+            // See if the previous value selected by the user is in this list, and move it to the top as default
+            var previousId = this.persistenceState.get(persistenceKey);
+            var previousItem = previousId && resolvedItems.find(item => item.persistenceId === previousId);
+            if (previousItem) {
+                resolvedItems = ([previousItem]).concat(resolvedItems.filter(item => item !== previousItem));
+            }
+        }
+
+        const result =
+            await vscode.window.showQuickPick(resolvedItems, options, token);
```
whiskeyjay: nit: same line?

src/webAppCreator.ts
StephenWeatherford: Yep, thanks.  Fixed.

src/webAppCreator.ts
StephenWeatherford: fixed

src/webAppCreator.ts
StephenWeatherford: fixed

src/wizard.ts
StephenWeatherford: What do you suggest?

src/wizard.ts
StephenWeatherford: fixed

src/wizard.ts
whiskeyjay: How about send this to the output channel instead? At least the full message can be seen. HTML should be readable to developers. Not saying this is a perfect solution, I'll create an issue with Kudu to ask for better server side error messages, for example, in JSON instead of HTML.

StephenWeatherford: As discussed, this would be during the wizard UI interaction, thus the warning bar seemed more appropriate here.

PR #130, opened 2017-10-12
Improvements to App Settings

### LINKED ISSUE #94: Show app setting value as a child node of the key
Today we have the key and value in the same node separated by a colon. I propose we display the value as a child node of the key for a few reasons:

1. If the key and/or value has a colon it can be hard to distinguish where one starts and the other ends
screen shot 2017-10-03 at 9 38 51 am
1. People often put secrets in application settings. This way they could look at all of their keys and individually show/hide the value as desired.
1. I see users changing values more often than keys. I was actually suprised when I clicked 'edit' and the key popped up. This way we could have an edit action on each node with only one input box each.

whiskeyjay:
Fix #94  (parts of it)

Use "=" instead of ":" to separate key and values. Now it aligns with OS behavior when you list environment variables from command line.

image

Separate commands for editing Key and Value.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
         "onCommand:deploymentSlot.SwapSlots",
         "onCommand:appSettings.Add",
         "onCommand:appSettings.Edit",
+        "onCommand:appSettings.Rename",
         "onCommand:appSettings.Delete",
         "onCommand:diagnostics.OpenLogStream",
         "onCommand:diagnostics.StopLogStream",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -121,12 +122,17 @@
             },
             {
                 "command": "appSettings.Edit",
-                "title": "Edit Setting...",
+                "title": "Edit...",
```
StephenWeatherford: Minor: I'd suggest "Edit Value..."

src/explorer/appSettingsNodes.ts
```diff
@@ -169,38 +169,37 @@ export class AppSettingNode extends NodeBase {
...
             ignoreFocusOut: true,
-            prompt: 'Enter setting key',
-            value: this.key,
-            validateInput: v => this.getParentNode<AppSettingsNode>().validateNewKeyInput(v, oldKey)
+            prompt: `Enter setting value for "${this.key}"`,
+            value: this.value
         });
 
-        if (!newKey) {
+        if (!newValue) {
```
StephenWeatherford: Isn't empty string allowed for values?

package.json
whiskeyjay: I believe in the context menu, with "Rename" next to it, the meaning of "Edit" should be clear enough. 🙂

PR #134, opened 2017-10-16
Fix local git

### LINKED ISSUE #115: Local Git Deploy shouldn't always push master
I see two options, we can either:

- always push current branch and mention the branch name in output for awareness.

or 

- show a QuickPick with branches. I think this is a bit too chatty, and VSC already provides an easy way to see the working branch and switching branches in the status bar.

nturinski:
Fixes #115 
*Deploy current branch to azure remote repo
*Deploy to Deployment Slots


src/explorer/siteNodeBase.ts
```diff
@@ -208,46 +192,53 @@ export class SiteNodeBase extends NodeBase {
...
+                if (err.message.indexOf('spawn git ENOENT') >= 0) {
+                    const input = await window.showErrorMessage(`Git must be installed to use Local Git Deploy.`, `Install`);
+                    if (input === 'Install') {
+                        opn(`https://git-scm.com/downloads`);
+                    }
+                    throw err;
+                } else if (err.message.indexOf('error: failed to push') >= 0) {
+                    const input = await window.showErrorMessage(`Push rejected due to Git history diverging. Force push?`, `Yes`);
+                    if (input === 'Yes') {
+                        forcePush = true;
```
EricJizbaMSFT: Can we just call `await git.push(['-f', remote, 'HEAD:master']);` here and get rid of the for-loop

src/explorer/siteNodeBase.ts
```diff
@@ -258,4 +249,23 @@ export class SiteNodeBase extends NodeBase {
         const serverFarmId = util.parseAzureResourceId(this.site.serverFarmId.toLowerCase());
         return await this.webSiteClient.appServicePlans.get(serverFarmId.resourcegroups, serverFarmId.serverfarms);
     }
+
+    protected async convertScmType(node: SiteNodeBase, config: WebSiteModels.SiteConfigResource): Promise<void> {
```
EricJizbaMSFT: nit: private instead of protected

Also I like `updateScmType` better than `convertScmType` (it matches `updateConfiguration` below)

src/explorer/siteNodeBase.ts
nturinski: When the await git.push(['-f', remote, 'HEAD:master']);  was  on line 221, there was no error handling.  I'd have to write another try/catch loop within my catch.

src/explorer/siteNodeBase.ts
nturinski: I like that better too! I was thinking about adding an info window that alarms the scmType is being updated even if it was originally none. Thoughts?

src/explorer/siteNodeBase.ts
EricJizbaMSFT: What error handling, though? We know git is already installed...and you're already doing a force push so the `error: failed to push` logic doesn't apply

src/explorer/siteNodeBase.ts
EricJizbaMSFT: Are you talking about removing the check on line 259 so that the message always displays? If so, sounds good to me

If you're talking about a separate message, than maybe consider the output window instead. I don't want two messages popping up within seconds

src/explorer/siteNodeBase.ts
```diff
@@ -156,37 +156,28 @@ export class SiteNodeBase extends NodeBase {
         }
     }
 
-    async localGitDeploy(): Promise<boolean> {
+    public async localGitDeploy(): Promise<boolean> {
```
whiskeyjay: public is the default, did you add "public" because it's required by tslint?

src/explorer/siteNodeBase.ts
```diff
@@ -156,37 +156,28 @@ export class SiteNodeBase extends NodeBase {
...
-                input = await window.showWarningMessage(`Deployment source for "${this._siteName}" is set to "${oldScmType}". Change to "LocalGit"?`, 'Yes');
-            }
-
-            if (oldScmType === 'None' || input === 'Yes') {
-                !this._isSlot ?
-                    await this.webSiteClient.webApps.updateConfiguration(this.site.resourceGroup, this._siteName, updateConfig) :
-                    await this.webSiteClient.webApps.updateConfigurationSlot(this.site.resourceGroup, this._siteName, updateConfig, this._slotName);
-            } else {
+            let update = await this.updateScmType(this, config);
+            if (!update) {
```
EricJizbaMSFT: In this case it's hard to tell what the boolean means. For example, I initially thought 'true' meant the update succeeded and 'false' meant it failed. 

Can we move `throw new UserCancelledError()` back into `updateScmType` instead of returning false?

src/explorer/siteNodeBase.ts
```diff
@@ -197,46 +188,42 @@ export class SiteNodeBase extends NodeBase {
...
-                let input = await window.showErrorMessage(`Push rejected due to Git history diverging. Force push?`, `Yes`)
+                const input = await window.showErrorMessage(`Push rejected due to Git history diverging. Force push?`, `Yes`);
                 if (input === 'Yes') {
-                    await git.push(['-f', remote, 'master']);
+                    await git.push(['-f', remote, 'HEAD:master']);
                 } else {
                     throw err;
                 }
-            } else {
-                throw err;
```
EricJizbaMSFT: I think we still want to `throw err` in this case, otherwise it will be lost

src/explorer/siteNodeBase.ts
```diff
@@ -197,46 +188,42 @@ export class SiteNodeBase extends NodeBase {
...
         }
 
         const newDeployment = !this._isSlot ?
             await this.webSiteClient.webApps.listDeployments(this.site.resourceGroup, this._siteName) :
-            await this.webSiteClient.webApps.listDeploymentsSlot(this.site.resourceGroup, this._siteName, this._slotName);
+            await this.webSiteClient.webApps.listDeploymentsSlot(this.site.resourceGroup, this._siteName, util.extractDeploymentSlotName(this.site));
         // if the oldDeployment has a length of 0, then there has never been a deployment
         if (newDeployment.length && oldDeployment.length &&
             newDeployment[0].deploymentId === oldDeployment[0].deploymentId) {
             await window.showErrorMessage(`Local Git repo is current with "${repo}".`);
```
whiskeyjay: This can be removed. Let the caller of this method handle errors and show message if necessary.

src/explorer/siteNodeBase.ts
```diff
@@ -247,4 +234,21 @@ export class SiteNodeBase extends NodeBase {
...
         return await this.webSiteClient.appServicePlans.get(serverFarmId.resourcegroups, serverFarmId.serverfarms);
     }
+
+    private async updateScmType(node: SiteNodeBase, config: WebSiteModels.SiteConfigResource): Promise<boolean> {
+        let input;
+        const oldScmType = config.scmType;
+        const updateConfig = config;
+        updateConfig.scmType = 'LocalGit';
+        input = await window.showWarningMessage(`Deployment source for "${node._siteName}" is set as "${oldScmType}".  Change to "LocalGit"?`, 'Yes');
+        if (input) {
```
whiskeyjay: Personally I prefer this to be explicit: if (input === 'Yes') - in case in the future someone adds a new option but forgets to update the if line.

src/explorer/siteNodeBase.ts
```diff
@@ -197,46 +188,42 @@ export class SiteNodeBase extends NodeBase {
...
                 window.showWarningMessage(`There ${status.files.length > 1 ? 'are' : 'is'} ${status.files.length} uncommitted change${status.files.length > 1 ? 's' : ''} in local repo "${workspace.rootPath}"`);
             }
-            await git.push(remote, 'master');
-        }
-        catch (err) {
+            await git.push(remote, 'HEAD:master');
+        } catch (err) {
             if (err.message.indexOf('spawn git ENOENT') >= 0) {
-                let input = await window.showErrorMessage(`Git must be installed to use Local Git Deploy.`, `Install`)
+                const input = await window.showErrorMessage(`Git must be installed to use Local Git Deploy.`, `Install`);
```
whiskeyjay: Declare a const string for "Install" and use the const, easier for localization.

src/explorer/siteNodeBase.ts
```diff
@@ -156,37 +156,28 @@ export class SiteNodeBase extends NodeBase {
...
-            if (oldScmType !== 'None') {
-                input = await window.showWarningMessage(`Deployment source for "${this._siteName}" is set to "${oldScmType}". Change to "LocalGit"?`, 'Yes');
-            }
-
-            if (oldScmType === 'None' || input === 'Yes') {
-                !this._isSlot ?
-                    await this.webSiteClient.webApps.updateConfiguration(this.site.resourceGroup, this._siteName, updateConfig) :
-                    await this.webSiteClient.webApps.updateConfigurationSlot(this.site.resourceGroup, this._siteName, updateConfig, this._slotName);
-            } else {
+            let update = await this.updateScmType(this, config);
```
whiskeyjay: const?

src/explorer/siteNodeBase.ts
```diff
@@ -156,37 +156,28 @@ export class SiteNodeBase extends NodeBase {
...
-            await this.webSiteClient.webApps.listPublishingCredentials(this.site.resourceGroup, this._siteName) :
-            await this.webSiteClient.webApps.listPublishingCredentialsSlot(this.site.resourceGroup, this._siteName, this._slotName);
-        const config = !this._isSlot ?
-            await this.webSiteClient.webApps.getConfiguration(this.site.resourceGroup, this._siteName) :
-            await this.webSiteClient.webApps.getConfigurationSlot(this.site.resourceGroup, this._siteName, this._slotName);
-        const oldDeployment = !this._isSlot ?
-            await this.webSiteClient.webApps.listDeployments(this.site.resourceGroup, this._siteName) :
-            await this.webSiteClient.webApps.listDeploymentsSlot(this.site.resourceGroup, this._siteName, this._slotName);
+        if (!this._isSlot) {
+            publishCredentials = await this.webSiteClient.webApps.listPublishingCredentials(this.site.resourceGroup, this._siteName);
```
whiskeyjay: Use Promise.all() to have all three tasks running simultaneously.

PR #135, opened 2017-10-16
Nturinski/enhancement108

### LINKED ISSUE #108: Swap production with deployment slots
Currently, deployment slots can only be swapped with another deployment slot.  User should be able to swap their production (the web app) with a deployment slot.

nturinski:
Fixes #108


package.json
```diff
@@ -285,6 +285,11 @@
                     "group": "3_appServiceDeployCommands@2"
                 },
                 {
+                    "command": "appService.DeploymentScript",
```
EricJizbaMSFT: What's this change? A bad merge maybe?

DeploymentScript is already listed below

src/deploymentSlotActions.ts
```diff
@@ -54,7 +54,12 @@ class SwapStep extends WizardStep {
 
     async prompt(): Promise<void> {
         const deploymentSlots: DeploymentSlotNode[] = await this.slot.getParentNode<DeploymentSlotsNode>().getChildren();
-        let otherSlots: QuickPickItemWithData<null>[] = [];
+        let otherSlots: QuickPickItemWithData<null>[] = [{
+            label: 'production',
```
EricJizbaMSFT: Does the portal use lower case 'p'? I'd prefer uppercase, but I'm fine if that's what the portal does

src/deploymentSlotActions.ts
```diff
@@ -76,21 +81,20 @@ class SwapStep extends WizardStep {
...
-                    } else {
-                        // the format of Azure error messages are JSON strings
-                        this.wizard.writeline(JSON.parse(err.message).Message);
-                    }
-                });
+            try {
+                this.targetSlot === 'production' ?
+                    await client.webApps.swapSlotWithProductionWithHttpOperationResponse(this.slot.site.resourceGroup, this.slot.site.repositorySiteName, { targetSlot: this.sourceSlot.label, preserveVnet: true }) :
+                    await client.webApps.swapSlotSlotWithHttpOperationResponse(this.slot.site.resourceGroup, this.slot.site.repositorySiteName, { targetSlot: this.targetSlot, preserveVnet: true }, this.sourceSlot.label);
+            } catch (err) {
```
EricJizbaMSFT: Remove this catch - doesn't do anything meanful

src/deploymentSlotActions.ts
```diff
@@ -76,21 +81,20 @@ class SwapStep extends WizardStep {
...
-                    if (!err && response.statusCode === 200) {
-                        this.wizard.writeline(`"${this.targetSlot}" was swapped with "${this.sourceSlot.label}"`);
-                    } else {
-                        // the format of Azure error messages are JSON strings
-                        this.wizard.writeline(JSON.parse(err.message).Message);
-                    }
-                });
+            try {
+                this.targetSlot === 'production' ?
+                    await client.webApps.swapSlotWithProductionWithHttpOperationResponse(this.slot.site.resourceGroup, this.slot.site.repositorySiteName, { targetSlot: this.sourceSlot.label, preserveVnet: true }) :
```
EricJizbaMSFT: Is there a `swapSlotWithProduction` method?

Not sure why the 'WithHttpOperationResponse' is used

src/extension.ts
```diff
@@ -166,7 +166,13 @@ export function activate(context: vscode.ExtensionContext) {
     });
     initAsyncCommand(context, 'deploymentSlot.SwapSlots', async (node: DeploymentSlotNode) => {
         if (node) {
-            await node.swapDeploymentSlots(outputChannel);
+            try {
+                outputChannel.appendLine('Initializing deployment slot swap...');
+                await node.swapDeploymentSlots(outputChannel);
+            } catch (err) {
+                outputChannel.appendLine(err.message);
```
EricJizbaMSFT: This is a somewhat common pattern used in this extension - but I would prefer if generic errors bubbled up to be caught in our telemetry

Additionally - I feel errors should use `window.showErrorMessage()` instead of output channel, which might not be open

src/deploymentSlotActions.ts
whiskeyjay: Portal uses "production". Since the slot name is part of the site URL, although case insensitive, it is usually all lower case.

src/deploymentSlotActions.ts
```diff
@@ -76,21 +81,20 @@ class SwapStep extends WizardStep {
...
-            await client.webApps.swapSlotSlot(this.slot.site.resourceGroup, this.slot.site.repositorySiteName, { targetSlot: this.targetSlot, preserveVnet: true }, this.sourceSlot.label, {},
-                (err, _result, _request, response) => {
-                    if (!err && response.statusCode === 200) {
-                        this.wizard.writeline(`"${this.targetSlot}" was swapped with "${this.sourceSlot.label}"`);
-                    } else {
-                        // the format of Azure error messages are JSON strings
-                        this.wizard.writeline(JSON.parse(err.message).Message);
-                    }
-                });
+            try {
```
whiskeyjay: Should the actual swap happen in "execute" instead of "prompt"?

src/extension.ts
```diff
@@ -166,7 +166,13 @@ export function activate(context: vscode.ExtensionContext) {
     });
     initAsyncCommand(context, 'deploymentSlot.SwapSlots', async (node: DeploymentSlotNode) => {
         if (node) {
-            await node.swapDeploymentSlots(outputChannel);
+            try {
+                outputChannel.appendLine('Initializing deployment slot swap...');
```
whiskeyjay: Is this necessary? When user executes the command, will there be a long wait without any UI indication of progress?

src/extension.ts
```diff
@@ -97,14 +97,8 @@ export function activate(context: vscode.ExtensionContext) {
                 vscode.commands.executeCommand('appService.Refresh', node.getParentNode());
             } catch (err) {
                 if (!(err instanceof UserCancelledError)) {
```
EricJizbaMSFT: Just remove the try/catch completely. Even UserCancelledError should bubble up. `initAsyncCommand` will list the result as 'Canceled' in our telemetry that way

src/extension.ts
```diff
@@ -97,14 +97,8 @@ export function activate(context: vscode.ExtensionContext) {
...
-                    try {
-                        // Azure REST error messages come as a JSON string with more details
-                        outputChannel.appendLine(JSON.parse(err.message).Message);
```
EricJizbaMSFT: Can you add this logic into `initAsyncCommand`? We still prefer to display `JSON.parse(err.message).Message`

nturinski: I believe the logic is in the util.errToString function to parse messages correctly.

src/extension.ts
nturinski: After starting the swap, it can take about a minute for smaller apps to finish.  I haven't tested it with larger web apps, but I suspect it might take longer.

src/extension.ts
```diff
@@ -235,6 +198,7 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
             } else {
                 result = 'Failed';
                 errorData = util.errToString(err);
+                vscode.window.showErrorMessage(errorData);
```
EricJizbaMSFT: errorData returns a stringified object including 'Error' and 'Message' fields. We just want the message part to be displayed

PR #138, opened 2017-10-19
Nturinski/local git tele

nturinski:
*Adds more Error types for Local Git telemetry
*Change sending telemetry to be consistent with Azure Function

### DIFF
--- a/src/ErrorData.ts
+++ b/src/ErrorData.ts
@@ -1,0 +1,32 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+export class ErrorData {
+    public readonly message: string;
+    public readonly errorType: string;
+    // tslint:disable-next-line:no-any
+    constructor(error: any) {
+        if (error instanceof Error) {
+            try {
+                this.errorType = JSON.parse(error.message).Code;
+                this.message = JSON.parse(error.message).Message;
+            } catch (err) {
+                this.errorType = error.constructor.name;
+                this.message = error.message;
+            }
+
+        } else if (typeof (error) === 'object' && error !== null) {
+            this.errorType = (<object>error).constructor.name;
+            this.message = JSON.stringify(error);
+            // tslint:disable-next-line:no-unsafe-any
+        } else if (error !== undefined && error !== null && error.toString && error.toString().trim() !== '') {
+            this.errorType = typeof (error);
+            // tslint:disable-next-line:no-unsafe-any
+            this.message = error.toString();
+        } else {
+            this.errorType = typeof (error);
+            this.message = 'Unknown Error';
+        }
+    }
+}
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -18,25 +18,19 @@ export class GitNotInstalledError extends Error {
 }
 
 export class LocalGitDeployError extends Error {
-    public readonly name: string;
-    public readonly message: string;
     public readonly servicePlanSize: string;
     constructor(error: Error, servicePlanSize: string) {
         super();
-        this.name = error.constructor.name;
         this.message = error.message;
         this.servicePlanSize = servicePlanSize;
     }
 }
 
 export class WizardFailedError extends Error {
-    public readonly name: string;
-    public readonly message: string;
     public readonly stepTitle: string;
     public readonly stepIndex: number;
     constructor(error: Error, stepTitle: string, stepIndex: number) {
         super();
-        this.name = error.constructor.name;
         this.message = error.message;
         this.stepTitle = stepTitle;
         this.stepIndex = stepIndex;
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -1,5 +1,21 @@
+import * as opn from 'opn';
+import { window } from 'vscode';
+
 export class UserCancelledError extends Error { }
-export class GitNotInstalledError extends Error { }
+export class GitNotInstalledError extends Error {
+    constructor() {
+        super();
+        this.showInstallPrompt();
+    }
+
+    public async showInstallPrompt(): Promise<void> {
+        const installString = `Install`;
+        const input = await window.showErrorMessage(`Git must be installed to use Local Git Deploy.`, installString);
+        if (input === 'Install') {
+            opn(`https://git-scm.com/downloads`);
+        }
+    }
+}
 
 export class LocalGitDeployError extends Error {
     public readonly name: string;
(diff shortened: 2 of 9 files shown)


src/ErrorData.ts
```diff
@@ -0,0 +1,32 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+export class ErrorData {
+    public readonly message: string;
+    public readonly errorType: string;
+    // tslint:disable-next-line:no-any
+    constructor(error: any) {
+        if (error instanceof Error) {
+            try {
+                this.errorType = JSON.parse(error.message).Code;
```
EricJizbaMSFT: Just curious - what's an example value for 'Code'? What does it look like?

src/explorer/siteNodeBase.ts
```diff
@@ -183,7 +183,7 @@ export class SiteNodeBase extends NodeBase {
         if (config.scmType !== 'LocalGit') {
             await this.updateScmType(this, config);
         }
-
+        const servicePlan = await this.getAppServicePlan();
```
EricJizbaMSFT: nit: Move this down to line 215 since that's the only place it's used

src/extension.ts
```diff
@@ -187,23 +188,36 @@ function initCommand(context: vscode.ExtensionContext, commandId: string, callba
...
 
         try {
             await callback(...args);
         } catch (err) {
+            if (err instanceof LocalGitDeployError) {
+                properties.servicePlan = err.servicePlanSize;
+            }
             if (err instanceof UserCancelledError) {
                 result = 'Canceled';
+                errorData = new ErrorData(err);
```
EricJizbaMSFT: No need for errorData in this case - setting result as 'Canceled' is enough

src/extension.ts
```diff
@@ -187,23 +188,36 @@ function initCommand(context: vscode.ExtensionContext, commandId: string, callba
...
+            } else if (err instanceof GitNotInstalledError) {
+                result = 'Failed';
+                errorData = new ErrorData(err);
             } else {
                 result = 'Failed';
-                errorData = util.errToString(err);
-                vscode.window.showErrorMessage(errorData);
+                errorData = new ErrorData(err);
+                vscode.window.showErrorMessage(errorData.message);
                 throw err;
```
EricJizbaMSFT: Don't throw 'err' here. We don't want to risk VS Code showing the error message again

src/extension.ts
```diff
@@ -187,23 +188,36 @@ function initCommand(context: vscode.ExtensionContext, commandId: string, callba
...
+                properties.servicePlan = err.servicePlanSize;
+            }
             if (err instanceof UserCancelledError) {
                 result = 'Canceled';
+                errorData = new ErrorData(err);
+            } else if (err instanceof GitNotInstalledError) {
+                result = 'Failed';
+                errorData = new ErrorData(err);
             } else {
                 result = 'Failed';
```
EricJizbaMSFT: You can just set `properties.result` directly and remove the `result` variable

src/explorer/siteNodeBase.ts
```diff
@@ -203,16 +203,16 @@ export class SiteNodeBase extends NodeBase {
                 if (input === 'Install') {
                     opn(`https://git-scm.com/downloads`);
                 }
-                throw err;
+                throw new GitNotInstalledError();
```
EricJizbaMSFT: Can you move the `showErrorMessage()` and git url logic either into `GitNotInstalledError` or directly into `initAsyncCommand`? (If you put it in `GitNotInstalledError`, I imagine it would be in a 'prompt' or 'showMessage' function that gets called by `initAsyncCommand`)

That way we could potentially throw this `GitNotInstalledError()` in other places and it's still handled the same way

src/ErrorData.ts
nturinski: From what I can tell, it's a one line word describing the error.  For instance, if you try to deploy two slots to production, the code will come back as "Conflict".

src/webAppZipPublisher.ts
```diff
@@ -197,7 +205,7 @@ class DeployStep extends WizardStep {
         const subscriptionStep = this.wizard.findStepOfType(SubscriptionStep);
 
         if (!subscriptionStep.subscription) {
-            throw new Error('A subscription must be selected first.');
+            throw new Error('A subscription must be selected first.';
```
EricJizbaMSFT: Add the parentheses back?

src/wizard.ts
```diff
@@ -25,21 +25,7 @@ export abstract class WizardBase {
             try {
                 await this.steps[i].prompt();
             } catch (err) {
-                this.sendErrorTelemetry(step, err);
-                this.onRunError(step, i, err);
```
EricJizbaMSFT: Can we remove onRunError and just add this code? It's the same in every place:
```
if (error instanceof UserCancelledError) {
    return;
}
throw new WizardFailedError(error, step.stepTitle, step.stepIndex);
```

src/wizard.ts
```diff
@@ -64,22 +50,7 @@ export abstract class WizardBase {
...
-                    };
-                } else {
-                    this._result = {
-                        status: 'Faulted',
-                        step: step,
-                        error: err
-                    };
-                }
-                return this._result;
+                this.onExecuteError(err, step);
```
EricJizbaMSFT: Can we remove onExecuteError and just add this code? It's the same in every place:
```
if (error instanceof UserCancelledError) {
    return;
}
throw new WizardFailedError(error, step.stepTitle, step.stepIndex);
```

Also - onExecuteError shouldn't display anything because initAsyncCommand does that

src/wizard.ts
```diff
@@ -118,20 +89,11 @@ export abstract class WizardBase {
...
-        }
-    }
+    protected abstract onRunError(error: Error, step: WizardStep)
 
     protected abstract beforeExecute(step?: WizardStep, stepIndex?: number);
 
-    protected abstract onExecuteError(error: Error, step?: WizardStep, stepIndex?: number)
+    protected abstract onExecuteError(error: Error, step: WizardStep)
 
     protected sendErrorTelemetry(step: WizardStep, error: any) {
```
EricJizbaMSFT: Remove this as it is no longer used, correct?

src/errors.ts
```diff
@@ -1 +1,44 @@
...
+    constructor(error: Error, servicePlanSize: string) {
+        super();
+        this.name = error.constructor.name;
+        this.message = error.message;
+        this.servicePlanSize = servicePlanSize;
+    }
+}
+
+export class WizardFailedError extends Error {
+    public readonly name: string;
```
EricJizbaMSFT: name is never used?

src/errors.ts
```diff
@@ -1 +1,44 @@
...
+    public readonly message: string;
```
EricJizbaMSFT: message should already exist on `Error` - I don't think you need to re-define it

src/wizard.ts
nturinski: It displays on the output pane so I didn't want to remove it without asking.  I see some value in having it on the output in case somebody wanted to investigate into the error message more and needed a reference for later.

EricJizbaMSFT: Okay makes sense - but that should apply for every step, though, correct? So we could still get rid of the 'onExecuteError' thing and display in output pane here?

nturinski:
@EricJizbaMSFT Thank you! I have learned a lot from this PR-- on my way to becoming an expert.

PR #140, opened 2017-10-19
Add multiple workspace to local git deploy

### LINKED ISSUE #97: Zip/Git deploy should support multi-root workspaces
VSCode 1.17 introduces multi-root workspaces. However, our Zip deploy and Local Git deploy always use the "first" folder. Instead we should show a quick pick if there are multiple workspace roots

nturinski:
Fixes #97

### DIFF
--- a/src/explorer/siteNodeBase.ts
+++ b/src/explorer/siteNodeBase.ts
@@ -162,7 +162,7 @@ export class SiteNodeBase extends NodeBase {
             throw new Error(`There is no open folder to deploy.`);
         }
 
-        const fsPath = await util.showWorkspaceFoldersQuickPick();
+        const fsWorkspaceFolder = await util.showWorkspaceFoldersQuickPick();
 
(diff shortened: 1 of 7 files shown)


src/explorer/siteNodeBase.ts
```diff
@@ -253,4 +257,26 @@ export class SiteNodeBase extends NodeBase {
...
+                }
+            }
+        });
+
+        const folderQuickPickOption = { placeHolder: `Select the folder to Local Git deploy.` };
+        const pickedItem = folderQuickPickItems.length == 1 ?
+            folderQuickPickItems[0] : await window.showQuickPick(folderQuickPickItems, folderQuickPickOption);
+
+        if (!pickedItem) {
+            throw new UserCancelledError();
```
whiskeyjay: When this function becomes shared code, I think it would be better to just return undefined instead of throwing UserCancelledError. Let the caller decide whether or not to throw what error.

PR #141, opened 2017-10-21
Refactor in preparation to sharing wizard

StephenWeatherford:
Refactored webappcreator to be more generic so it can be used to create function apps, too.  Doing this before pulling it out into the shared repo to make it easier to understand the changes (and kept previous filename for the same reason, will change afterwards).


src/webAppCreator2.ts
```diff
@@ -0,0 +1,57 @@
...
+            this.persistence));
+    }
+
+    protected beforeExecute(_step: WizardStep, stepIndex: number) {
+        if (stepIndex == 0) {
+            this.writeline('Creating new Web App...');
+        }
+    }
+
+    protected onExecuteError(error: Error) {
```
EricJizbaMSFT: This was removed in nathan's last PR and isn't used anymore

src/webAppCreator.ts
```diff
@@ -11,16 +13,19 @@ import WebSiteManagementClient = require('azure-arm-website');
 import * as WebSiteModels from '../node_modules/azure-arm-website/lib/models';
 import * as util from './util';
 
-export class WebAppCreator extends WizardBase {
-    constructor(output: vscode.OutputChannel, readonly azureAccount: AzureAccountWrapper, subscription: SubscriptionModels.Subscription, persistence?: vscode.Memento) {
+export type WebsiteOS = "linux" | "windows";
+export type AppKind = "app" | "functionapp";
```
EricJizbaMSFT: nit: Did you pick these names or are you using values from Azure?

If you picked them, I would prefer 'webapp' over 'app'. If you didn't pick them - shame on Azure :P

src/webAppCreator.ts
```diff
@@ -11,16 +13,19 @@ import WebSiteManagementClient = require('azure-arm-website');
...
+
+export abstract class BaseWebsiteCreator extends WizardBase {
```
EricJizbaMSFT: nit: the naming is inconsistent with `WizardBase` and `WebsiteCreatorStepBase`

It should be `WebsiteCreatorBase` to match

src/webAppCreator.ts
```diff
@@ -235,22 +240,22 @@ class ResourceGroupStep extends WebAppCreatorStepBase {
...
-        super(wizard, 'Select or create App Service Plan', azureAccount, persistence);
+    constructor(wizard: WizardBase, azureAccount: AzureAccountWrapper, private _appKind: AppKind, private _websiteOS: WebsiteOS, persistence: vscode.Memento) {
+        super(wizard, 'Select or create an App Service Plan', azureAccount, persistence);
     }
 
     async prompt(): Promise<void> {
         const createNewItem: QuickPickItemWithData<WebSiteModels.AppServicePlan> = {
             persistenceId: "$new",
-            label: '$(plus) Create New App Service Plan',
+            label: '$(plus) Create New App Service',
```
EricJizbaMSFT: Why remove 'Plan'?

src/webAppCreator.ts
```diff
@@ -402,11 +412,23 @@ class AppServicePlanStep extends WebAppCreatorStepBase {
     }
 }
 
-class WebsiteStep extends WebAppCreatorStepBase {
+interface WebsiteStepResources {
+    title: string;    // like "Create Web App"
```
EricJizbaMSFT: The only difference in these strings is the Resource name (Web App). Can't we infer that from the appKind?

src/webAppCreator.ts
```diff
@@ -600,12 +647,36 @@ class WebsiteNameStep extends WebAppCreatorStepBase {
         return this._websiteName;
     }
 
-    get suggestedRGAndPlanName(): string {
-        return this._suggestedRGAndPlanName;
+    public suggestRelatedName(): Promise<string> {
```
EricJizbaMSFT: Can we rename this to `getSuggestedName`? The current name makes me think it will prompt the user to 'suggest' a name

src/webAppCreator.ts
```diff
@@ -3,6 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+// TODO: Will rename this file to BaseWebsiteCreator after this PR, to make changes easier to understand
```
EricJizbaMSFT: I feel like this is a sign the file is too big 😂: 
screen shot 2017-10-23 at 10 40 26 am

We could split it up now or when we move this to the shared package

src/webAppCreator.ts
StephenWeatherford: Azure values, and of course they've evolved and can't be changed, so I don't blame them.

src/webAppCreator.ts
StephenWeatherford: It's mostly just a sign that GitHub doesn't handle renames well.  However, I agree that the file is too big.

src/webAppCreator.ts
StephenWeatherford: Because whoops.

src/webAppCreator.ts
StephenWeatherford: You could if you want to have English terminology inside of localized strings.

src/webAppCreator.ts
StephenWeatherford: => computeRelatedName

src/webAppCreator2.ts
StephenWeatherford: Thanks.  Merges seem harder in git than VSTS.

src/webAppCreator.ts
EricJizbaMSFT: 'Web App' or 'Function App' are product names - so I don't think they would be localized anyway

StephenWeatherford:
Unfortunately they are:

[cid:image001.png@01D34BF5.4911CB50]
However, most of them probably are understandable, so I’m fine removing it.

PR #144, opened 2017-10-24
Display multiline error msg in output log

### LINKED ISSUE #142: Multiline error messages are not displayed to the user
The `vscode.window.showErrorMessage` api can only handle one line at a time.

@nturinski and I discussed a possible solution. We could detect if the error message contains a new line and display something generic like "The operation failed - check the output window for more details" and then print the error to the output window

nturinski:
Fixes #142 

Do we want to display in output for every error or just multiline?

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -191,6 +191,7 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
     context.subscriptions.push(vscode.commands.registerCommand(commandId, async (...args: any[]) => {
         const start = Date.now();
         const properties: { [key: string]: string; } = {};
+        const output = util.getOutputChannel();
         properties.result = 'Succeeded';
         let errorData: ErrorData | undefined;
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -214,7 +214,13 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
             } else {
                 properties.result = 'Failed';
                 errorData = new ErrorData(err);
-                vscode.window.showErrorMessage(errorData.message);
+                if (errorData.message.includes('\n')) {
+                    util.getOutputChannel().appendLine(errorData.message)
+                    vscode.window.showErrorMessage('An error has occured. Check output log for more details.');
```
EricJizbaMSFT: nit: Can this be output _window_ instead of _log_? If it just says 'log', then users might try to find a log file

src/extension.ts
```diff
@@ -214,7 +214,13 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
             } else {
                 properties.result = 'Failed';
                 errorData = new ErrorData(err);
-                vscode.window.showErrorMessage(errorData.message);
+                if (errorData.message.includes('\n')) {
+                    util.getOutputChannel().appendLine(errorData.message)
```
EricJizbaMSFT: Should we call outputChannel.show()? I lean yes - just because the output window has multiple 'channels' and it could be hard for users to find the 'App Service' one

EricJizbaMSFT:
If the error is _not_ multiline, we should definitely show it using `showErrorMessage`. However, I agree it might be worth it to show it in the output window _as well_ (just in case it's a long message that gets cutoff). I definitely wouldn't call 'outputChannel.show()' in that case though - don't want to pop the same exact text into the user's face twice

src/extension.ts
nturinski: I agree; it'll also hide that nasty UI glitch where the dropdown menu is black when you click it for the first time.

src/extension.ts
nturinski: Your wish is my command 👑

PR #150, opened 2017-10-25
tslint fix on explorer nodes

nturinski:


src/explorer/AppSettingsNodes.ts
```diff
@@ -35,39 +40,41 @@ export class AppSettingsNode extends NodeBase {
...
         if (this._settings.properties) {
-            for (let key in this._settings.properties) {
-                children.push(new AppSettingNode(
-                    key,
-                    this._settings.properties[key],
-                    this.getTreeDataProvider(),
-                    this
-                ));
+            for (const key in this._settings.properties) {
+                if (this._settings.properties[key]) {
```
StephenWeatherford: This is a functionality change, right?  It's valid to have a property that's an empty string, so this seems wrong...?

(BTW, not a good idea to put functionality changes into a PR like this.)

src/explorer/AppSettingsNodes.ts
```diff
@@ -138,39 +151,38 @@ export class AppSettingsNode extends NodeBase {
...
+        this.value = value;
     }
 
-    getTreeItem(): TreeItem {
+    public getTreeItem(): TreeItem {
         return {
             label: this.label,
             collapsibleState: TreeItemCollapsibleState.None,
-            contextValue: "applicationSettingItem",
+            contextValue: `applicationSettingItem`,
```
StephenWeatherford: single quotes (not template string)?

StephenWeatherford: Not a big deal, although I thought there was a rule against that.

tslint.json
```diff
@@ -11,7 +11,7 @@
         "no-exec-script": true,
         "no-function-constructor-with-string-args": true,
         "no-http-string": [
-            true,
+            false, // changed
```
StephenWeatherford: Is this turning the rule off?  I'd rather leave it on but disable per line.

tslint.json
```diff
@@ -58,7 +58,7 @@
         "no-duplicate-variable": true,
         "no-empty": true,
         "no-floating-promises": true,
-        "no-for-in-array": true,
+        "no-for-in-array": false, // changed
```
StephenWeatherford: Do we really want to turn this off?

src/explorer/AppSettingsNodes.ts
EricJizbaMSFT: @nturinski I assume you changed this because of 'no-unexternalized-strings'? That rule should be turned off in this repo until we support localization

src/explorer/AppSettingsNodes.ts
EricJizbaMSFT: Yeah I think the correct fix would be to make 'value' an optional string in the AppSettingNode constructor

whiskeyjay: A value should always be required for an app setting. Can it be:
```
const value = this._settings.properties[key] ? this._settings.properties[key] : '';
children.push(new AppSettingNode(
    key,
    value,
    this.getTreeDataProvider(),
    this
));
```

nturinski: It was the change recommended by TSLint for for-in array loops.  Ideally, they recommend not using for-in loops on arrays in general.  However, after speaking with Eric, I believe that we would want to use a for-of loop anyway.

https://stackoverflow.com/questions/29285897/what-is-the-difference-between-for-in-and-for-of-in-javascript

Should I make that change in my PR or hold off?

tslint.json
nturinski: Okay, can do.

src/explorer/AppSettingsNodes.ts
nturinski: I'm using Jay's implementation for now.  I had to disable the "forin" tslint rule which states: "Requires a `for ... in` statement to be filtered with an `if` statement."

EricJizbaMSFT: You are correct that it requires an if statement, but it requires a specific _type_ of if statement, which should always apply:
```
if (this._settings.properties.hasOwnProperty(key)) {
```

It's used to 'Prevents accidental iteration over properties inherited from an object’s prototype.'

EricJizbaMSFT: I would still prefer to keep 'no-for-in' and 'forin' rules. I believe this would satisfy both rules:
```
Object.keys(this._settings.properties).forEach((key: string) => {
    children.push(new AppSettingNode(
        key,
        this._settings.properties[key],
        this.getTreeDataProvider(),
        this
    ));
});
```

StephenWeatherford: +1

src/explorer/AppSettingsNodes.ts
```diff
@@ -3,30 +3,35 @@
...
 import { AppServiceDataProvider } from './AppServiceExplorer';
-import { SubscriptionModels } from 'azure-arm-resource';
-import { TreeItem, TreeItemCollapsibleState } from 'vscode';
-import WebSiteManagementClient = require('azure-arm-website');
+import { NodeBase } from './NodeBase';
 
 export class AppSettingsNode extends NodeBase {
+    public readonly site: WebSiteModels.Site;
+    public readonly subscription: SubscriptionModels.Subscription;
+    public readonly _isSlot: boolean;
```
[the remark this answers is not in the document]
nturinski: I created public get methods in cases where it may be used outside of the class.  I think AppSettingsNodes specifically made never need that functionality and I could probably get rid of it, but the wizard does get the subscription and site from the SiteNode itself.

I plan to make all the variable name changes in another PR (where private variables are prefixed with an underscore)

nturinski:
Are there any requested changes that I have missed?

PR #154, opened 2017-10-26
Remove error output from wizard onError

### LINKED ISSUE #146: Wizard error is displayed twice in output window
Repro Steps:
1. Turn off wifi
1. Create new web app
1. Go through steps until you get an error

Expected: error only printed once in output window
Actual:
screen shot 2017-10-24 at 3 09 28 pm

nturinski:
Fixes #146 
Added Error: prefix to output messages because it is not always immediately obvious that the outputted message is an error.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -216,12 +216,12 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
             } else {
                 properties.result = 'Failed';
                 errorData = new ErrorData(err);
-                output.appendLine(errorData.message);
+                output.appendLine(`Error: ${errorData.message}`);
                 if (errorData.message.includes('\n')) {
                     output.show();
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -221,7 +221,7 @@ function initAsyncCommand(context: vscode.ExtensionContext, commandId: string, c
                     output.show();
                     vscode.window.showErrorMessage('An error has occured. Check output window for more details.');
                 } else {
-                    vscode.window.showErrorMessage(errorData.message);
+                    vscode.window.showErrorMessage(`Error: ${errorData.message}`);
```
EricJizbaMSFT: I think you meant to do this on line 219 instead of here?

nturinski: You are correct. 😞 Fixing.

PR #155, opened 2017-10-26
Nturinski/issue97v2

### LINKED ISSUE #97: Zip/Git deploy should support multi-root workspaces

nturinski:
Fixes #97 

Pulled code from #140 to reimplement.  It seemed easier and less error-prone to just propagate the changes rather than trying to merge in the old PR.

VSCode has a natural behavior when showing the QuickPick window when no Workfolders are open that override our error message.  Which would we like the user to see?
image

image


src/util.ts
```diff
@@ -144,4 +145,31 @@ export function parseAzureResourceId(resourceId: string): { [key: string]: strin
     }
 
     return result;
-}
\ No newline at end of file
+}
+
+export async function showWorkspaceFoldersQuickPick(placeHolderString: string): Promise<vscode.WorkspaceFolder> {
```
EricJizbaMSFT: I don't have strong preferences in regards to your description of the PR, but whatever logic you decide on should be in this function (Rather than the current implementation where every caller of showWorkspaceFoldersQuickPick has to check if workspaceFolders is defined and throw a very similar error message)

src/util.ts
```diff
@@ -144,4 +145,31 @@ export function parseAzureResourceId(resourceId: string): { [key: string]: strin
...
+    const folderQuickPickItems = vscode.workspace.workspaceFolders.map((value) => {
```
EricJizbaMSFT: Don't you still have to check if 'vscode.workspace.workspaceFolders' is defined here? That's what I meant when I said move the error handling logic into this function

nturinski: If there are no opened workspaceFolders, VS Code shows the user "There are no entries to pick from" as displayed above.  This will result in the user having to cancel, so we shouldn't have to worry about error handling.  Our error was actually never shown to the user.

PR #160, opened 2017-10-30
Implement add deployment slot

### LINKED ISSUE #156: Create new deployment slots

nturinski:
Fixes #156 

I changed the implementation of AppServiceNodes.  There are some features that are exclusive to different plan tiers, so I thought it would be useful to check the AppServicePlan when populating children.  For example, Basic tier does not allow for DeploymentSlots.


src/explorer/AppServiceNode.ts
```diff
@@ -40,15 +40,19 @@ export class AppServiceNode extends SiteNodeBase {
         }
 
         const treeDataProvider = this.getTreeDataProvider<AppServiceDataProvider>();
-
+        const nodes = [];
+        const appServicePlan = await this.getAppServicePlan();
+        if (appServicePlan.sku.tier !== 'Basic') {
+            // Basic Service Plans do not support deployment slots
```
EricJizbaMSFT: I think we usually 'gray things out' if something isn't supported, rather than removing it completely. Since we can't 'gray it out', maybe we just show a message that says 'unsupported on basic plan' in the tree? Something analagous to the portal:

screen shot 2017-10-30 at 3 03 56 pm

src/explorer/DeploymentSlotsNode.ts
```diff
@@ -60,4 +61,60 @@ export class DeploymentSlotsNode extends NodeBase {
...
+        // a webapp cannot have more than 4 slots (excluding production)
+        const slotNodes = await this.getChildren();
+        const slotLabels = slotNodes.map(node => {
+            return node.label;
+        });
+
+        while (!slotNameOkay) {
+            slotName = await window.showInputBox({
+                prompt: 'Enter a unique name for the new deployment slot',
+                validateInput: (value: string) => {
```
EricJizbaMSFT: nit: I prefer this as a separate functions for the sake of readability

src/explorer/DeploymentSlotsNode.ts
```diff
@@ -60,4 +61,60 @@ export class DeploymentSlotsNode extends NodeBase {
...
+                    value = value ? value.trim() : '';
+                    if (!value.match(/^[a-z0-9\-]{1,60}$/ig)) {
+                        return 'Name should be 1-60 characters long and can only include alphanumeric characters and hyphens.';
+                    }
+
+                    // Can not have identical slot names OR production
+                    if (!this.checkSlotNameAvailability(value, slotLabels)) {
```
EricJizbaMSFT: nit: why not replace:
```
if (!this.checkSlotNameAvailability(value, slotLabels)) {
```
with 
```
if (slotName !== 'production' && slotLabels.indexOf(slotName) === -1) {
```

src/explorer/DeploymentSlotsNode.ts
```diff
@@ -60,4 +61,60 @@ export class DeploymentSlotsNode extends NodeBase {
...
+        let slotName = '';
+        let slotNameOkay = false;
+
+        // a webapp cannot have more than 4 slots (excluding production)
+        const slotNodes = await this.getChildren();
+        const slotLabels = slotNodes.map(node => {
+            return node.label;
+        });
+
+        while (!slotNameOkay) {
```
EricJizbaMSFT: slotNameOkay is never assigned to - I would just remove the while loop

src/explorer/DeploymentSlotsNode.ts
whiskeyjay: Use the WebsiteManagementClient.checkNameAvailability() method for this job.

src/explorer/DeploymentSlotsNode.ts
whiskeyjay: Duplicated code, is it possible to share the same code with Web App name validation?

EricJizbaMSFT: I was just talking about the validate function, not the whole prompt function, but it's fine either way

src/explorer/DeploymentSlotsNode.ts
EricJizbaMSFT: It's up to Nathan - but I actually prefer the existing code because it can be done synchronously inside the inputbox validate function

src/explorer/DeploymentSlotsNode.ts
```diff
@@ -60,4 +61,50 @@ export class DeploymentSlotsNode extends NodeBase {
...
+        return await window.showInputBox({
+            prompt: 'Enter a unique name for the new deployment slot',
+            validateInput: (value: string) => {
+                value = value ? value.trim() : '';
+                if (!value.match(/^[a-z0-9\-]{1,60}$/ig)) {
+                    return 'Name should be 1-60 characters long and can only include alphanumeric characters and hyphens.';
+                }
+
+                // Can not have identical slot names OR production
+                if (value === 'production' || slotLabels.indexOf(value) !== -1) {
```
whiskeyjay: Test case:
1. Create an Web App called "a-b".
2. Create another Web App called "a".
3. Now try to add a new deployment slot called "b" to Web App "a".

Now the new slot's domain name will be "a-b", which is already taken by the Web App in Step 1.

src/explorer/AppServiceNode.ts
EricJizbaMSFT: FYI - I'm still waiting on this before I'll approve the PR. If you feel strongly, I will also approve if you file an issue for the 0.4.0 milestone and we can discuss our options there

nturinski: Here's my current working solution.  I welcome suggestions for the Node Label.

image

image

nturinski: Child node language copies the portal's (pictured above)

src/explorer/DeploymentSlotsNode.ts
whiskeyjay: Since we are going to revisit this (and the Web App Creator Wizard) when async InputBox validation is supported, let's just create an issue for now and use the current code.

src/explorer/AppServiceNode.ts
EricJizbaMSFT: I might put 'N/A for Basic Service Plan' in parentheses, but otherwise looks good to me

nturinski:
Are the gray icons okay?  They are just some assets I made by altering the hex color of the .svg files.  Any opinions on it?

src/explorer/DeploymentSlotsNode.ts
nturinski: #166

tasq
```diff
@@ -0,0 +1,723 @@
+[33mcommit c2ec88196d217d965f07ca67ce61dafd0f8ccd31[m[33m ([m[1;36mHEAD -> [m[1;32mnturinski/issue97[m[33m)[m
```
EricJizbaMSFT: What is this file? Please remove :)

PR #168, opened 2017-11-07
User/cha/logpoints start

changsi-an:
Please take a look at my change regarding logpoints.


src/logPointsManager.ts
```diff
@@ -0,0 +1,420 @@
+var req = require("request");
```
changsi-an: These file contains many classes. Since the current code base is pretty flat, we are not sure where to put things. Can you help suggest so we can maybe categorize things into the right folder?

src/logPointsManager.ts
```diff
@@ -0,0 +1,420 @@
...
+            new Buffer(publishCredential.publishingUserName + ":" + publishCredential.publishingPassword)
+                .toString("base64")
+        };
+
+        var r = req.defaults({
+            baseUrl: "https://" + siteName + ".scm.azurewebsites.net/",
+            headers: headers,
+        });
+
+        r.cookie(`ARRAffinity=${affinityValue}`);
```
changsi-an: This is the reason why we send rest calls manually. Because the KuduClient dependency does not seem to support setting cookies. If this is not the case, please let us know.

src/logPointsManager.ts
```diff
@@ -0,0 +1,420 @@
...
+        }
+
+        return this._json;
+    };
+}
+
+export class LogPointsSessionAttach extends WizardBase {
+    private _cachedPublishCredential: WebSiteModels.User
+
+    readonly hasSlot: boolean;
```
changsi-an: We were debating whether we want to add "public" qualifiers to the public fields, since we saw that the existing code does not seem to do so. 

Please suggest.

whiskeyjay:
Great stuff! I'll look into it tomorrow (currently we are preparing a release for the coming Connect event).

src/logPointsManager.ts
whiskeyjay: We are working on enabling tslint, and once that is in, "public" will be required by tslint.

src/logPointsManager.ts
changsi-an: Sounds good, addressed in https://github.com/Microsoft/vscode-azureappservice/pull/177 instead.

src/logPointsManager.ts
changsi-an: Addressed in https://github.com/Microsoft/vscode-azureappservice/pull/177 instead.

package.json
```diff
@@ -144,6 +144,11 @@
                 "command": "diagnostics.StopLogStream",
                 "title": "Stop Streaming Logs",
                 "category": "Azure App Service"
+            },
+            {
+                "command": "diagnostics.SetLogPoints",
+                "title": "Launch logpoints debug session",
```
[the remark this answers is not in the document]
changsi-an: Addressed in https://github.com/Microsoft/vscode-azureappservice/pull/177 instead.

src/extension.ts
```diff
@@ -212,6 +213,13 @@ export function activate(context: vscode.ExtensionContext) {
             node.stopLogStream();
         }
     });
+    initAsyncCommand(context, 'diagnostics.SetLogPoints', async (node: SiteNodeBase) => {
+        if (node) {
+            const wizard = new LogPointsSessionAttach(outputChannel, azureAccount, node.site, node.subscription);
+            await wizard.run();
+
```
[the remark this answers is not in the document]
changsi-an: Addressed in https://github.com/Microsoft/vscode-azureappservice/pull/177 instead.

changsi-an:
This PR is abandoned in favor of https://github.com/Microsoft/vscode-azureappservice/pull/177 .

PR #179, opened 2017-11-14
Add Configure Deployment Source

nturinski:
+Add context menu to change scmType of web app (unable to do this for Deployment Slots)
-Remove Local Git Deploy from context menu

*In preparation to have extension to auto-detect scmType and pick appropriate deployment method.


nturinski:
The build will break until this PR https://github.com/Microsoft/vscode-azuretools/pull/17 is merged.

package.json
```diff
@@ -118,8 +118,8 @@
                 "category": "Azure App Service"
             },
             {
-                "command": "appService.LocalGitDeploy",
-                "title": "Local Git Deploy to Web App",
+                "command": "appService.EditScmType",
```
EricJizbaMSFT: nit: It bothers me that the command and the title are so different

Can we merge the names to be more similar?

src/extension.ts
```diff
@@ -134,6 +134,12 @@ export function activate(context: vscode.ExtensionContext): void {
             });
         }
     });
+    initAsyncCommand(context, 'appService.EditScmType', async (node: AppServiceNode) => {
+        if (node) {
+            const updatedScmType = await node.editScmType();
```
EricJizbaMSFT: Check if updatedScmType is undefined before displaying the message

package.json
nturinski: I don't mind renaming it.  

The portal action is setting up a deployment source whereas the configuration object holds a scmType property.  If we want it to be more similar, I'd lean more towards Deployment Source since that is more self-explanatory than scmType.

nturinski:
Redone with new ui changes.  Leverages node picker to use command palette .

nturinski:
This will break until https://github.com/Microsoft/vscode-azuretools/pull/27 is merged

Issue #4, opened 2017-08-21
Some commands shall prompt for context when executed from command palette


whiskeyjay:
For example, the commands in the App Service node context menu, when called directly from the command palette, don't do anything currently. The correct behavior should be a quick pick list of available App Services so the user can pick which one to execute the command against.

whiskeyjay:
For now we can just hide commands that requires a deep context from the web app explorer from the command palette.

PR #181, opened 2017-11-15
Leverage shared AzureExplorer

EricJizbaMSFT:
One main benefit is reducing the amount of copied code, but it also adds support for the following:
1. Load more
1. showNodePicker (so that all commands can be run from the command palette)

Improves https://github.com/Microsoft/vscode-azureappservice/issues/4

Sorry this is a huge change - I didn't think there was a way to move over in small chunks (Also I left the file names as they are so that it's hopefully easier to view)

Related to https://github.com/Microsoft/vscode-azuretools/pull/18


EricJizbaMSFT:
That's fine - I took care of it. Btw @nturinski I'm completely on your side now in terms of strictNullChecks. I think this was a great example where it could've prevented a bug (and I'm fine with letting people use the non-null-assertion if that's what it takes to turn this on)

package.json
```diff
@@ -184,79 +184,7 @@
                     "when": "never"
                 },
                 {
-                    "command": "appService.OpenInPortal",
```
StephenWeatherford: What are all these changes about?

PR #182, opened 2017-11-16
User/cha/setlogpoints

changsi-an:
Add the function to set/remove logpoints.

Recover the logpoints of a document from local cache.


src/diagnostics/logPointsClient.ts
```diff
@@ -122,23 +146,37 @@ export class MockLogpointsDebuggerClient extends LogPointsDebuggerClientBase imp
...
+    }
+
     public enumerateProcesses(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User): Promise<CommandRunResult<IEnumerateProcessResponse>> {
         return this.makeCallAndLogException<IEnumerateProcessResponse>(siteName, affinityValue, publishCredential, "curl -X GET http://localhost:32923/os/processes?applicationType=Node.js");
     }
 
     public attachProcess(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User, data: AttachProcessRequest): Promise<CommandRunResult<IAttachProcessResponse>> {
         return this.makeCallAndLogException<IAttachProcessResponse>(siteName, affinityValue, publishCredential,
-            `curl -X POST -H "Content-Type: application/json" -d '{"processId":"${data.processId}","codeType":"javascript"}' http://localhost:32923/debugger/session/${data.sessionId}/debugee`);
+                                                                    `curl -X POST -H "Content-Type: application/json" -d '{"processId":"${data.processId}","codeType":"javascript"}' http://localhost:32923/debugger/session/${data.sessionId}/debugee`);
```
mike-kaufman: whitespace looks off here...

src/diagnostics/logPointsClient.ts
```diff
@@ -185,30 +223,47 @@ export class CommandRunResult<ResponseType extends { error?: {}, data?: {} }> {
...
     }
 }
 
-// tslint:disable:max-classes-per-file
 export class LoadSourceRequest {
     constructor(public sessionId: string, public debugId: string, public sourceId: string) {
     }
 }
 
+export class SetLogpointRequest {
```
mike-kaufman: Dont' need to address this now, but I'm not sure why these are classes - interfaces would be more flexible here, as `{ sessionId, debugId, sourceId}` is duck-type compatible with your class LoadSourceRequest.

src/diagnostics/structs/Logpoint.ts
```diff
@@ -0,0 +1,8 @@
+export class Logpoint {
```
mike-kaufman: this could just be a interface & things would be more flexible.

src/diagnostics/structs/LogpointsCollection.ts
```diff
@@ -0,0 +1,71 @@
+import * as util from 'util';
+import * as vscode from 'vscode';
+import { Logpoint } from './Logpoint';
+// tslint:disable:align
+
+export class LogpointsCollection {
```
mike-kaufman: general concern here about keeping state in sync between your idea of logpoints & the actual logpoints that exist on the client.  Probably need some "refresh" logic that will completely refresh this data based on what the debugger claims is "truth".

src/telemetry/reporter.ts
```diff
@@ -2,7 +2,7 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-
+/* tslint:disable */
```
mike-kaufman: This is disabling linting for the entire file.  Can we disable individual rules on individual lines via `//tslint:disable-next-line:<rule>`?  Its a little more time-consuming, but it let's us enforce "no new linter errors".

src/util.ts
```diff
@@ -2,13 +2,12 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-
-import { reporter } from './telemetry/reporter';
+/* tslint:disable */
```
mike-kaufman: same here.

src/diagnostics/LogPointsManager.ts
```diff
@@ -0,0 +1,190 @@
+import * as util from 'util';
+import * as vscode from 'vscode';
+import { ISetLogpointResponse } from './logPointsClient';
+import { RemoteScriptSchema } from './remoteScriptDocumentProvider';
+import { Logpoint } from './structs/Logpoint';
+import { LogpointsCollection } from './structs/LogpointsCollection';
+
+class DebugSessionManager {
+    private _logpointsCollectionMapping: { [documentUri: string]: LogpointsCollection };
```
mike-kaufman: we're going to need to account for cases where the service gets restarted & all this state is invalidated.

src/diagnostics/logPointsClient.ts
```diff
@@ -2,12 +2,15 @@ import * as req from "request";
 
 import * as child_process from 'child_process';
 import * as WebSiteModels from '../../node_modules/azure-arm-website/lib/models';
+// tslint:disable:max-classes-per-file
```
mike-kaufman: Can you break up into multiple files vs just disabling the linter here?

src/telemetry/reporter.ts
changsi-an: This change should be reverted. It's a mistake.

src/diagnostics/structs/LogpointsCollection.ts
changsi-an: What you said is planed. We don't have a API for retrieving logpoints from server yet. After we have that (someone is working on it, so soon), we will have a "refresh" logic.

src/diagnostics/logPointsClient.ts
changsi-an: Will do

src/util.ts
changsi-an: This change should be reverted. It's a mistake.

mike-kaufman: :+1:  Please double check as I saw it in a few places.

src/diagnostics/LogPointsManager.ts
changsi-an: Added a task to track it for now.

src/diagnostics/logPointsClient.ts
changsi-an: This is due to 
```
"align": [

            true,

            "parameters",

            "arguments",

            "statements"

        ]
```
in tslint.json

src/diagnostics/structs/Logpoint.ts
changsi-an: Done

src/diagnostics/logPointsClient.ts
changsi-an: Done.

src/diagnostics/logPointsClient.ts
changsi-an: It's done as well.

PR #184, opened 2017-11-17
Update package.json and account extension dts

whiskeyjay:


package.json
```diff
@@ -497,7 +496,6 @@
         "vscode": "^1.0.0"
     },
     "dependencies": {
-        "archiver": "^2.0.3",
```
EricJizbaMSFT: You can remove from third party notices, too

PR #187, opened 2017-11-21
Remove Close button on warningMessages

### LINKED ISSUE #185: Use 'cancel' instead of 'close' on messages

nturinski:
Fixes #185 

I used "Not Now" for the log streaming prompt as regardless of which button you click, log streaming will start, but will not be "enabled" and the container won't be restarted.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -190,13 +190,14 @@ export function activate(context: vscode.ExtensionContext): void {
         }
 
         const client: WebSiteManagementClient = nodeUtils.getWebSiteClient(node);
-        const enableButton = 'Yes';
+        const enableButton: vscode.MessageItem = { title: 'Yes' };
+        const notNowButton: vscode.MessageItem = { title: 'Not Now', isCloseAffordance: true };
         const isEnabled = await vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, p => {
             p.report({ message: 'Checking container diagnostics settings...' });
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -169,13 +169,14 @@ export function activate(context: vscode.ExtensionContext): void {
...
-            const enableButton = 'Yes';
+            const enableButton: vscode.MessageItem = { title: 'Yes' };
+            const notNowButton: vscode.MessageItem = { title: 'Not Now', isCloseAffordance: true };
             const isEnabled = await vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, p => {
                 p.report({ message: 'Checking container diagnostics settings...' });
                 return node.isHttpLogsEnabled();
             });
 
-            if (!isEnabled && enableButton === await vscode.window.showWarningMessage('Do you want to enable logging and restart this container?', enableButton)) {
+            if (!isEnabled && enableButton === await vscode.window.showWarningMessage('Do you want to enable logging and restart this container?', enableButton, notNowButton)) {
```
EricJizbaMSFT: Usually we would give them 'Okay', 'Not Now', _and_ 'Cancel'

nturinski: I don't think this prompt is very clear as if you were to hit 'Cancel' today, it would still show a log-stream (what's available without enabling logging)

EricJizbaMSFT: I don't have a strong opinion on this - sounds fine to me

PR #197, opened 2017-11-28
Use Azure API to check name availability

### LINKED ISSUE #166: Use client API to check slot name availiability
When VS Code can handle async input validation, use the client API to check new slot names availability.

nturinski:
Fixes #166 

Confirmed that this works with the current build of VS Code

### DIFF
--- a/src/explorer/DeploymentSlotsTreeItem.ts
+++ b/src/explorer/DeploymentSlotsTreeItem.ts
@@ -10,6 +10,7 @@ import { IAzureNode, IAzureParentNode, IAzureParentTreeItem, IAzureTreeItem, Use
 import * as util from '../util';
 import { nodeUtils } from '../utils/nodeUtils';
 import { DeploymentSlotTreeItem } from './DeploymentSlotTreeItem';
+import { ResourceNameAvailability } from 'azure-arm-website/lib/models';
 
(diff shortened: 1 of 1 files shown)


src/explorer/DeploymentSlotsTreeItem.ts
```diff
@@ -75,19 +74,20 @@ export class DeploymentSlotsTreeItem implements IAzureParentTreeItem {
         return new DeploymentSlotTreeItem(newSite);
     }
 
-    private async promptForSlotName(slotLabels: string[]): Promise<string | undefined> {
+    private async promptForSlotName(node: IAzureParentNode<DeploymentSlotsTreeItem>): Promise<string | undefined> {
         return await window.showInputBox({
             prompt: 'Enter a unique name for the new deployment slot',
             ignoreFocusOut: true,
-            validateInput: (value: string) => {
+            validateInput: async (value: string) => {
```
whiskeyjay: Do you need to update the vscode engines value to the latest one in package.json in order to use async validation?

nturinski: No, I don't think so.  I tested it with npm update and with the package.json that is currently in the repo and both seem to work correctly.

nturinski:
I even reset my commit history to try to hide it.  So quick to check PRs 🥇

Issue #194, opened 2017-11-28
Re-enable error report when making Kudu calls


changsi-an:
Pertinent code

https://github.com/Microsoft/vscode-azureappservice/blob/bc69cfbeade8a095413c212d8e1a372c1d99e9a2/src/diagnostics/logPointsClient.ts#L50

And

https://github.com/Microsoft/vscode-azureappservice/blob/bc69cfbeade8a095413c212d8e1a372c1d99e9a2/src/diagnostics/structs/CommandRunResult.ts#L17

The error logging code was disabled. Because `util.getOutputChannel()` is not available from a debug adapter process. Some refactor need to be done to re-enable error reporting.

EricJizbaMSFT:
Logpoints was removed from this repo https://github.com/microsoft/vscode-azureappservice/pull/1101

PR #198, opened 2017-11-28
Use real Kudu apis to communicate with the remote agent.

changsi-an:
This PR primarily contains the change to use the Kudu apis to communicate with the real remote Agent.


src/diagnostics/LogPointsDebugAdapter.ts
```diff
@@ -81,7 +91,7 @@ export class LogPointsDebugAdapter extends LoggingDebugSession {
         if (command === 'loadSource') {
             const sourceId: string = args;
             const request: ILoadSourceRequest = { sessionId: this._sessionId, debugId: this._debugId, sourceId };
-            logPointsDebuggerClient.loadSource(null, null, null, request).then(
+            logPointsDebuggerClient.loadSource(this._siteName, this._affinityValue, this.getPublishCrednetial(), request).then(
```
mike-kaufman: `this.getPublishCrednetial()` is mis-spelled.

src/diagnostics/logPointsSessionWizard.ts
```diff
@@ -1,331 +1,321 @@
-import { SubscriptionModels } from 'azure-arm-resource';
```
mike-kaufman: why is this entire files showing as a diff?  Are the line-endings messed up?

src/diagnostics/structs/CommandRunResult.ts
```diff
@@ -10,12 +12,22 @@ export class CommandRunResult<ResponseType extends { error?: {}, data?: {} }> {
...
         return this._json;
     }
+
+    private parseStdOut(): void {
+        try {
+            const outputJson = JSON.parse(this.output);
+
+            this._stdout = outputJson.stdout;
+        } catch (err) {
+            this._stdout = null;
```
mike-kaufman: is it useful to log something when we can't parse stdout?

src/diagnostics/LogPointsDebugAdapter.ts
```diff
@@ -4,10 +4,12 @@ import {
     Thread
 } from 'vscode-debugadapter';
 import { DebugProtocol } from 'vscode-debugprotocol';
+import * as WebSiteModels from '../../node_modules/azure-arm-website/lib/models';
```
mike-kaufman: this import is weird.  why are you referencing something from inside node_modules? Can you just say `import * as WebSiteModesl from 'azure-arm-website'`?

whiskeyjay: There are two d.ts files in the referenced package, models and operations. Another option would be (since you are using just the type User):
```
import { User } from 'azure-arm-website/lib/models';
```

src/diagnostics/structs/CommandRunResult.ts
changsi-an: Agree. This is what https://github.com/Microsoft/vscode-azureappservice/issues/194 is talking about. This class is used in both the extension host and the debug adapter. Their respective logging method is very different. So I will have to make it a separate task, refactor and unify them then.

src/diagnostics/LogPointsDebugAdapter.ts
changsi-an: Fixed.

src/diagnostics/LogPointsDebugAdapter.ts
changsi-an: Used the way @whiskeyjay suggested.

src/diagnostics/logPointsSessionWizard.ts
changsi-an: It should be right now.

src/diagnostics/logPointsClient.ts
```diff
@@ -93,57 +116,75 @@ export class KuduLogPointsDebuggerClient extends LogPointsDebuggerClientBase imp
...
+        const request = req.defaults({});
+        request.cookie(`ARRAffinity=${affinityValue}`);
+        const opts = {
+            uri: `${KuduLogPointsDebuggerClient.getBaseUri(siteName)}/api/zip/site/wwwroot/`,
+            headers: {
+                'Content-Type': 'multipart/form-data'
             },
-            (err, response, body) => {
+            auth: KuduLogPointsDebuggerClient.getAuth(publishCredential),
+            body: fs.createReadStream(path.join(__dirname, '../../../resources/ssh-client.zip'))
```
whiskeyjay: I'm a bit concerned here:

1. A Zip file is not as easy to track with Git. Is there a specific reason to send the files as a Zip, like upload performance or something?
2. What if the customer's website happens to have a file in the root folder also called ssh-client.js (or ssh-client.js.map)?

changsi-an: 1. For both convenience and performance considerations. At this point, we don't know how many files we need to deploy to Kudu container. Potentially we will deploy a `node_moudules` set of files as well. The zip is created in another repository, so we have the content versioned in that repository already. I can change it to uploading one file today. But it is likely to change to uploading a zip again in the future anyway. Optionally we can webpack everything together into one file. But the content of the end result will not make sense for the context of an extension code base. So it acts more like just a binary.

This is supposed to be a utility that once tested, will not change frequently.

2. That's a valid point. I can optionally put it to "/home/logpoints_sshclient". How does that sound?

mike-kaufman: I'd suggest we have a directory structure like this:
  - `/home/logpoints/ssh-client`   // for ssh-client
  - `/home/logpoints/logs`           
  - `/home/logpoints/logs/agent`  //for logpoint-agent logs

changsi-an: PTAL

changsi-an: PTAL

whiskeyjay: Thanks for the explanation. I understand the ssh-client is versioned in a separate repository. However, this still doesn't solve the problem in **this** repository that there is no easy way to know the version of the current ssh-client.zip, nor can you tell which version of the ssh-client.zip it was from a commit history. Maybe adding a txt file describing the version of the zip file? Anyways, this is not something worth blocking the merge.

PR #200, opened 2017-11-29
Restrict Logpoints feature to selected app services types

changsi-an:
Filter out the ineligible app services and prevent them from using Logpoints feature

### DIFF
--- a/src/diagnostics/logPointsSessionWizard.ts
+++ b/src/diagnostics/logPointsSessionWizard.ts
@@ -56,6 +56,7 @@ export class LogPointsSessionAttach extends WizardBase {
     }
 
     protected initSteps(): void {
+        this.steps.push(new EligibilityCheck(this));
         if (util.isSiteDeploymentSlot(this.site)) {
             this.selectedDeploymentSlot = this.site;
         } else {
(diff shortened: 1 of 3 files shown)


mike-kaufman:
commit history on your branch here is looking really crazy.  Maybe you can rebase off master?

src/diagnostics/logPointsSessionWizard.ts
```diff
@@ -78,6 +79,40 @@ export class LogPointsSessionAttach extends WizardBase {
...
+
+        if (!linuxFxVersion) {
+            throw new Error('Cannot read "linuxFxVersion"');
+        }
+
+        const [framework, imageName] = linuxFxVersion.split('|');
+        const enabledImages = vscode.workspace.getConfiguration('appService').get<string[]>('enabledDockerImages');
+
+        if ('docker' !== framework.toLocaleLowerCase() || enabledImages.indexOf(imageName.toLocaleLowerCase()) === -1) {
+            throw new Error(`Please use one of the supported docker image. The ${framework}|${imageName} combination is not supported`);
```
mike-kaufman: for the error message, we probably want something that says what are the supported image names.  Or a link to something that lists them out.

changsi-an: I'd prefer adding a link. The image names are long. It might not necessarily provide a good experience to user if we enumerate them one by one and cluster the output in a small information prompt. So it's better to use a link. Will add a backlog item.

mike-kaufman: Please put the issue in the wiki - I don't want to lose sight of this before we ship out the MVP.  The concern here is if someone tries to use logpoints on an unsupported image, they have no idea how to determine which instance is supported, and they'll just give up.

changsi-an: It is in the wiki.

changsi-an:
The commit history has been fixed. Merging now...

PR #202, opened 2017-12-01
Nturinski/deploy

nturinski:
Will be broken until https://github.com/Microsoft/vscode-azuretools/pull/33 is merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,9 @@
         "onCommand:appService.Stop",
         "onCommand:appService.Restart",
         "onCommand:appService.Delete",
+        "onCommand:appService.Deploy",
+        "onCommand:appService.ConfigureDeploymentSource",
         "onCommand:appService.DeploymentScript",
-        "onCommand:appService.DeployZipPackage",
         "onCommand:appService.OpenVSTSCD",
         "onCommand:deploymentSlot.SwapSlots",
         "onCommand:appSettings.Add",
@@ -64,21 +65,21 @@
                 }
             },
             {
-                "command": "appService.Refresh",
-                "title": "Refresh",
-                "category": "Azure App Service",
-                "icon": {
-                    "light": "resources/light/refresh.svg",
-                    "dark": "resources/dark/refresh.svg"
-                }
-            },
-            {
                 "command": "appService.Deploy",
                 "title": "Deploy to Web App",
                 "category": "Azure App Service",
                 "icon": {
                     "light": "resources/light/Deploy.svg",
                     "dark": "resources/dark/Deploy.svg"
+                }
+            },
+            {
+                "command": "appService.Refresh",
+                "title": "Refresh",
+                "category": "Azure App Service",
+                "icon": {
+                    "light": "resources/light/refresh.svg",
+                    "dark": "resources/dark/refresh.svg"
                 }
             },
             {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -72,6 +72,15 @@
                     "dark": "resources/dark/refresh.svg"
                 }
             },
+            {
+                "command": "appService.Deploy",
```
EricJizbaMSFT: Add this to activationEvents

package.json
```diff
@@ -293,44 +307,34 @@
                     "group": "2_appServiceControlCommands@4"
                 },
                 {
-                    "command": "appService.DeployZipPackage",
+                    "command": "appService.OpenVSTSCD",
                     "when": "view == azureAppService && viewItem == appService",
                     "group": "3_appServiceDeployCommands@1"
                 },
-                {
-                    "command": "appService.LocalGitDeploy",
```
EricJizbaMSFT: You can remove `LocalGitDeploy` and `DeployZipPackage` completely correct? They're still in this file a few times

src/explorer/SiteTreeItem.ts
```diff
@@ -109,19 +109,6 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
-    public async localGitDeploy(client: WebSiteManagementClient): Promise<void> {
-        const fsWorkspaceFolder = await util.showWorkspaceFoldersQuickPick('Select the folder to Local Git deploy.');
-        try {
-            await this.siteWrapper.localGitDeploy(fsWorkspaceFolder.uri.fsPath, client, getOutputChannel());
-        } catch (err) {
-            if (err instanceof UserCancelledError) {
-                throw err;
-            }
-            const appServicePlan = await getAppServicePlan(this.site, client);
-            throw new SiteActionError(err, appServicePlan.sku.size);
```
EricJizbaMSFT: I would add this SiteActionError logic back to the main Deploy command

package.json
```diff
@@ -249,6 +258,11 @@
                     "command": "appService.Refresh",
                     "when": "view == azureAppService",
                     "group": "navigation@2"
+                },
+                {
+                    "command": "appService.Deploy",
+                    "when": "view == azureAppService",
+                    "group": "navigation@3"
```
EricJizbaMSFT: I would put this between create and refresh. Refresh is consistently the last command for all our extensions

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azureappservice",
```
EricJizbaMSFT: Not sure if you were just waiting for it to be published, but make sure to update the version of the appservice npm package in this file

package.json
```diff
@@ -293,44 +298,34 @@
                     "group": "2_appServiceControlCommands@4"
                 },
                 {
-                    "command": "appService.DeployZipPackage",
```
whiskeyjay: Why doesn't the App Service node have the deploy command?

EricJizbaMSFT: Per discussion with @fiveisprime we thought this was a confusing entry point and replaced it with the one on the title bar. The node entry point implied you were deploying the app service itself (as opposed to deploying TO the app service)

whiskeyjay: What does it mean by "deploying the app service itself"? The app service is already there, why would someone think the service itself needs to be deployed again?

whiskeyjay: I'll approve for now. We can discuss this a bit more offline.

PR #203, opened 2017-12-01
Base64 encode the POST body before sending it out.

changsi-an:
(Require agent to be updated to the latest.)

### DIFF
--- a/src/diagnostics/logPointsClient.ts
+++ b/src/diagnostics/logPointsClient.ts
@@ -68,8 +68,11 @@ export class KuduLogPointsDebuggerClient extends LogPointsDebuggerClientBase imp
         };
     }
 
-    private static base64Encode(s: string): string {
-        const buf = Buffer.from(s, 'utf8');
+    private static base64Encode(payload: string | object): string {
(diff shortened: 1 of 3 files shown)


src/diagnostics/logPointsClient.ts
```diff
@@ -202,9 +223,16 @@ export class MockLogpointsDebuggerClient extends LogPointsDebuggerClientBase imp
     }
 
     public attachProcess(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User, data: IAttachProcessRequest): Promise<CommandRunResult<IAttachProcessResponse>> {
+        const jsonPayload = JSON.stringify({
```
dhanvikapila: Maybe better to do the JSON.stringify inside the base64encode method on a JS object? That way the clients can send you any data instead of just a string?

changsi-an: Sounds like a good idea to remove some duplication. will do.

changsi-an: Done.

PR #205, opened 2017-12-04
Capture and display the error when setting up a logpoint.

changsi-an:
Surface errors from setting a log point on the UI.

### DIFF
--- a/src/diagnostics/LogPointsDebugAdapter.ts
+++ b/src/diagnostics/LogPointsDebugAdapter.ts
@@ -114,6 +114,7 @@ export class LogPointsDebugAdapter extends LoggingDebugSession {
                     if (result.isSuccessful()) {
                         response.body = result.json;
                     } else {
+                        logger.error(`Cannot set logpoint. ${result.error}`);
                         response.body = {
                             error: result.error
                         };
--- a/src/diagnostics/LogPointsManager.ts
+++ b/src/diagnostics/LogPointsManager.ts
@@ -1,5 +1,6 @@
 import * as util from 'util';
 import * as vscode from 'vscode';
+import * as appserviceUtil from '../util';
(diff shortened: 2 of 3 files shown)


src/diagnostics/LogPointsDebugAdapter.ts
```diff
@@ -114,6 +114,7 @@ export class LogPointsDebugAdapter extends LoggingDebugSession {
                     if (result.isSuccessful()) {
                         response.body = result.json;
                     } else {
+                        logger.error(`Cannot set logpoint, the error is: ${result.error}`);
```
whiskeyjay: Nit: do you think "Cannot set logpoint. ${result.error}" would be a bit cleaner?

src/diagnostics/LogPointsManager.ts
```diff
@@ -81,6 +82,11 @@ class DebugSessionManager {
         }
 
         const result: ISetLogpointResponse = await this._debugSession.customRequest("setLogpoint", { scriptId, lineNumber, columnNumber, expression });
+        if (result.error) {
+            vscode.window.showErrorMessage("Cannot set logpoint successfully. Please refer to DEBUG CONSOLE for details.");
+            appserviceUtil.getOutputChannel().appendLine(`Cannot set logpoint successfully, the error is ${result.error}.`);
```
whiskeyjay: Do you want to show the output here? - in case the output window is not already opened.

src/diagnostics/LogPointsDebugAdapter.ts
changsi-an: Done

src/diagnostics/LogPointsManager.ts
changsi-an: Yes. Added that.

PR #212, opened 2017-12-05
Clear out the glyphs after a debug session termination.

changsi-an:
And also improve the error message given when user try to set a logpoint to an inactive document.

### DIFF
--- a/src/diagnostics/LogPointsManager.ts
+++ b/src/diagnostics/LogPointsManager.ts
@@ -68,6 +68,12 @@ class DebugSessionManager {
         logpointsCollection.updateTextEditorDecroration();
     }
 
+    public removeLogpointGlyphsFromDocument(documentUri: vscode.Uri): void {
+        const logpointsCollection = this.getLogpointCollectionForDocument(documentUri);
+        logpointsCollection.clearRegistry();
+        logpointsCollection.updateTextEditorDecroration();
+    }
+
     public async addLogpoint(scriptId: string, lineNumber: number, columnNumber: number): Promise<ILogpoint> {
         const expression = await vscode.window.showInputBox({
             ignoreFocusOut: true,
@@ -125,6 +131,7 @@ export class LogPointsManager extends vscode.Disposable {
         });
 
         vscode.debug.onDidTerminateDebugSession((debugSession) => {
+            this.onDebugSessionClose(debugSession);
             delete this._debugSessionManagerMapping[debugSession.id];
         });
 
@@ -158,7 +165,7 @@ export class LogPointsManager extends vscode.Disposable {
         const debugSessionManager = this._debugSessionManagerMapping[params.vscodeDebugSessionId];
 
         if (!debugSessionManager) {
-            vscode.window.showErrorMessage(`Cannot find debug session with id ${params.vscodeDebugSessionId}`);
+            vscode.window.showErrorMessage(`The debug session associated with this file has expired. Please close this file and open it again from LOADED SCRIPTS explorer of an active logpoints session. More details can be found here - http://aka.ms/logpoints#setting-up-logpoints`);
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


src/diagnostics/LogPointsManager.ts
```diff
@@ -158,7 +165,7 @@ export class LogPointsManager extends vscode.Disposable {
         const debugSessionManager = this._debugSessionManagerMapping[params.vscodeDebugSessionId];
 
         if (!debugSessionManager) {
-            vscode.window.showErrorMessage(`Cannot find debug session with id ${params.vscodeDebugSessionId}`);
+            vscode.window.showErrorMessage(`The debug session associated with this file has expired. Please close this file and open it again from LOADED SCRIPTS explorer of an active debug session.`);
```
dhanvikapila: Please update the message to remove 'debug' from it. Say 'active logpoints session.' instead.

dhanvikapila: Please update the message to remove 'debug' from it. Say 'active logpoints session.' instead.

src/diagnostics/LogPointsManager.ts
```diff
@@ -198,6 +205,23 @@ export class LogPointsManager extends vscode.Disposable {
...
+
+        if (!debugSessionManager) {
+            // If debugSessionManager does not exist, it might have been handled already.
+            return;
+        }
+
+        if (!vscode.window.activeTextEditor) {
+            return;
+        }
+        const documentUri = vscode.window.activeTextEditor.document.uri;
```
dhanvikapila: Is there a scenario where the document is undefined or uri is undefined?

src/diagnostics/LogPointsManager.ts
```diff
@@ -68,6 +68,12 @@ class DebugSessionManager {
         logpointsCollection.updateTextEditorDecroration();
     }
 
+    public removeLogpointGlyphsFromDocument(documentUri: vscode.Uri): void {
+        const logpointsCollection = this.getLogpointCollectionForDocument(documentUri);
```
dhanvikapila: This could be undefined if the input was bad?

src/diagnostics/LogPointsManager.ts
changsi-an: No. An active text editor has to have a document uri.

src/diagnostics/LogPointsManager.ts
changsi-an: The `getLogpointCollectionForDocument()` is responsible for creating a collection if it is not there. So `logpointsCollection ` can be guaranteed to non-undefined.

src/diagnostics/LogPointsManager.ts
changsi-an: Done. A link to documentation is added too.

src/diagnostics/LogPointsManager.ts
```diff
@@ -158,7 +165,7 @@ export class LogPointsManager extends vscode.Disposable {
         const debugSessionManager = this._debugSessionManagerMapping[params.vscodeDebugSessionId];
 
         if (!debugSessionManager) {
-            vscode.window.showErrorMessage(`Cannot find debug session with id ${params.vscodeDebugSessionId}`);
+            vscode.window.showErrorMessage(`The debug session associated with this file has expired. Please close this file and open it again from LOADED SCRIPTS explorer of an active logpoints session. More details can be found here - https://aka.ms/logpoints`);
```
mike-kaufman: Great to see the hyper-link!  Can you deep link to an appropriate faq section?  e.g., http://aka.ms/logpoints#setting-up-logpoints links through to the setup section, since it looks like the forwarding will carry through the stuff after the `#` through on the redirect, so if there's a `#faq` section, or a `#errors` section, that we can send ppl there.

changsi-an: Sounds good. Done.

PR #214, opened 2017-12-05
Add another enabled image name and don't test against image tags.

changsi-an:


src/diagnostics/logPointsSessionWizard.ts
```diff
@@ -69,14 +69,6 @@ export class LogPointsSessionAttach extends WizardBase {
         this.steps.push(new StartDebugAdapterStep(this));
         this.output.show();
     }
-
-    protected onExecuteError(error: Error): void {
```
changsi-an: Due to some recent refactoring. The method is not used anymore.

src/diagnostics/logPointsSessionWizard.ts
```diff
@@ -104,11 +96,16 @@ class EligibilityCheck extends WizardStep {
...
+        const imageName = fullImageName.split(':')[0];
+        const enabledImages = vscode.workspace.getConfiguration('appService').get<string[]>('enabledDockerImages') || [];
+        const enabledImagesTagless = enabledImages.map((name) => {
+            return name.split(':')[0].toLocaleLowerCase();
+        });
 
-        if ('docker' !== framework.toLocaleLowerCase() || enabledImages.indexOf(imageName.toLocaleLowerCase()) === -1) {
-            throw new Error(`Please use one of the supported docker image. The ${framework}|${imageName} combination is not supported`);
+        if ('docker' !== framework.toLocaleLowerCase() || enabledImagesTagless.indexOf(imageName.toLocaleLowerCase()) === -1) {
+            throw new Error(`Please use one of the supported docker image. ${imageName} is not supported for starting a Logpoints session. More details can be found here - https://aka.ms/logpoints`);
```
changsi-an: Currently , there is no easy way to make the [https://aka.ms/logpoints](https://aka.ms/logpoints) link into a organic link in the error message... Let me know if this is not the case.

mike-kaufman: note sure I understand the comment.  You should be able to link to sections of the doc that `http://aka.ms/logpoints/`, e.g. http://aka.ms/logpoints/#sending-feedback links into the #sending-feedback section.

changsi-an: I meant that it appears a plain text, not a clickable link.

PR #215, opened 2017-12-06
Show web app state in explorer

### LINKED ISSUE #78: Show the service's state
We allow stopping, starting, etc. But we can't tell if the service is running or stopped right now. Add a state indicator.

image

whiskeyjay:
Fix #78

### DIFF
--- a/src/explorer/DeploymentSlotTreeItem.ts
+++ b/src/explorer/DeploymentSlotTreeItem.ts
@@ -18,7 +18,8 @@ export class DeploymentSlotTreeItem extends SiteTreeItem {
     }
 
     public get label(): string {
-        return `${this.siteWrapper.slotName} (${this.site.state})`;
+        const state = this.site.state;
+        return `${this.siteWrapper.slotName} ${state && state.toLowerCase() !== 'running' ? '(' + state + ')' : ''}`;
     }
 
     public get iconPath(): { light: string, dark: string } {
--- a/src/explorer/DeploymentSlotTreeItem.ts
+++ b/src/explorer/DeploymentSlotTreeItem.ts
@@ -17,11 +17,6 @@ export class DeploymentSlotTreeItem extends SiteTreeItem {
         super(site);
     }
 
-    public get label(): string {
-        const state = this.site.state;
-        return `${this.siteWrapper.slotName} ${state && state.toLowerCase() !== 'running' ? '(' + state + ')' : ''}`;
-    }
-
     public get iconPath(): { light: string, dark: string } {
         return {
             light: path.join(__filename, '..', '..', '..', '..', 'resources', 'light', 'DeploymentSlot_color.svg'),
--- a/src/explorer/DeploymentSlotTreeItem.ts
+++ b/src/explorer/DeploymentSlotTreeItem.ts
@@ -18,7 +18,7 @@ export class DeploymentSlotTreeItem extends SiteTreeItem {
     }
 
     public get label(): string {
-        return this.siteWrapper.slotName;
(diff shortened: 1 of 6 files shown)


src/explorer/SiteTreeItem.ts
```diff
@@ -110,6 +120,12 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
     public async editScmType(client: WebSiteManagementClient): Promise<string> {
         return await this.siteWrapper.editScmType(client);
     }
+
+    private async reloadSite(client: WebSiteManagementClient): Promise<void> {
```
nturinski: I think the start/stop functions use the Azure API to confirm that the web app has started/stopped.  Do we need to reacquire the resource through the API to update (Running)/(Stopped) and reconstruct the SiteWrapper to confirm the state?

src/explorer/DeploymentSlotTreeItem.ts
```diff
@@ -18,7 +18,7 @@ export class DeploymentSlotTreeItem extends SiteTreeItem {
     }
 
     public get label(): string {
-        return this.siteWrapper.slotName;
+        return `${this.siteWrapper.slotName} (${this.site.state})`;
```
nturinski: nit: Having all the web apps say (Running) makes the explorer look very cluttered.  What do you think about only displaying (Stopped) and omitting (Running) for apps that are on?

src/explorer/SiteTreeItem.ts
whiskeyjay: Good point! I'll make a bit of adjustment...

src/explorer/SiteTreeItem.ts
```diff
@@ -49,6 +49,16 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
+        await this.siteWrapper.start(client);
+        await this.reloadSite(client);
+    }
+
+    public async stop(client: WebSiteManagementClient): Promise<void> {
+        await this.siteWrapper.stop(client);
+        await this.reloadSite(client);
+    }
+
     public browse(): void {
```
nturinski: Currently if a web app is Stopped and then you "Restart" the app through the context menu, the state doesn't refresh unless you manually refresh the node.

whiskeyjay: Yep, I'll handle that case.

src/explorer/SiteTreeItem.ts
EricJizbaMSFT: No need to reload the site or reconstruct the sitewrapper. Just use the existing `siteWrapper.getState` method

src/extension.ts
```diff
@@ -83,7 +83,8 @@ export function activate(context: vscode.ExtensionContext): void {
         const siteType = util.isSiteDeploymentSlot(node.treeItem.site) ? 'Deployment Slot' : 'Web App';
         outputChannel.show();
         outputChannel.appendLine(`Starting ${siteType} "${node.treeItem.site.name}"...`);
-        await node.treeItem.siteWrapper.start(nodeUtils.getWebSiteClient(node));
+        await node.treeItem.start(nodeUtils.getWebSiteClient(node));
+        node.treeDataProvider.refresh(node, false);
```
EricJizbaMSFT: I thought this didn't work in the released version of VS Code - don't you still need to refresh the parent?

In any case, I would prefer if you used `node.refresh()` and then we only have to change it in one place (aka [here](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/src/treeDataProvider/AzureNode.ts#L62)) once VS Code releases the tree fix

PR #216, opened 2017-12-06
Check streaming log availability before starting logpoints session.

changsi-an:
Some refactoring to move step classes out of the wizard file. So we get rid of the "Too many classes" lint error


src/diagnostics/wizardSteps/CheckLogStreamAvailability.ts
```diff
@@ -0,0 +1,44 @@
+import * as vscode from 'vscode';
```
changsi-an: This is the only new file added. The rest of this PR is merely refactoring.

PR #219, opened 2017-12-07
show set logpoint errors

digeff:
fixes some issues preventing errors from showing when failing to set a logpoint

### DIFF
--- a/src/diagnostics/LogPointsManager.ts
+++ b/src/diagnostics/LogPointsManager.ts
@@ -88,7 +88,7 @@ class DebugSessionManager {
         }
 
         const result: ISetLogpointResponse = await this._debugSession.customRequest("setLogpoint", { scriptId, lineNumber, columnNumber, expression });
-        if (result.error) {
+        if (result.error !== undefined && result.error !== null) {
             vscode.window.showErrorMessage("Cannot set logpoint successfully. Please refer to DEBUG CONSOLE for details.");
             appserviceUtil.getOutputChannel().show();
             appserviceUtil.getOutputChannel().appendLine(`Cannot set logpoint successfully, the error is ${result.error}.`);
--- a/src/diagnostics/logPointsClient.ts
+++ b/src/diagnostics/logPointsClient.ts
@@ -78,15 +78,15 @@ export class KuduLogPointsDebuggerClient extends LogPointsDebuggerClientBase imp
 
     public async startSession(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User): Promise<CommandRunResult<IStartSessionResponse>> {
         await this.uploadSshClient(siteName, affinityValue, publishCredential);
-        return this.makeCallAndLogException<IStartSessionResponse>(siteName, affinityValue, publishCredential, "curl -X POST http://localhost:32923/debugger/session");
+        return this.makeCallAndLogException<IStartSessionResponse>(siteName, affinityValue, publishCredential, "curl -s -S -X POST http://localhost:32923/debugger/session");
     }
 
     public closeSession(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User, data: ICloseSessionRequest): Promise<CommandRunResult<ICloseSessionResponse>> {
-        return this.makeCallAndLogException<IStartSessionResponse>(siteName, affinityValue, publishCredential, `curl -X DELETE http://localhost:32923/debugger/session/${data.sessionId}`);
+        return this.makeCallAndLogException<IStartSessionResponse>(siteName, affinityValue, publishCredential, `curl -s -S -X DELETE http://localhost:32923/debugger/session/${data.sessionId}`);
     }
 
     public enumerateProcesses(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User): Promise<CommandRunResult<IEnumerateProcessResponse>> {
-        return this.makeCallAndLogException<IEnumerateProcessResponse>(siteName, affinityValue, publishCredential, "curl -X GET http://localhost:32923/os/processes?applicationType=Node.js");
+        return this.makeCallAndLogException<IEnumerateProcessResponse>(siteName, affinityValue, publishCredential, "curl -s -S -X GET http://localhost:32923/os/processes?applicationType=Node.js");
     }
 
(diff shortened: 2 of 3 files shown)


src/diagnostics/LogPointsManager.ts
```diff
@@ -88,7 +88,7 @@ class DebugSessionManager {
         }
 
         const result: ISetLogpointResponse = await this._debugSession.customRequest("setLogpoint", { scriptId, lineNumber, columnNumber, expression });
-        if (result.error) {
+        if (result.error !== undefined) {
```
changsi-an: Why? Couldn't `result.error` be `null`?

src/diagnostics/logPointsSessionWizard.ts
```diff
@@ -345,7 +346,11 @@ class StartDebugAdapterStep extends WizardStep {
             instanceId: this._wizard.selectedInstance.name,
             sessionId: this._wizard.sessionId,
             debugId: this._wizard.debuggerId
-        });
+        };
+
+        this._wizard.writeline(`Starting debugging session with ${JSON.stringify(logpointsConfiguration)}`);
```
changsi-an: This seems to be temporary debug code, it is not useful to end user.

src/diagnostics/logPointsClient.ts
```diff
@@ -78,15 +78,15 @@ export class KuduLogPointsDebuggerClient extends LogPointsDebuggerClientBase imp
 
     public async startSession(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User): Promise<CommandRunResult<IStartSessionResponse>> {
         await this.uploadSshClient(siteName, affinityValue, publishCredential);
-        return this.makeCallAndLogException<IStartSessionResponse>(siteName, affinityValue, publishCredential, "curl -X POST http://localhost:32923/debugger/session");
+        return this.makeCallAndLogException<IStartSessionResponse>(siteName, affinityValue, publishCredential, "curl -s -S -X POST http://localhost:32923/debugger/session");
```
changsi-an: What does "-s" (lower case) try to achieve? 
```
-s, --silent        Silent mode (don't output anything)
```

Wouldn't it mean no output can be seen?

src/diagnostics/LogPointsManager.ts
changsi-an: Why? Couldn't result.error be null?

src/diagnostics/logPointsSessionWizard.ts
```diff
@@ -345,7 +346,11 @@ class StartDebugAdapterStep extends WizardStep {
             instanceId: this._wizard.selectedInstance.name,
             sessionId: this._wizard.sessionId,
             debugId: this._wizard.debuggerId
-        });
+        };
```
changsi-an: This seems to be temporary debug code, `instanceId`, `sessionId` and `debugId` are not useful to end user and have never been in the picture.

src/diagnostics/structs/CommandRunResult.ts
```diff
@@ -26,6 +26,9 @@ export class CommandRunResult<ResponseType extends { error?: {}, data?: {} }> {
             const outputJson = JSON.parse(this.output);
 
             this._stdout = outputJson.stdout;
+            if (!this.error) {
```
changsi-an: I don't seem to find usage of this field within this PR. I guess it's good to have, but can you explain your intention?

digeff: If I do this, then the stderr then appears on the Output window.

src/diagnostics/logPointsSessionWizard.ts
digeff: I thought this could be useful to have for troubleshooting/diagnostic issues. If you don't think this is useful for the users I can put these flags under some #if _DEBUG_

src/diagnostics/logPointsClient.ts
digeff: We tested these parameters with Mike, and we get to see the stderr without any timing information, which is what we want to show the user

src/diagnostics/LogPointsManager.ts
digeff: I'm not familiar with this code. Do you know if result.error can be null, and what is the semantics of that?

src/diagnostics/structs/CommandRunResult.ts
changsi-an: We don't really need this either way. The same information will be shown in the DEBUG CONSOLE when debugger starts. It's the payload of the first debug protocol message. Suggest remove altogether.

src/diagnostics/logPointsClient.ts
changsi-an: As long as you tested and saw that we have the correct JSON response from stdout. I am fine with it.

src/diagnostics/LogPointsManager.ts
changsi-an: It depends on what the server responds with. It's purely modeling the REST API response payload. From the perspective of VS Code, there is not point of making any assumption. The semantic is that as long as result.error is non-empty, it's an error. 

Suggest revert this change.

src/diagnostics/logPointsSessionWizard.ts
changsi-an: We don't really need this either way. The same information will be shown in the DEBUG CONSOLE when debugger starts. It's the payload of the first debug protocol message. Suggest remove altogether.

src/diagnostics/LogPointsManager.ts
digeff: I added this because the request was failing with error = "" so we weren't seeing any error messages because the if evaluated to false.
Changing this to: if (result.error !== undefined && result.error !== null) {

changsi-an: Sounds good.

src/diagnostics/logPointsSessionWizard.ts
digeff: You are right, removing this change...

src/diagnostics/logPointsClient.ts
digeff: Yes, this show what we want

Issue #195, opened 2017-11-28
initCommand() should support multiple parameters.


changsi-an:
[initCommand](https://github.com/Microsoft/vscode-azureappservice/blob/bc69cfbeade8a095413c212d8e1a372c1d99e9a2/src/extension.ts#L260) should allow `callback` to receive multiple parameters, currently it only pass over `context`.

changsi-an:
It's fixed in PR #220

PR #220, opened 2017-12-07
Allow callback of initCommand to accept multiple parameters.

changsi-an:
Addresses #195

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -257,18 +257,21 @@ export function activate(context: vscode.ExtensionContext): void {
         logPointsManager.toggleLogpoint(uri);
     });
 
-    context.subscriptions.push(vscode.commands.registerCommand('diagnostics.LogPoints.OpenScript', openScript));
+    initCommand(context, 'diagnostics.LogPoints.OpenScript', openScript);
 }
 
 // tslint:disable-next-line:no-empty
 export function deactivate(): void {
 }
 
-function initCommand<T>(extensionContext: vscode.ExtensionContext, commandId: string, callback: (context?: T) => void): void {
-    initAsyncCommand(extensionContext, commandId, async (context?: T) => callback(context));
+// tslint:disable-next-line:no-any
+function initCommand<T>(extensionContext: vscode.ExtensionContext, commandId: string, callback: (context?: T, ...rest: any[]) => void): void {
+    // tslint:disable-next-line:no-any
+    initAsyncCommand(extensionContext, commandId, async (context?: T, ...remaining: any[]) => callback.apply(null, [context].concat(remaining)));
 }
 
-function initAsyncCommand<T>(extensionContext: vscode.ExtensionContext, commandId: string, callback: (context?: T) => Promise<void>): void {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -280,7 +283,7 @@ function initAsyncCommand<T>(extensionContext: vscode.ExtensionContext, commandI
             if (args.length === 0) {
                 await callback();
             } else {
-                await callback(<T>args[0]);
+                await callback.apply(null, args);
```
EricJizbaMSFT: I'm not sure setting `this` to `null` will work in all cases? What if the callback has a different value they want for `this`?

src/extension.ts
```diff
@@ -257,18 +257,21 @@ export function activate(context: vscode.ExtensionContext): void {
...
 }
 
 // tslint:disable-next-line:no-empty
 export function deactivate(): void {
 }
 
-function initCommand<T>(extensionContext: vscode.ExtensionContext, commandId: string, callback: (context?: T) => void): void {
-    initAsyncCommand(extensionContext, commandId, async (context?: T) => callback(context));
+// tslint:disable-next-line:no-any
+function initCommand<T>(extensionContext: vscode.ExtensionContext, commandId: string, callback: (context?: T, ...rest: any[]) => void): void {
```
EricJizbaMSFT: It doesn't seem like the type parameter `T` is providing much value anymore and is just adding complexity

What if we just simplify this to be:
```
function initCommand(context: vscode.ExtensionContext, commandId: string, callback: (...args: any[]) => any)
```

EricJizbaMSFT: And then you can just call `await callback(...args);` down below

The cosmos repo has the simplified signature of `initCommand` and `initAsyncCommand`
https://github.com/Microsoft/vscode-cosmosdb/blob/master/src/extension.ts#L145

changsi-an: I thought about that too. Not sure if there would be any special purpose of type T, I just wanted to minimize the impact to the codebase as much as possible.

Will change it in another PR.

src/extension.ts
changsi-an: I guess that the caller has to remember to use arrow functions all the time. Otherwise, is that anyway to maintain that?

mike-kaufman: `callback.apply(null, args)` is equivalent to `callback(...args)`.  If caller wants a specific this context when the callback is invoked, they need to do the binding in the callback.  e.g.,

```
callback = function() {
    myFunc.apply(myThis, args);
}
```

changsi-an: Yeah, was wondering how TS transcompile would avoid this problem.

I guess after this change, it requires all callers to remember always use arrow functions as callback to preserve `this`. Does it make sense?

mike-kaufman: Yeah, either arrow functions, or explicitly bound functions via [Function's bind() method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind), or equivalent as above.

PR #223, opened 2017-12-08
Address some comments from last PR

changsi-an:
Addresses some PR concerns and improve the signature of init*Command() methods.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -268,14 +268,15 @@ export function deactivate(): void {
 }
 
 // tslint:disable-next-line:no-any
-function initCommand<T>(extensionContext: vscode.ExtensionContext, commandId: string, callback: (context?: T, ...rest: any[]) => void): void {
+function initCommand(extensionContext: vscode.ExtensionContext, commandId: string, callback: (...args: any[]) => void): void {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -268,14 +268,15 @@ export function deactivate(): void {
 }
 
 // tslint:disable-next-line:no-any
-function initCommand<T>(extensionContext: vscode.ExtensionContext, commandId: string, callback: (context?: T, ...rest: any[]) => void): void {
+function initCommand(extensionContext: vscode.ExtensionContext, commandId: string, callback: (...args: any[]) => void): void {
     // tslint:disable-next-line:no-any
-    initAsyncCommand(extensionContext, commandId, async (context?: T, ...remaining: any[]) => callback.apply(null, [context].concat(remaining)));
+    initAsyncCommand(extensionContext, commandId, async (...args: any[]) => Promise.resolve(callback(...args)));
```
EricJizbaMSFT: Since this is an async arrow function - does it need to be `await Promise.resolve(`?

I don't think this would functionally affect the user, but we track how long commands take in our telemetry

changsi-an: That's a valid point. Will remove.

PR #225, opened 2017-12-09
Send an activation ping before we start th logpoints session.

changsi-an:
Also add a timeout guard in case the Kudu/ARM calls takes too long due to inactivity of the app service.

### DIFF
--- a/src/utils/logpointsUtil.ts
+++ b/src/utils/logpointsUtil.ts
@@ -1,0 +1,24 @@
+import { setTimeout } from "timers";
+
+export const DEFAULT_TIMEOUT = 20000;
+// tslint:disable-next-line:export-name
+// tslint:disable-next-line:no-any
+export function callWithTimeout<T>(proc: (...args: any[]) => Promise<T>, timeout: number): Promise<T> {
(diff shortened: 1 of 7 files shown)


src/utils/logpointsUtil.ts
```diff
@@ -0,0 +1,23 @@
+import { setTimeout } from "timers";
+
+// tslint:disable-next-line:export-name
+// tslint:disable-next-line:no-any
+export function callWithTimeout<T>(proc: (...args: any[]) => Promise<T>, timeout: number = 60000): Promise<T> {
```
msft-cwells: Are we really defaulting to a 60-second timeout?!  That seems like way too long.

changsi-an: Will change to 20 secs.

mike-kaufman: Alternatively, since no timeout value is correct for all uses, you coul djust remove the default and force callers to be explicit w/ what they thing.

changsi-an: @mike-kaufman Sounds good. Done that too.

PR #228, opened 2017-12-11
Fix some bugs in error display logic around setting/removing logpoints.

changsi-an:


src/diagnostics/wizardSteps/EligibilityCheck.ts
```diff
@@ -27,11 +27,16 @@ export class EligibilityCheck extends WizardStep {
             throw new Error('Cannot read "linuxFxVersion"');
         }
 
-        const [framework, imageName] = linuxFxVersion.split('|');
-        const enabledImages = vscode.workspace.getConfiguration('appService').get<string[]>('enabledDockerImages');
-
-        if ('docker' !== framework.toLocaleLowerCase() || enabledImages.indexOf(imageName.toLocaleLowerCase()) === -1) {
-            throw new Error(`Please use one of the supported docker image. The ${framework}|${imageName} combination is not supported`);
+        const [framework, fullImageName] = linuxFxVersion.split('|');
```
changsi-an: This is a regression due to merge. Just added it back here.

src/diagnostics/LogPointsManager.ts
```diff
@@ -92,9 +93,12 @@ class DebugSessionManager {
 
         const result: ISetLogpointResponse = await this._debugSession.customRequest("setLogpoint", { scriptId, lineNumber, columnNumber, expression });
         if (result.error !== undefined && result.error !== null) {
-            vscode.window.showErrorMessage("Cannot set logpoint successfully. Please refer to DEBUG CONSOLE for details.");
+            vscode.window.showErrorMessage("Cannot set logpoint successfully. Please refer to DEBUG CONSOLE and https://aka.ms/logpoints for details.");
```
digeff: Should we use a more specific link?

src/diagnostics/LogPointsManager.ts
```diff
@@ -92,9 +93,12 @@ class DebugSessionManager {
...
             appserviceUtil.getOutputChannel().show();
-            appserviceUtil.getOutputChannel().appendLine(`Cannot set logpoint successfully, the error is ${result.error}.`);
+            const errorMessage = result.error.message === '' ?
+                'Cannot set logpoint. Please refer to- https://aka.ms/logpoints for more details.' :
```
digeff: is that - on purpose?

src/diagnostics/LogPointsManager.ts
```diff
@@ -92,9 +93,12 @@ class DebugSessionManager {
...
+                `Cannot set logpoint, the error is "${result.error.message}". Please refer to- https://aka.ms/logpoints for more details.`;
```
digeff: Most of the texts are the same. Maybe you could make it so the if is only about ? "" : `the error is "${result.error.message}`

src/diagnostics/LogPointsManager.ts
changsi-an: No. Will remove

src/diagnostics/LogPointsManager.ts
```diff
@@ -195,9 +209,11 @@ export class LogPointsManager extends vscode.Disposable {
...
         let logpoint = debugSessionManager.getLogpointAtLocation(uri, line);
         if (logpoint) {
-            debugSessionManager.removeLogpoint(logpoint);
-            debugSessionManager.unregisterLogpoint(uri, logpoint);
-            this._outputChannel.appendLine(`Removed logpoint at line ${logpoint.line} in ${params.path}`);
+            const succeed = await debugSessionManager.removeLogpoint(logpoint);
+            if (succeed) {
+                debugSessionManager.unregisterLogpoint(uri, logpoint);
+                this._outputChannel.appendLine(`Removed logpoint at line ${logpoint.line} in ${params.path}`);
+            }
```
digeff: What happens if we don't succeed?

src/diagnostics/LogPointsManager.ts
changsi-an: I prefer not. Who knows whether it will become something else in the translation. Having a logic like that might invalidate translation in some other language, which does not follow the same lexical order as English.

src/diagnostics/wizardSteps/EligibilityCheck.ts
```diff
@@ -27,11 +27,16 @@ export class EligibilityCheck extends WizardStep {
...
+        // Remove the 'tag' portion of the image name.
+        const imageName = fullImageName.split(':')[0];
```
digeff: Is it possible to the array to be empty here?

src/diagnostics/wizardSteps/EligibilityCheck.ts
```diff
@@ -27,11 +27,16 @@ export class EligibilityCheck extends WizardStep {
...
+        const enabledImages = vscode.workspace.getConfiguration('appService').get<string[]>('enabledDockerImages') || [];
+        const enabledImagesTagless = enabledImages.map((name) => {
+            return name.split(':')[0].toLocaleLowerCase();
```
digeff: Is it possible for the array to be empty here?

src/diagnostics/LogPointsManager.ts
changsi-an: We will not unregister the logpoint. The glyph will still show. And the error message is emitted from within the `removeLogpoint()` method.

src/diagnostics/wizardSteps/EligibilityCheck.ts
changsi-an: I don't think so. Above we tested site.kind to filter out windows, so when here it is linux, the linuxFxVersion will have an value we need.

src/diagnostics/LogPointsManager.ts
```diff
@@ -104,8 +108,18 @@ class DebugSessionManager {
...
     }
 
-    public async removeLogpoint(logpoint: ILogpoint): Promise<void> {
-        await this._debugSession.customRequest("removeLogpoint", logpoint.id);
+    public async removeLogpoint(logpoint: ILogpoint): Promise<boolean> {
+        const result: IRemoveLogpointResponse = await this._debugSession.customRequest("removeLogpoint", logpoint.id);
+        if (result.error !== undefined && result.error !== null) {
+            vscode.window.showErrorMessage("Cannot remove logpoint successfully. Please refer to DEBUG CONSOLE and https://aka.ms/logpoints for details.");
+            appserviceUtil.getOutputChannel().show();
+            const errorMessage = result.error.message === '' ?
```
digeff: This is pretty similar to the addLogpoint logic. Would it make sense to put the error checking and UI logic on a higher layer? and leave the business logic here?

src/diagnostics/wizardSteps/EligibilityCheck.ts
changsi-an: I don't think so. Even `''.split(':')` gets `['']`.

src/diagnostics/LogPointsManager.ts
digeff: We do this kind of things pretty often in VS (having placeholders in the strings {0}) and it's not a problem for localization...

src/diagnostics/LogPointsManager.ts
changsi-an: The text message is different. Again, if we over-unify it, it might cause trouble for later translation. I prefer keep language template intact.

src/diagnostics/LogPointsManager.ts
changsi-an: Will change to `https://aka.ms/logpoints#setting-logpoints`

src/diagnostics/LogPointsManager.ts
changsi-an: It is not a placeholder for a data item, it's a part of the sentence. Because it includes the "the error is" part.

src/diagnostics/LogPointsManager.ts
```diff
@@ -105,7 +108,13 @@ class DebugSessionManager {
     }
 
     public async removeLogpoint(logpoint: ILogpoint): Promise<void> {
-        await this._debugSession.customRequest("removeLogpoint", logpoint.id);
+        const result: IRemoveLogpointResponse = await this._debugSession.customRequest("removeLogpoint", logpoint.id);
+        if (result.error !== undefined && result.error !== null) {
+            const errorMessage = result.error.message === '' ?
+                'Cannot remove logpoint. Please refer to https://aka.ms/logpoints for more details.' :
+                `Cannot remove logpoint, the error is "${result.error.message}". Please refer to https://aka.ms/logpoints for more details.`;
+            throw new Error(errorMessage);
```
changsi-an: The error dumping statements are removed in favor of throwing an exception. Because the underlying exception capturing logic will both showing a error prompt and logging a message to the output pane. It is done before sending out the telemetry.

PR #235, opened 2017-12-13
Enables File Nodes and remote editing (with feature flag)

### LINKED ISSUE #45: Enable FilesNodes
Add FilesNodes back when the node can tell a full end-to-end user story.

nturinski:
Fixes #45 

Enables FIleNodes if remoteEditing flag is enabled.


src/explorer/editors/FileEditor.ts
```diff
@@ -0,0 +1,40 @@
...
+        return node.treeItem.label;
+    }
+
+    async getData(node: IAzureNode<FileTreeItem>): Promise<string> {
+        return await node.treeItem.kuduClient.getFile(node.treeItem.path);
+    }
+
+    async getSize(node: IAzureNode<FileTreeItem>): Promise<number> {
+        // this is not implemented for Azure App Services
+        return node ? 0 : 0;
```
EricJizbaMSFT: Why not just `return 0`?

src/explorer/editors/FileEditor.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    }
+
+    async updateData(node: IAzureNode<FileTreeItem>): Promise<string> {
+        const localPath: string = vscode.window.activeTextEditor.document.fileName;
```
EricJizbaMSFT: Make sure you test this stuff on different OS's since you're dealing with file paths

src/explorer/editors/fileEditor.ts
```diff
@@ -0,0 +1,40 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { BaseEditor, IAzureNode } from 'vscode-azureextensionui';
```
EricJizbaMSFT: You have two `fileEditor.ts` files. Gotta figure out that casing stuff haha

src/extension.ts
```diff
@@ -281,13 +286,61 @@ export function activate(context: vscode.ExtensionContext): void {
         await logPointsManager.toggleLogpoint(uri);
     });
 
+    initAsyncCommand(context, 'appService.editFile', async (node: IAzureNode<FileTreeItem>) => {
```
EricJizbaMSFT: This isn't really `editFile` - it's `showFile`

src/extension.ts
```diff
@@ -281,13 +286,61 @@ export function activate(context: vscode.ExtensionContext): void {
...
+        await fileEditor.showEditor(node);
+    });
+
+    initEvent(context, 'appService.fileEditor.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, (doc: vscode.TextDocument) => fileEditor.onDidSaveTextDocument(context.globalState, doc));
+    context.subscriptions.push(vscode.commands.registerCommand('diagnostics.LogPoints.OpenScript', openScript));
```
EricJizbaMSFT: Was this a bad merge? I don't understand why you're adding the logpoints command

src/extension.ts
```diff
@@ -281,13 +286,61 @@ export function activate(context: vscode.ExtensionContext): void {
...
 // tslint:disable-next-line:no-empty
 export function deactivate(): void {
 }
 
+function initEvent<T>(context: vscode.ExtensionContext, eventId: string, event: vscode.Event<T>, callback: (...args: any[]) => any) {
+    context.subscriptions.push(event(wrapAsyncCallback(eventId, (...args: any[]) => Promise.resolve(callback(...args)))));
+}
+
+function wrapAsyncCallback(callbackId, callback: (...args: any[]) => Promise<any>): (...args: any[]) => Promise<any> {
+    return async (...args: any[]) => {
```
EricJizbaMSFT: wrapAsyncCallback has a lot of copied code from initAsyncCommand. initAsyncCommand is supposed to call wrapAsyncCallback

src/KuduClient.ts
```diff
@@ -25,6 +25,25 @@ export class KuduClient {
         });
     }
 
+    getFile(path: string): Promise<string> {
```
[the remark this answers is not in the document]
EricJizbaMSFT: FYI - I'm okay with you addressing this comment in a separate PR. It might not be a trivial change and I don't want to delay this for a long time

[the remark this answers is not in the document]
nturinski: It seems like it might take a bit of refactoring to accomplish this.  I made an issue for this and will take care of it after release.

https://github.com/Microsoft/vscode-azureappservice/issues/237

src/explorer/editors/FileEditor.ts
nturinski: I had to do this due to this ts error: [ts] 'node' is declared but its value is never read.

I looked into optional abstract functions in Typescript, but I don't think that's currently a thing.

src/extension.ts
nturinski: Me neither.  I'm assuming it must have happened on the merge.  Very odd.

src/explorer/editors/FileEditor.ts
EricJizbaMSFT: Change the name of node to _node and it will ignore the error

src/extension.ts
```diff
@@ -83,7 +88,7 @@ export function activate(context: vscode.ExtensionContext): void {
         const siteType = util.isSiteDeploymentSlot(node.treeItem.site) ? 'Deployment Slot' : 'Web App';
         outputChannel.show();
         outputChannel.appendLine(`Starting ${siteType} "${node.treeItem.site.name}"...`);
-        await node.treeItem.start(nodeUtils.getWebSiteClient(node));
+        await node.treeItem.siteWrapper.start(nodeUtils.getWebSiteClient(node));
```
EricJizbaMSFT: Looks like another merge issue. This should still be node.treeItem.start

Same with the stop command below

src/extension.ts
```diff
@@ -282,33 +287,52 @@ export function activate(context: vscode.ExtensionContext): void {
...
-        const properties: { [key: string]: string; } = {};
-        const output = util.getOutputChannel();
+        let properties: { [key: string]: string; } = {};
         properties.result = 'Succeeded';
-        let errorData: ErrorData | undefined;
+        let errorData: ErrorData | undefined = null;
+        const output = util.getOutputChannel();
 
         try {
             if (args.length === 0) {
```
EricJizbaMSFT: Remove this if/else case and just do `await callback(...args);`

(The only reason we had stuff like this was because we had initAsyncCommand<T> and <T> applied to the first argument. Like I said yesterday, the <T> stuff just complicated it with little benefit)

src/extension.ts
```diff
@@ -282,33 +287,52 @@ export function activate(context: vscode.ExtensionContext): void {
...
+
+function wrapAsyncCallback(commandId, callback: (...args: any[]) => Promise<any>): (...args: any[]) => Promise<any> {
+    return async (...args: any[]) => {
         const start = Date.now();
-        const properties: { [key: string]: string; } = {};
-        const output = util.getOutputChannel();
+        let properties: { [key: string]: string; } = {};
         properties.result = 'Succeeded';
-        let errorData: ErrorData | undefined;
+        let errorData: ErrorData | undefined = null;
```
EricJizbaMSFT: Don't set this to null

src/extension.ts
```diff
@@ -282,33 +287,52 @@ export function activate(context: vscode.ExtensionContext): void {
...
+    context.subscriptions.push(vscode.commands.registerCommand(commandId, wrapAsyncCallback(commandId, callback)));
+}
+
+
+function wrapAsyncCallback(commandId, callback: (...args: any[]) => Promise<any>): (...args: any[]) => Promise<any> {
+    return async (...args: any[]) => {
         const start = Date.now();
-        const properties: { [key: string]: string; } = {};
-        const output = util.getOutputChannel();
+        let properties: { [key: string]: string; } = {};
```
EricJizbaMSFT: Leave this as const

src/explorer/editors/FileEditor.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import * as vscode from 'vscode';
+import * as util from '../../util';
+import { BaseEditor, IAzureNode } from 'vscode-azureextensionui';
+import { FileTreeItem } from '../FileTreeItem';
+import { KuduClient } from '../../KuduClient';
+import { getOutputChannel } from '../../util';
+import { nodeUtils } from '../../utils/nodeUtils';
+
+
```
EricJizbaMSFT: Do you still have "tslint.autoFixOnSave" on? I feel like you've got several extra whitespace lines in your PR, and that should've automatically fixed it for you

src/extension.ts
nturinski: Actually, this one I changed because I didn't think that the start/stop methods lived in the treeItem itself, but in the siteWrapper.

whiskeyjay: I added start/stop to the treeItem, which calls siteWrapper and updates the status text in node label.

src/explorer/editors/FileEditor.ts
nturinski: It's on, but the config setting isn't recognized.  I'll figure this out later.

PR #236, opened 2017-12-13
Automatically open logstream of a site/deploymentSlot when launch debug session.

changsi-an:
and switch to the streaming log pane after debug session is launched.

Also fixes a bug that picking a deployment slot did not work correctly with opening a streaming log logic.

Also fixes a bug that error message from attaching a process was not displayed correctly.

Also Refactors the code that extracting the scm sub-domain name.

### DIFF
--- a/src/diagnostics/wizardSteps/PromptSlotSelection.ts
+++ b/src/diagnostics/wizardSteps/PromptSlotSelection.ts
@@ -1,8 +1,9 @@
 import { Site } from 'azure-arm-website/lib/models';
 import * as vscode from 'vscode';
-import { UserCancelledError } from 'vscode-azureextensionui';
+import { IAzureNode, IAzureParentNode, IAzureTreeItem, UserCancelledError } from 'vscode-azureextensionui';
+import { DeploymentSlotsTreeItem } from '../../explorer/DeploymentSlotsTreeItem';
+import { SiteTreeItem } from '../../explorer/SiteTreeItem';
 import * as util from '../../util';
-import { callWithTimeout, DEFAULT_TIMEOUT } from '../../utils/logpointsUtil';
 import { WizardStep } from '../../wizard';
 import { LogPointsSessionWizard } from '../LogPointsSessionWizard';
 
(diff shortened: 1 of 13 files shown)


src/diagnostics/wizardSteps/PromptSlotSelection.ts
```diff
@@ -49,22 +51,24 @@ export class PromptSlotSelection extends WizardStep {
...
-        return allDeploymentSlots.filter((slot) => {
-            return slot.repositorySiteName === this.site.name;
-        });
+    private async getDeploymentSlotsTreeItems(): Promise<SiteTreeItem[]> {
+        const appServiceTreeItem = this._wizard.uiTreeItem.treeItem;
+        const result = await appServiceTreeItem.loadMoreChildren(<IAzureParentNode<SiteTreeItem>>this._wizard.uiTreeItem);
+
+        let deploymentSlotsCategoryNode: DeploymentSlotsTreeItem;
+        if (result && result.length > 0) {
+            deploymentSlotsCategoryNode = <DeploymentSlotsTreeItem>result[0];
```
msft-cwells: You know from the structure of the tree nodes that there will always be only a single category node at this level, and no other children?

changsi-an: Not exactly, there are usually 3 categories. The first one is the deployment slots category, the other two are WebWorkers and AppSettings category.

I can do an `instanceof` style test to loop the three(or any).

src/diagnostics/wizardSteps/PromptSlotSelection.ts
```diff
@@ -49,22 +51,24 @@ export class PromptSlotSelection extends WizardStep {
...
-            () => {
-                return client.webApps.listByResourceGroup(this.site.resourceGroup, { includeSlots: true });
-            },
-            DEFAULT_TIMEOUT);
-        return allDeploymentSlots.filter((slot) => {
-            return slot.repositorySiteName === this.site.name;
-        });
+    private async getDeploymentSlotsTreeItems(): Promise<SiteTreeItem[]> {
+        const appServiceTreeItem = this._wizard.uiTreeItem.treeItem;
+        const result = await appServiceTreeItem.loadMoreChildren(<IAzureParentNode<SiteTreeItem>>this._wizard.uiTreeItem);
```
EricJizbaMSFT: You should never call `treeItem.loadMoreChildren` directly. Instead change the type of `this._wizard.uiTreeItem` to `IAzureParentNode<SiteTreeItem>` and call `this._wizard.uiTreeItem.getCachedChildren()`

If you call 'loadMoreChildren', then you only get the "next batch" to be displayed in the tree. However, if you call 'getCachedChildren', you get all of the nodes that have been loaded so far. The operation is potentially much faster if the user has already opened the slots in the tree

src/diagnostics/wizardSteps/PromptSlotSelection.ts
EricJizbaMSFT: Yeah I wouldn't rely on it being the first node. We actually just added a fix recently to sort nodes in the tree alphabetically which would probably break you

src/util.ts
```diff
@@ -18,6 +18,10 @@ export function extractSiteName(site: WebSiteModels.Site): string {
     return isSiteDeploymentSlot(site) ? site.name.substring(0, site.name.lastIndexOf('/')) : site.name;
 }
 
+export function extractSiteScmSubDomainName(site: WebSiteModels.Site): string {
```
EricJizbaMSFT: You don't have to change it for this PR, but just FYI we're trying to move away from using `site` directly and instead using the `SiteWrapper` class from our appservice npm package. In this case, I believe `SiteWrapper.appName` duplicates this logic https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/SiteWrapper.ts#L56

(We created that npm package because the Azure Functions extension uses a lot of the same code and we didn't want it to be duplicated)

src/diagnostics/wizardSteps/PromptSlotSelection.ts
EricJizbaMSFT: NOTE: This isn't really an issue in the case of slots because it always load all of the slots at once (aka, slots doesn't support a continuation token or load more of any sort). It's just the principle of the thing haha

changsi-an: I see. Will change to `getCachedChildren()`. 

To your other FYI for AzureTreeDataProvider.showNodePicker. I really tried that and felt it useful, but unfortunately it does not give a desired behavior, in that, whatever context value I use, it always prompts user to select something starting from selecting an subscription, then a App Service, then a deployment slot. This won't work for us, because user asks to initiate a command from an already selected AppService node, we would like her to select a underneath deployment slot directly. 

Also our selection logic requires the list to include the app service itself. From my observation, when it gets to the deployment slots level, it only displays the real deployment slots, and will leave the parent app service out.

src/util.ts
changsi-an: Sounds good. Will add an issue to change to using SiteWrapper later.

src/diagnostics/wizardSteps/PromptSlotSelection.ts
EricJizbaMSFT: I will keep your scenario in mind for 'showNodePicker' and let you know if/when we add support for it. If you want that behavior, then it's probably useful to other people as well

changsi-an: Thanks @EricJizbaMSFT Much appreciated

whiskeyjay: I don't think we need to worry about a large number of children in deployment slots. I think the slot limit on standard websites is 4.

src/util.ts
changsi-an: Added #240

Issue #230, opened 2017-12-12
Upon starting a session, also send account information to differentiate identity


changsi-an:
The `startSession` method should send with account information.

PR #239, opened 2017-12-14
When start a new session, send the account identity as well.

changsi-an:
This addresses issue #230

### DIFF
--- a/src/diagnostics/logPointsClient.ts
+++ b/src/diagnostics/logPointsClient.ts
@@ -19,12 +19,13 @@ import { IRemoveLogpointRequest } from './structs/IRemoveLogpointRequest';
 import { IRemoveLogpointResponse } from './structs/IRemoveLogpointResponse';
 import { ISetLogpointRequest } from './structs/ISetLogpointRequest';
 import { ISetLogpointResponse } from './structs/ISetLogpointResponse';
+import { IStartSessionRequest } from './structs/IStartSessionRequest';
 import { IStartSessionResponse } from './structs/IStartSessionResponse';
 
 export interface ILogPointsDebuggerClient {
     call<ResponseType>(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User, command: string): Promise<CommandRunResult<ResponseType>>;
 
-    startSession(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User): Promise<CommandRunResult<IStartSessionResponse>>;
+    startSession(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User, data: IStartSessionRequest): Promise<CommandRunResult<IStartSessionResponse>>;
 
     closeSession(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User, data: ICloseSessionRequest): Promise<CommandRunResult<ICloseSessionResponse>>;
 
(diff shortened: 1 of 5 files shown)


src/azure-account.api.d.ts
```diff
@@ -0,0 +1,33 @@
+import { Event } from 'vscode';
```
mike-kaufman: hmmm, there's not an existing `typings` folder for things like this?  Would be great to hand this back to owners of azure account so they can embed correct typings in their module.

src/diagnostics/wizardSteps/GetUnoccupiedInstance.ts
```diff
@@ -73,4 +76,25 @@ export class GetUnoccupiedInstance extends WizardStep {
...
+        if (!sessions || sessions.length <= 0) {
+            throw new Error('Cannot find active sessions');
+        }
+
+        const filteredSessions = sessions.filter((session) => {
+            return session.tenantId === currentTenentId;
+        });
+
+        if (!filteredSessions || filteredSessions.length !== 1) {
+            throw new Error(`Cannot locate unique session with tenent ID ${currentTenentId}`);
```
mike-kaufman: this error message may be more useful if you included the count of sessions found.  Also, I think that filter will return an empty array if nothing matches. (you should dbl check), but if so you can simplify things a bit.

src/diagnostics/logPointsClient.ts
```diff
@@ -76,9 +77,15 @@ export class KuduLogPointsDebuggerClient extends LogPointsDebuggerClientBase imp
         return buf.toString('base64');
     }
 
-    public async startSession(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User): Promise<CommandRunResult<IStartSessionResponse>> {
+    public async startSession(siteName: string, affinityValue: string, publishCredential: WebSiteModels.User, data: IStartSessionRequest): Promise<CommandRunResult<IStartSessionResponse>> {
         await this.uploadSshClient(siteName, affinityValue, publishCredential);
-        return this.makeCallAndLogException<IStartSessionResponse>(siteName, affinityValue, publishCredential, "curl -s -S -X POST http://localhost:32923/debugger/session");
+        const jsonPayloadBase64 = KuduLogPointsDebuggerClient.base64Encode({
+            username: data.username
```
mike-kaufman: what is `data.username`?  Is this the user that you're auth'd against azure?

EricJizbaMSFT: Are you  sure it's not supposed to be publishCredential.userName?

src/azure-account.api.d.ts
EricJizbaMSFT: I agree it's annoying. However, we mitigated this issue by only having 'azure-account.api.d.ts' in our UI package.

Please remove this file and use IAzureNode.userId if you really need the azure userId
https://github.com/Microsoft/vscode-azuretools/pull/49

EricJizbaMSFT: I've published the 0.4.1 version of the ui package. You should be able to get userId now.

changsi-an: Sounds good. I will rework the PR. Thanks for taking care of that.

src/diagnostics/logPointsClient.ts
changsi-an: `data.username` is the email address of the active Azure tenant identity.

I don't remember what publishCredential.userName is like. It will at most be the same as userId, so userId is always better to use for our purpose. We want to have a unique identifier for the current logged in user to differentiate "who" initiated the debug session.

src/diagnostics/wizardSteps/GetUnoccupiedInstance.ts
changsi-an: Thanks for the suggestion. I suppose this whole chunk of work will be removed during the rework.

changsi-an:
Reworked on this PR, bumped the UI library to 0.4.1 and use `userId` from `IAzureNode<>` directly.

PR #245, opened 2017-12-15
Allow users to choose a folder to deploy

### LINKED ISSUE #209: Allow choosing a folder to zip deploy
It would be amazing if we could prompt the user to select the folder to zip deploy. Often in web apps we deploy a `dist` or `public` folder. Let's allow the user to select the folder to deploy so we don't deploy everything, including source code.

cc @timheuer as FYI

whiskeyjay:
Fix #209 

Tried a few options, seems this is the simplest solution.


src/util.ts
```diff
@@ -88,17 +89,31 @@ export async function showWorkspaceFoldersQuickPick(placeHolderString: string):
             }
         }) :
         [];
-    // VS Code will handle [] by alerting user there are no workspaces opened
+    folderQuickPickItems.splice(0, 0, browse);
```
EricJizbaMSFT: nit: This could be `folderQuickPickItems.unshift(browse);`

PR #269, opened 2018-01-10
Ignore KuduClient, lint DeploymentSwap and ErrorData

nturinski:


src/ErrorData.ts
```diff
@@ -10,7 +10,9 @@ export class ErrorData {
     constructor(error: any) {
         if (error instanceof Error) {
             try {
+                // tslint:disable-next-line:no-unsafe-any
```
EricJizbaMSFT: You can remove this entire file if you want. I should've removed it in this commit, but I forgot: https://github.com/Microsoft/vscode-azureappservice/commit/3e1c8a665840537a5a0de7512ecccbc863d2e055

src/SwapStep.ts
```diff
@@ -0,0 +1,78 @@
...
+import { SubscriptionModels } from 'azure-arm-resource';
+// tslint:disable-next-line:no-require-imports
+import WebSiteManagementClient = require('azure-arm-website');
+import { IAzureNode, UserCancelledError } from 'vscode-azureextensionui';
+import { DeploymentSlotTreeItem } from './explorer/DeploymentSlotTreeItem';
+import { IQuickPickItemWithData } from './util';
+import { nodeUtils } from './utils/nodeUtils';
+import { WizardBase, WizardStep } from './wizard';
+
+export class SwapStep extends WizardStep {
```
EricJizbaMSFT: nit: If we're moving this to another file, I would prefer to name it 'SwapSlotStep' (otherwise it's too generic)

PR #270, opened 2018-01-11
Remove "tools" from the name.

fiveisprime:
"Tools" is synonymous with "extension" and ends up creating some redundancy in
the docs: "Azure App Service Tools extension," so I removed it. This also
creates consistency between the other extensions.


nturinski:
For what it's worth, I agree with @fiveisprime

PR #273, opened 2018-01-12
Error handling for not found files

### LINKED ISSUE #251: File editing need better error handling
1. Create a web app.
2. Deploy to production.
3. Create a new slot.
4. Expand the Files node from production.
5. Swap slot with production.
6. Because of #250, the Files node doesn't auto-refresh, now from the Files node, try to edit a file that doesn't exists in the production (previously the new slot).

You'll get this error:
image

And this error message in the editor (which is not the content of the file because the file doesn't exist):
image

When you save, the file will be uploaded as a new file.

The correct behavior should be a message saying what the server error response say, in this case "'/home/site/wwwroot/cseries.azcli' not found."

nturinski:
Fixes #251


nturinski:
@whiskeyjay Currently the shared KuduClient does get files correctly.  We have to update the swagger.json to generate code that can handle strings.  It currently tries to automatically JSON parse the response which works for folders, but not files.

PR #275, opened 2018-01-18
Fixing some minor bugs in logpoints

mike-kaufman:
Two bug fixes:
1.  In some cases, `site.kind` can have a value like `app,linux,container`.  Updated the regular expression to be more tolerant of this pattern.
2.  In some cases,  `vscode.workspace.workspaceFolders` can be undefined.  Check for undefined now before indexing into array.  `startDebugging(...)` API is tolerant of undefined values for the folder.

### DIFF
--- a/src/diagnostics/wizardSteps/StartDebugAdapterStep.ts
+++ b/src/diagnostics/wizardSteps/StartDebugAdapterStep.ts
@@ -20,7 +20,8 @@ export class StartDebugAdapterStep extends WizardStep {
 
             vscode.commands.executeCommand('workbench.view.debug');
         });
-        await vscode.debug.startDebugging(vscode.workspace.workspaceFolders[0], {
+        const folder = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0] : undefined;
+        await vscode.debug.startDebugging(folder, {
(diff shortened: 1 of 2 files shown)


src/diagnostics/wizardSteps/StartDebugAdapterStep.ts
```diff
@@ -20,7 +20,8 @@ export class StartDebugAdapterStep extends WizardStep {
 
             vscode.commands.executeCommand('workbench.view.debug');
         });
-        await vscode.debug.startDebugging(vscode.workspace.workspaceFolders[0], {
+        const folder = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0] : undefined;
```
changsi-an: See https://github.com/Microsoft/vscode/issues/37559 for more details. 

The code was that way because 'undefined' didn't work before. If you found out  'undefined' can work now, it means the above bug fix has been released. Then we should always use ‘undefined’. 

`vscode.workspace.workspaceFolders[0]` is arbitrary and does not really have any significance in our business.

mike-kaufman: makes sense.  Fixed now.

PR #277, opened 2018-01-24
Update changelog for v6.0

nturinski:


CHANGELOG.md
```diff
@@ -1,6 +1,21 @@
...
+### Added
+- Auto-browse after web app and slot creation
+- Link to a great tutorial to get started in the README
+- View and edit a deployment slot's files (feature flag required)
+
+### Removed
+- Tools from the name
+
+### Fixed
+- [Minor bugs fixed](https://github.com/Microsoft/vscode-azureappservice/pull/275)
```
EricJizbaMSFT: Use the same link format as below

CHANGELOG.md
```diff
@@ -1,6 +1,21 @@
...
+
+
```
EricJizbaMSFT: Remove extra blank line

CHANGELOG.md
```diff
@@ -1,6 +1,21 @@
...
+## 0.6.0 - 2018-01-25
+Happy New Year, everybody!  Welcome to 2018!
+
+### Added
+- Auto-browse after web app and slot creation
+- Link to a great tutorial to get started in the README
+- View and edit a deployment slot's files (feature flag required)
+
+### Removed
+- Tools from the name
```
EricJizbaMSFT: Not worth mentioning IMO

CHANGELOG.md
nturinski: Well, these bug fixes are for logpoints.  They were not filed as issues for the milestone v.6.0 and don't show up

EricJizbaMSFT: If the bugs aren't important enough to create an issue for on github, then it's not worth linking to the PR.

In any case, there's always the chance that we have issues we forgot to create or tag properly on github. If you use the consistent format from below, then we can always add issues later and tag them as this milestone. With your current link, we can't make changes without doing another release

PR #285, opened 2018-01-31
Add build/release badges to README

whiskeyjay:


nturinski:
After we merge #286, the FolderNodes error about not being able to map on void should be fixed,

PR #286, opened 2018-01-31
Refactor Folder and Files Nodes to use VS Code Azure Kudu

nturinski:


src/explorer/FolderTreeItem.ts
```diff
@@ -38,8 +38,9 @@ export class FolderTreeItem implements IAzureParentTreeItem {
     public async loadMoreChildren(node: IAzureNode<FolderTreeItem>): Promise<IAzureTreeItem[]> {
         const webAppClient: WebSiteManagementClient = nodeUtils.getWebSiteClient(node);
         const kuduClient: KuduClient = await this.siteWrapper.getKuduClient(webAppClient);
-
-        const fileList = await kuduClient.vfs.getItem(this.folderPath);
+        const httpResponse: kuduIncomingMessage = <kuduIncomingMessage>(await kuduClient.vfs.getItemWithHttpOperationResponse(this.folderPath)).response;
```
whiskeyjay: (This probably should have been asked a while ago) Can this line throw exceptions? If an exception was thrown, can it be handled more gracefully than just letting VSC catching it?

nturinski: Kudu returns an error that bubbles up and is displayed to the user if the API call fails.  Do you think it'd be helpful to create an error node if it throws an exception?

whiskeyjay: It should be fine if there is a way for users to retry - either by clicking the load more again or explicitly a retry button on a message bar. I am a bit afraid of raw error from network being displayed on UI won't be very readable.

src/explorer/editors/FileEditor.ts
```diff
@@ -39,13 +44,10 @@ export class FileEditor extends BaseEditor<IAzureNode<FileTreeItem>> {
...
-        const publishingCredential = await node.treeItem.siteWrapper.getWebAppPublishCredential(webAppClient);
-        const kuduClient = new KuduClient(node.treeItem.siteWrapper.appName, publishingCredential.publishingUserName, publishingCredential.publishingPassword);
-
-        const localPath: string = vscode.window.activeTextEditor.document.fileName;
+        const kuduClient: KuduClient = await node.treeItem.siteWrapper.getKuduClient(webAppClient);
+        const localFile: Readable = fs.createReadStream(vscode.window.activeTextEditor.document.uri.fsPath);
         const destPath: string = node.treeItem.path;
-
-        await kuduClient.uploadFile(localPath, destPath);
+        await kuduClient.vfs.putItem(localFile, destPath, { customHeaders: { ['If-Match']: node.treeItem.etag } });
```
EricJizbaMSFT: Just curious - did we ignore the etag stuff before this change?

src/explorer/FileTreeItem.ts
```diff
@@ -10,11 +10,18 @@ export class FileTreeItem implements IAzureTreeItem {
...
     public readonly commandId: string = 'appService.showFile';
+    private _etag: string | undefined = undefined; // cannot create _etag on creation due to Kudu API calls
 
     constructor(readonly siteWrapper: SiteWrapper, readonly label: string, readonly path: string) {
     }
 
     public get id(): string {
         return `${this.siteWrapper.id}/File`;
     }
+    public get etag(): string {
```
EricJizbaMSFT: nit: Why not just do `public etag: string | undefined;`

src/KuduClient.ts
```diff
@@ -5,9 +5,11 @@
 
 import * as kuduApi from 'kudu-api';
 import * as request from 'request';
+import { IncomingMessage } from 'http';
 
 export type kuduFile = { mime: string, name: string, path: string };
 export type webJob = { name: string, Message: string };
+export type kuduIncomingMessage = IncomingMessage & { body: string };
 
 /*
```
EricJizbaMSFT: Please delete the file-related methods to prevent new code from leveraging it

src/explorer/editors/FileEditor.ts
nturinski: The old API ignored the etag by default.

src/explorer/FileTreeItem.ts
nturinski: Because I have to set the etag in a different class, I thought that it made sense to have getter/setter methods.  Should I just get rid of the private property?

src/KuduClient.ts
nturinski: When you say file-related methods, do you mean the type definitions that I created?

PR #289, opened 2018-01-31
Sort web apps and deployment slots after creation

### LINKED ISSUE #260: The new created web app doesn't list in alphabetical order unless manual refreshing
Repro Step:

1. Click 'Create New Web App' to create a new web app.
2. Check the order of the new created web app.

Expect:
The new created web app lists in alphabetical order.

Actual:
The new created web app lists at the first. Please see the attachment.

Attachment:
compare

nturinski:
Fixes #260 
Refresh tree to sort nodes after app/slot creation

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -134,7 +134,7 @@ export function activate(context: vscode.ExtensionContext): void {
         }
 
         const createdApp = <IAzureNode<WebAppTreeItem>>await node.createChild();
-
+        tree.refresh();
         // prompt user to deploy to newly created web app
         const yesButton: vscode.MessageItem = { title: 'Yes' };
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -134,7 +134,7 @@ export function activate(context: vscode.ExtensionContext): void {
         }
 
         const createdApp = <IAzureNode<WebAppTreeItem>>await node.createChild();
-
+        await tree.refresh();
```
EricJizbaMSFT: It would be better to fix the UI package so that it inserts the node in the sorted position after creation finishes. Calling refresh like this will re-fetch _all_ resources from Azure, which is unnecessary

Plus fixing it directly in the UI package will fix it for everyone :)

PR #292, opened 2018-02-01
Enable tslint on entire project

### LINKED ISSUE #288: Tslint is ignoring many files
Our lint script in package.json is using 'src/*.ts' which doesn't include subfolders of 'src'.

The script can be simplified from this:
```
tslint --project tsconfig.json src/*.ts -e src/*.d.ts --type-check -t verbose
```
to this:
```
tslint --project tsconfig.json -t verbose
```

nturinski:
Fixes #288


src/explorer/SiteTreeItem.ts
```diff
@@ -101,9 +102,7 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         this._logStream = kuduClient.getLogStream().on('data', chunk => {
             this._logStreamOutputChannel.append(chunk.toString());
         }).on('error', err => {
-            util.sendTelemetry('ConnectToLogStreamError', { name: err.name, message: err.message });
-            this._logStreamOutputChannel.appendLine('Error connecting to log-streaming service:');
-            this._logStreamOutputChannel.appendLine(err.message);
```
PrashanthCorp: Are these covered by `throw(err)` you just added?

src/extension.ts
```diff
@@ -22,14 +23,16 @@ import { LoadedScriptsProvider, openScript } from './explorer/loadedScriptsExplo
...
 import * as util from "./util";
 import { nodeUtils } from './utils/nodeUtils';
 
 // tslint:disable-next-line:export-name
 // tslint:disable-next-line:max-func-body-length
 export function activate(context: vscode.ExtensionContext): void {
-    context.subscriptions.push(new Reporter(context));
+    const packageInfo: IPackageInfo = getPackageInfo(context);
+    const reporter = new TelemetryReporter(packageInfo.name, packageInfo.version, packageInfo.aiKey);
+    context.subscriptions.push(reporter);
```
PrashanthCorp: Just curious, was this change mainly because the Reporter class was used just once?

src/explorer/FolderTreeItem.ts
```diff
@@ -40,7 +40,8 @@ export class FolderTreeItem implements IAzureParentTreeItem {
         const kuduClient: KuduClient = await this.siteWrapper.getKuduClient(webAppClient);
         const httpResponse: kuduIncomingMessage = <kuduIncomingMessage>(await kuduClient.vfs.getItemWithHttpOperationResponse(this.folderPath)).response;
         // response contains a body with a JSON parseable string
-        const fileList = JSON.parse(httpResponse.body);
+        // tslint:disable-next-line:no-unsafe-any
+        const fileList: kuduFile[] = JSON.parse(httpResponse.body);
```
EricJizbaMSFT: nit: I prefer this rather than disabling the rule:
```
const fileList: kuduFile[] = JSON.parse(<string>httpResponse.body);
```

src/explorer/SiteTreeItem.ts
```diff
@@ -68,7 +68,8 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         const isSsl = this.site.hostNameSslStates.findIndex(value =>
             value.name === defaultHostName && value.sslState === `Enabled`);
         // tslint:disable-next-line:no-http-string
-        const uri = `${isSsl ? 'https://' : 'http://'}${defaultHostName}`;
+        const uri = `${isSsl !== -1 ? 'https://' : 'http://'}${defaultHostName}`;
```
EricJizbaMSFT: Rather than change this line, you could change isSsl from 'findIndex' to 'some', which returns a boolean and "Determines whether the specified callback function returns true for any element of an array."

src/telemetry/reporter.ts
```diff
@@ -3,30 +3,16 @@
...
-
-interface IPackageInfo {
+export interface IPackageInfo {
     name: string;
     version: string;
     aiKey: string;
 }
 
-function getPackageInfo(context: vscode.ExtensionContext): IPackageInfo | undefined {
+export function getPackageInfo(context: vscode.ExtensionContext): IPackageInfo | undefined {
```
EricJizbaMSFT: nit: I'd rename this file and move it somewhere else. It doesn't fit under 'telemetry/reporter' anymore

src/explorer/SiteTreeItem.ts
```diff
@@ -101,9 +102,7 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         this._logStream = kuduClient.getLogStream().on('data', chunk => {
             this._logStreamOutputChannel.append(chunk.toString());
         }).on('error', err => {
-            util.sendTelemetry('ConnectToLogStreamError', { name: err.name, message: err.message });
-            this._logStreamOutputChannel.appendLine('Error connecting to log-streaming service:');
-            this._logStreamOutputChannel.appendLine(err.message);
+            throw (err);
```
EricJizbaMSFT: I'm not sure how to test out this code (aka I don't know how to repro an error). @whiskeyjay can you help determine if this would be okay?

Mainly - I'm concerned that `.on('error', err => {` might send the error in chunks and the user might not see the whole error if we just throw the error immediately like this

src/extension.ts
nturinski: Yeah, we used to have the wrapper handle telemetry, but now it's handled through the AzureActionHandler when we register the commands.

src/explorer/FolderTreeItem.ts
nturinski: tslint is still yelling at me :(

This fixes it though:
```
<kuduFile[]>JSON.parse(httpResponse.body)
```

I think it's because the return type of JSON.parse() is any which was causing the warning.

src/explorer/SiteTreeItem.ts
whiskeyjay: You can disconnect your network (or disable the network adapter from Device Manager) then an error will occur if the streaming has already started:

Connecting to log-streaming service...
2018-02-01T23:53:26  Welcome, you are now connected to log-streaming service.
2018-02-01T23:54:26  No new trace in the past 1 min(s).
Error connecting to log-streaming service:
read ECONNRESET
Disconnected from log-streaming service.

I don't think one single error will be sent in chunks.

whiskeyjay: I suppose with "throw", errors will be shown in a message bar instead? When users connect to log stream, their eyes are on the output panel, usually at the bottom part of the window, looking for logs; why make them having to look across the screen for errors?

nturinski: Discussed with Jay-- the outputChannel for log-streaming is a different channel than where errors are normally displayed for App Service.  I think it's worth preserving the old experience.

src/explorer/SiteTreeItem.ts
nturinski: Not exactly.  We were trying to streamline the error handling, but see the thread above to see why that won't work :(

nturinski:
I'm in the processing of leverage azuretools's logstreaming, but the work is more significant than I thought it would be so I reverted logstream back for now, and will refactor it in a new PR.  Just wanted to get tslint in here finally.

PR #304, opened 2018-02-09
Remove ids from files and folders nodes

### LINKED ISSUE #303: Files Node isn't supported for a project deployed through GitHub
Should we gray it out if we detect that a Web App was deployed through GitHub?

nturinski:
This was causing an issue and there is no reason for these nodes to have ids as they do not change state or are opened in the portal.

This will break until azuretools is updated to where id properties are optional.

Fixes #303


nturinski:
I'll just wait until we publish to merge it then

PR #305, opened 2018-02-09
Remove open on create and move to deploy

### LINKED ISSUE #302: Remove web apps and slots opening a browser upon creation
It is very distracting to have a browser be opened every time I create a web app or slot.  Granted, I doubt a user will be creating as many slots/apps as I am, but it's generally pretty annoying to have something open up on its own a la an ad.

This also distracts users from the Deploy to App pop-up that happens upon creation which is currently the most streamlined entry point to deploying as there is only one prompt to select a project.

I understand that we want to make the user aware that the app is created, but there is an output message that is shown and a "Creating..." node in the explorer.

nturinski:
Fixes #302

### DIFF
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -98,6 +98,11 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         return await this.siteWrapper.editScmType(client);
     }
 
+    public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: OutputChannel, configureSecitonName: string, confirmDeployment: boolean = false): Promise<void> {
+        await this.siteWrapper.deploy(fsPath, client, outputChannel, configureSecitonName, confirmDeployment);
+        this.browse();
+    }
+
     private createLabel(state: string): string {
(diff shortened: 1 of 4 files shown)


src/explorer/SiteTreeItem.ts
```diff
@@ -98,6 +98,11 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         return await this.siteWrapper.editScmType(client);
     }
 
+    public async deploy(fsPath: string, client: WebSiteManagementClient, outputChannel: OutputChannel, configureSecitonName: string, confirmDeployment: boolean = false): Promise<void> {
```
EricJizbaMSFT: confirmDeployment should default to true

StephenWeatherford: nit: spell configureSecitonName

StephenWeatherford: suggestion: Some people argue that having booleans defaulting to true is a bad idea (non-intuitive), in which case you could change the flag to dontConfirmDeployment=false or some such.

nturinski:
Closing this for now.  I don't want to change this until we can reliably know when the website has been successfully built and is servable.

PR #315, opened 2018-03-01
Fix for deployment slot open in portal

### LINKED ISSUE #311: Fail to open in portal for the created deployment slots
Repro Step:
1. Expand one app and right click 'Deployment Slots', then select 'Create New Deployment Slot'.
2. Enter a valid unique name.
3. Choose one configuration source for the deployment slot.
4. Right click the new deployment slot and select 'Open In Portal'.
5. Check whether the deployment slot can be opened in portal successfully or not.

Expect:
The deployment slot can be opened in portal successfully.

Actual:
Fail to open in portal for the new created deployment slot.
error
More Info:
The same deployment slot can be opened in portal successfully in Cloud Explorer.

nturinski:
Fixes #311

Depends on this PR: https://github.com/Microsoft/vscode-azuretools/pull/94

<del>When creating slots, the name availability check included the app's name as well, basically padding the slot name with n characters.  This would prevent users from creating slots with 64 characters.</del>

### DIFF
--- a/src/explorer/DeploymentSlotsTreeItem.ts
+++ b/src/explorer/DeploymentSlotsTreeItem.ts
@@ -98,7 +98,7 @@ export class DeploymentSlotsTreeItem implements IAzureParentTreeItem {
                     return `The slot name "${value}" is not available.`;
                 }
 
-                const nameAvailability: ResourceNameAvailability = await nodeUtils.getWebSiteClient(node).checkNameAvailability(`$${util.extractSiteName(node.treeItem.site)}-{value}`, 'Slot');
+                const nameAvailability: ResourceNameAvailability = await nodeUtils.getWebSiteClient(node).checkNameAvailability(`${util.extractSiteName(node.treeItem.site)}-${value}`, 'Slot');
                 if (!nameAvailability.nameAvailable) {
                     return nameAvailability.message;
                 }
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -81,7 +81,8 @@ export function activate(context: vscode.ExtensionContext): void {
             node = <IAzureNode<WebAppTreeItem>>await tree.showNodePicker(WebAppTreeItem.contextValue);
         }
 
-        node.openInPortal();
+        node.treeItem.contextValue === 'deploymentSlot' ? node.openInPortal(`${node.parent.parent.treeItem.id}/slots/${node.treeItem.label}`) : node.openInPortal();
```
EricJizbaMSFT: I believe `node.treeItem.id` would work (which is different than `node.id`)? If that doesn't work - use siteWrapper.id somehow rather than manually constructing the id

src/explorer/DeploymentSlotsTreeItem.ts
```diff
@@ -98,7 +98,7 @@ export class DeploymentSlotsTreeItem implements IAzureParentTreeItem {
                     return `The slot name "${value}" is not available.`;
                 }
 
-                const nameAvailability: ResourceNameAvailability = await nodeUtils.getWebSiteClient(node).checkNameAvailability(`${util.extractSiteName(node.treeItem.site)}-${value}`, 'Slot');
+                const nameAvailability: ResourceNameAvailability = await nodeUtils.getWebSiteClient(node).checkNameAvailability(`${value}`, 'Slot');
```
EricJizbaMSFT: This might fix the problem that you described, but it breaks duplicate names. Even though I already have 'erijiz1', I'm not getting an error:
screen shot 2018-03-01 at 12 18 36 pm
screen shot 2018-03-01 at 12 18 30 pm

PR #318, opened 2018-03-02
Fix zip glob/ignore settings

EricJizbaMSFT:
1. The ignore setting can be a string or string array
1. The scope should be 'resource' so that these work in multi-root workspace scenarios

Requires this PR to fully work: https://github.com/Microsoft/vscode-azuretools/pull/98


EricJizbaMSFT:
FYI @fiveisprime your wish has been fulfilled!

PR #320, opened 2018-03-06
Add min validation time and add elapsed seconds to telemetry

StephenWeatherford:
Reworking of last PR - stop trying to report success/failure to the user, and just log simple telemetry for status codes after a deployment e.g.:

[{"code":200,"elapsed":4},{"code":200,"elapsed":9},{"code":200,"elapsed":14},{"code":200,"elapsed":19},{"code":200,"elapsed":24},{"code":200,"elapsed":29},{"code":200,"elapsed":34},{"code":200,"elapsed":40},{"code":200,"elapsed":45},{"code":200,"elapsed":50},{"code":200,"elapsed":55},{"code":200,"elapsed":60},{"code":200,"elapsed":65}] 

We can use this to help efforts to report success/failure in the future.

### DIFF
--- a/src/validateWebSite.ts
+++ b/src/validateWebSite.ts
@@ -9,8 +9,7 @@ import * as requestP from 'request-promise';
 import { URL } from 'url';
 import { isNumber } from 'util';
 import { OutputChannel } from 'vscode';
-import { SiteWrapper } from "vscode-azureappservice";
-import { callWithTelemetryAndErrorHandling, IActionContext, parseError } from 'vscode-azureextensionui';
+import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
 import TelemetryReporter from "vscode-extension-telemetry";
 import { SiteTreeItem } from './explorer/SiteTreeItem';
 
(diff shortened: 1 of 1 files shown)


src/validateWebSite.ts
```diff
@@ -24,78 +23,57 @@ type WebError = {
...
+
+        properties.statusCodes = JSON.stringify(statusCodes);
     });
 }
 
-function log(siteWrapper: SiteWrapper, outputChannel: OutputChannel, message: string): void {
-    outputChannel.appendLine(`${(new Date()).toLocaleTimeString()} ${siteWrapper.appName}: ${message}`);
+async function delay(delayMs: number): Promise<void> {
+    // tslint:disable-next-line:no-string-based-set-timeout // tslint false positive
+    await new Promise<void>((resolve, _reject): void => { setTimeout(resolve, delayMs); });
```
EricJizbaMSFT: nit: You don't have to disable the rule if you specify the type of resolve. Also you don't have to include 'reject' at all:
```
await new Promise((resolve: () => void): void => { setTimeout(resolve, delayMs); })
```

src/validateWebSite.ts
```diff
@@ -24,78 +23,57 @@ type WebError = {
...
-                // a status code 200-300 indicates success
                 currentStatusCode = response.statusCode;
-                isSuccess = response.statusCode >= 200 && response.statusCode < 400;
-                currentStatusMessage = response.statusMessage;
-            } catch (e) {
-                const error = <WebError>e;
-                const response = error.response || {};
-                isSuccess = false;
+            } catch (error) {
+                const response = (<WebError>error).response || {};
```
EricJizbaMSFT: nit: Check if error is null or undefined before getting error.response

src/validateWebSite.ts
```diff
@@ -24,78 +23,57 @@ type WebError = {
...
             resolveWithFullResponse: true
         };
         let currentStatusCode: number = 0;
-        let currentStatusMessage: string = '';
-
-        properties.statusCodes = '';
-        properties.timedOut = 'false';
-
-        log(siteWrapper, outputChannel, `Checking for successful response from ${uri}.`);
+        const statusCodes: { code: number, elapsed: number, reported?: boolean }[] = [];
```
EricJizbaMSFT: nit: remove 'reported' from the type since its never used?

src/validateWebSite.ts
StephenWeatherford: It doesn't seem reasonable that I should have to check on every catch if someone threw null or undefined. You disagree?

EricJizbaMSFT: Well most of the time people use parseError(), which automatically checks for them

And yes - I'm very strict/cautious about error handling and telemetry. I just think the amount of work we have to do is worth the amount of frustration from a user if they got a null ref exception because of telemetry

But on the other hand I recognize its a bit of a philosophical debate since you're unlikely to get null or undefined here. So its not something I'd block a PR for

PR #322, opened 2018-03-06
Deployment script prompt

### LINKED ISSUE #224: Improve performance of deploy
I'll have to think through the experience for making this work, but I want to make sure the need is captured. This will ensure Windows users are able to deploy compiled modules.

Things to consider:

* An app could be one of many runtimes (and may be custom)
* How can we know to alert a user that they need to enable npm install on deploy?
* At what point in the process should this be configured?

See also [vscode-azuretools#7](https://github.com/Microsoft/vscode-azuretools/issues/7) and the [Kudu docs](https://github.com/projectkudu/kudu/wiki/Configurable-settings#enabledisable-build-actions)

nturinski:
Fixes #224


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azureappservice",
     "displayName": "Azure App Service",
     "description": "An Azure App Service management extension for Visual Studio Code.",
-    "version": "0.5.1",
+    "version": "0.6.0",
```
EricJizbaMSFT: nit: I prefer you bump the version in the same PR that you do the changelog

package.json
```diff
@@ -490,18 +490,19 @@
...
-                    "default": "",
+                    "default": [],
                     "description": "Defines which files in the workspace to ignore for Zip deploy. This applies to Zip deploy only, has no effect on other deployment methods."
                 },
                 "appService.showRemoteFiles": {
                     "type": "boolean",
                     "default": false,
                     "description": "Show or hide the File Nodes of a Web App.  Caution: Manually editing your Web App's files could cause unexpected behavior."
                 },
-                "appService.showSavePrompt": {
```
EricJizbaMSFT: Don't delete the 'showSavePrompt' setting

package.json
```diff
@@ -490,18 +490,19 @@
...
+                "appService.neverPromptBuildDuringDeploy": {
+                    "scope": "resource",
```
EricJizbaMSFT: This flag doesn't match the other flags. Do something like "showBuildDuringDeployPrompt"

src/explorer/DeploymentSlotTreeItem.ts
```diff
@@ -31,7 +32,7 @@ export class DeploymentSlotTreeItem extends SiteTreeItem {
     }
 
     public async loadMoreChildren(_node: IAzureParentNode<DeploymentSlotTreeItem>): Promise<IAzureTreeItem[]> {
-        return workspace.getConfiguration().get("appService.showRemoteFiles") ?
+        return workspace.getConfiguration('appService').get(configurationSettings.showRemoteFiles) ?
```
EricJizbaMSFT: nit: you could do this in contants.ts:
```
export const extensionPrefix: string = 'appService';
```

src/explorer/SiteTreeItem.ts
```diff
@@ -124,6 +136,33 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             });
     }
 
+    private async enableScmDoBuildDuringDeploy(fsPath: string, runtime: string): Promise<void> {
+        const yesButton: MessageItem = { title: 'Yes' };
+        const noButton: MessageItem = { title: 'No', isCloseAffordance: true };
```
EricJizbaMSFT: Use 'Dont' warn again' to be consistent with our other dialogs. It also makes it more clear that this action is different than just closing the dialog

src/explorer/SiteTreeItem.ts
```diff
@@ -124,6 +136,33 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
+                opn('https://aka.ms/Kwwkbd');
+            }
+        }
+        if (input === yesButton) {
+            let oldSettings: string[] | string = workspace.getConfiguration('appService', Uri.file(fsPath)).get(constants.configurationSettings.zipIgnorePattern);
+            if (typeof oldSettings === "string") {
+                oldSettings = [oldSettings];
+                // settings have to be an array to concat the proper zipIgnoreFolders
+            }
+            workspace.getConfiguration('appService', Uri.file(fsPath)).update(constants.configurationSettings.zipIgnorePattern, oldSettings.concat(zipIgnoreFolders), ConfigurationTarget.WorkspaceFolder);
```
EricJizbaMSFT: Don't pass in `ConfigurationTarget.WorkspaceFolder`. That's already the default

src/explorer/SiteTreeItem.ts
```diff
@@ -124,6 +136,33 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
+            await fse.writeFile(path.join(fsPath, constants.deploymentFileName), constants.deploymentFile);
+        } else {
+            workspace.getConfiguration('appService', Uri.file(fsPath)).update(constants.configurationSettings.neverPromptBuildDuringDeploy, true, ConfigurationTarget.WorkspaceFolder);
```
EricJizbaMSFT: Only update this if they click "Don't warn again". If they close the dialog, we should warn again

src/explorer/SiteTreeItem.ts
nturinski: I did 'Don't show again' since it's not really a warning

src/explorer/SiteTreeItem.ts
```diff
@@ -124,6 +136,33 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             });
     }
 
+    private async enableScmDoBuildDuringDeploy(fsPath: string, runtime: string): Promise<void> {
+        const yesButton: MessageItem = { title: 'Yes' };
+        const dontShowAgainButton: MessageItem = { title: "Don't show again", isCloseAffordance: true };
```
EricJizbaMSFT: This button isn't the close affordance. There should still be a 'close' button that means "I don't want to enable scm right now, but warn me again next time"

src/explorer/SiteTreeItem.ts
```diff
@@ -124,6 +136,33 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
+        const learnMoreButton: MessageItem = { title: 'Learn More' };
+        const zipIgnoreFolders: string[] = constants.getIgnoredFoldersForDeployment(runtime);
+        const buildDuringDeploy: string = `Run build script during deployment?  The "${zipIgnoreFolders.join(", ")}" directory will be built during the deployment, rather than zipped, resulting in a faster deployment.`;
```
EricJizbaMSFT: How about "Would you like to configure your project for faster deployment?"

I'm worried that we'll just confuse users if we include too much of the "how" in the message, and I think that's what the "learn more" button is for.

PR #323, opened 2018-03-06
Cancel website validation when redeploying

StephenWeatherford:


src/validateWebSite.ts
```diff
@@ -24,13 +24,32 @@ type WebError = {
...
+export function cancelWebsiteValidation(siteTreeItem: SiteTreeItem): void {
+    const cancellation = cancellations.get(siteTreeItem.id);
+    if (cancellation) {
+        cancellations.delete(siteTreeItem.id);
+        cancellation.canceled = true;
+    }
+}
 
 export async function validateWebSite(siteTreeItem: SiteTreeItem, outputChannel: OutputChannel, telemetryReporter: TelemetryReporter): Promise<void> {
+    cancelWebsiteValidation(siteTreeItem);
```
EricJizbaMSFT: Do you need to call cancelWebsiteValidation again here? Shouldn't this be taken care of since you call it before deploy?

StephenWeatherford: Yes, but that assumes the caller is doing the right thing.

PR #326, opened 2018-03-07
Some cancel telemetry

StephenWeatherford:
The only things I can put in quickly, but should handle the major scenarios.

### DIFF
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -110,9 +110,10 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         outputChannel: OutputChannel,
         telemetryReporter: TelemetryReporter,
         configurationSectionName: string,
-        confirmDeployment: boolean = true
+        confirmDeployment: boolean = true,
+        telemetryProperties: { [key: string]: string }
     ): Promise<void> {
-        await this.siteWrapper.deploy(fsPath, client, outputChannel, configurationSectionName, confirmDeployment);
+        await this.siteWrapper.deploy(fsPath, client, outputChannel, configurationSectionName, confirmDeployment, telemetryProperties);
 
         // Don't wait
         validateWebSite(this, outputChannel, telemetryReporter).then(
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -8,12 +8,12 @@ import * as WebSiteModels from 'azure-arm-website/lib/models';
 import * as opn from 'opn';
 import { ExtensionContext, OutputChannel, window } from 'vscode';
 import { ILogStream, SiteWrapper } from 'vscode-azureappservice';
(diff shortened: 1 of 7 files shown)


src/explorer/SiteTreeItem.ts
```diff
@@ -110,9 +110,10 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         outputChannel: OutputChannel,
         telemetryReporter: TelemetryReporter,
         configurationSectionName: string,
-        confirmDeployment: boolean = true
+        confirmDeployment: boolean = true,
+        telemetryProperties: { [key: string]: string }
```
EricJizbaMSFT: `TelemetryProperties`

src/extension.ts
```diff
@@ -153,16 +157,23 @@ export function activate(context: vscode.ExtensionContext): void {
...
+            fsPath = await util.showWorkspaceFoldersQuickPick("Select the folder to deploy", this.properties);
             node = target;
         }
 
         if (!node) {
-            node = <IAzureNode<WebAppTreeItem>>await tree.showNodePicker(WebAppTreeItem.contextValue);
+            try {
+                node = <IAzureNode<WebAppTreeItem>>await tree.showNodePicker(WebAppTreeItem.contextValue);
+            } catch (err2) {
+                if (err2 instanceof UserCancelledError) {
```
EricJizbaMSFT: Do `parseError(err2).isUserCancelledError` (we've had weird bugs in the past where `instanceof` doesn't work if there are somehow multiple versions of the ui package)

src/extension.ts
```diff
@@ -206,9 +217,9 @@ export function activate(context: vscode.ExtensionContext): void {
         const yesButton: vscode.MessageItem = { title: 'Yes' };
         const noButton: vscode.MessageItem = { title: 'No', isCloseAffordance: true };
         if (await vscode.window.showInformationMessage('Deploy to deployment slot?', yesButton, noButton) === yesButton) {
```
EricJizbaMSFT: Suggest doing the same 'deployingToWebApp' that you did for CreateWebApp above

src/util.ts
```diff
@@ -87,6 +87,7 @@ export async function showWorkspaceFoldersQuickPick(placeHolderString: string):
     const pickedItem = await vscode.window.showQuickPick(folderQuickPickItems, folderQuickPickOption);
 
     if (!pickedItem) {
+        telemetryProperties.cancelStep = placeHolderString;
```
EricJizbaMSFT: nit: I'd prefer if this was just a constant string like "showWorkspaceFolders". We're never going to have _multiple_ calls to showWorkspaceFolders in one command so I'd rather have something consistent across all commands

src/util.ts
```diff
@@ -99,6 +100,7 @@ export async function showWorkspaceFoldersQuickPick(placeHolderString: string):
         });
 
         if (!browseResult) {
+            telemetryProperties.cancelStep = 'showWorkspaceFoldersQuickPick';
```
EricJizbaMSFT: How about 'showWorkspaceFoldersBrowse'

StephenWeatherford: Fixed all.

PR #355, opened 2018-03-15
Refactor swapSlot to be a normal command instead of a wizard

EricJizbaMSFT:
Rather than switching this over to use the shared Wizard in the UI package, I'll just change it not to be a wizard anymore (since its only one step)


src/commands/swapSlots.ts
```diff
@@ -0,0 +1,54 @@
...
+                label: slot.treeItem.client.slotName,
+                description: '',
+                data: slot.treeItem
+            };
+
+            otherSlots.push(otherSlot);
+        }
+    }
+
+    const quickPickOptions = { placeHolder: `"${sourceSlotClient.slotName}" will be swapped with the destination slot.`, ignoreFocusOut: true };
```
StephenWeatherford: This always felt confusing to me.  Maybe something along the lines of "Select which slot to swap with {slotname}"?

PR #360, opened 2018-03-22
Add deploy resource context menu to apps

### LINKED ISSUE #284: Remove repetitive actions when deploying
Perhaps I'm mistaken, but I remember there being a right click context menu item to deploy (it might've been both zip and git deploy actually), but now it's gone in favor of the new deploy button.

image

It took me an embarrassingly long time to figure out what changed, and I couldn't figure out how to redeploy my app until I came and read the docs. Even once I found it, clicking deploy forces you to go pick a bunch of options which could be seeded for the user if the deploy is executed through the context menu.

nturinski:
Fixes #284 

I opted to label the context action as "Deploy Resource" for the following reasons:

1. Staying consistent to Azure* that typically refers to projects you deploy to web apps as "resources".
2. Makes it more clear that the web app/slot is being ACTED on as the Resource/Project is the object of the sentence.  I think that this is more clear than "Deploy to Web App"

*https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-deploy-cli

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
             },
             {
                 "command": "appService.Deploy",
-                "title": "Deploy to Web App",
+                "title": "Deploy Resource",
                 "category": "Azure App Service",
                 "icon": {
                     "light": "resources/light/Deploy.svg",
@@ -311,7 +311,7 @@
                     "group": "2_appServiceControlCommands@4"
                 },
                 {
-                    "command": "appService.OpenVSTSCD",
+                    "command": "appService.Deploy",
                     "when": "view == azureAppService && viewItem == appService",
                     "group": "3_appServiceDeployCommands@1"
                 },
@@ -319,6 +319,11 @@
                     "command": "appService.ConfigureDeploymentSource",
                     "when": "view == azureAppService && viewItem == appService",
                     "group": "3_appServiceDeployCommands@2"
+                },
+                {
+                    "command": "appService.OpenVSTSCD",
+                    "when": "view == azureAppService && viewItem == appService",
+                    "group": "3_appServiceDeployCommands@3"
                 },
                 {
                     "command": "diagnostics.OpenLogStream",
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
             },
             {
                 "command": "appService.Deploy",
-                "title": "Deploy Resource",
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
Wait but the article you linked is exactly the scenario we want stay away from as much as possible (in terms of wording). That article is about deploying a resource for the first time - aka you're creating a _new_ web app. We want to make as clear as possible that this is deploying your current project _to an existing web app_.

I think "to" is very important -  "Deploy Web App" and "Deploy to Web App" mean very different things. Without the "to", its synonymous with "Create Web App".

fiveisprime:
It seems to me that "resource" is used in that article because deploying an ARM template could be pretty much anything. I don't like "resource" here - it's generic for no reason.

I agree with Eric that "to" is the super important part.

Be specific and include "to".. I would really rather this be "Deploy to Web App."

nturinski:
Right, "Deploy Web App" is confusing because there are two things that could be considered a web app.  Your app on Azure and the web app project.  That's why I would opt for something more generic then, like local project or local web app.

I thought the issue we had a few releases ago was that all context menu items assume the "Azure Web App" to be the subject of the verb since that is the item that is right-clicked.  Such as, Start.  Start what?  The Azure Web App.  Or Open in Portal.  What is being opened?  The Azure Web App.

I think if we word it as "Deploy to Web App," we've changed the assumed subject of the context menu.  Deploy what?  The local project.

The alternative, "Deploy Local Project," is assuming that the Azure Web App is the subject.   Deploy to what?  The Azure Web App.

I don't have a strong opinion on this, but I do want to avoid the issues that caused us to remove the context menu item in the first place.  "Deploy to Web App" is word for word what we had action labeled as in the past.

EricJizbaMSFT:
I have a strong opinion _against_ the following:
1. Deploy Resource
1. Deploy to Resource
1. Deploy Web App

The reasons we removed this entry point still exist and I think this entry point will be confusing for some people no matter what. We're only adding this back because we've had several users request it. I'd be fine with any of the following:
1. Deploy
1. Deploy to Web App
1. Deploy to Slot
1. Deploy local project
1. Deploy workspace
1. Deploy workspace folder
1. Deploy folder
1. Deploy zip

PR #361, opened 2018-03-22
Don't prompt destructive deploy action on newly created web apps

### LINKED ISSUE #347: We shouldn't ask to verify deployment if the web app was created during the deploy wizard
@nturinski @fiveisprime @jinglouMSFT

nturinski:
There is a local listener that flips a boolean that is listening for a change on the tree.  If a web app is created, the boolean is flipped to true and the user is not prompted about destructive deploys.

Fixes #347

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -161,6 +161,12 @@ export function activate(context: vscode.ExtensionContext): void {
             node = target;
         }
 
+        let newApp: boolean = false;
+        // if the tree changes here, a new app was created
+        tree.onDidChangeTreeData(() => {
+            newApp = true;
+        });
+
         if (!node) {
             try {
                 node = <IAzureNode<WebAppTreeItem>>await tree.showNodePicker(WebAppTreeItem.contextValue);
@@ -173,7 +179,7 @@ export function activate(context: vscode.ExtensionContext): void {
         }
 
         try {
-            await node.treeItem.deploy(fsPath, outputChannel, ui, reporter, extensionPrefix, true, this.properties);
+            await node.treeItem.deploy(fsPath, outputChannel, ui, reporter, extensionPrefix, !newApp, this.properties);
         } catch (err) {
             if (parseError(err).isUserCancelledError) {
                 throw err;


src/extension.ts
```diff
@@ -161,6 +161,12 @@ export function activate(context: vscode.ExtensionContext): void {
             node = target;
         }
 
+        let newApp: boolean = false;
+        // if the tree changes here, a new app was created
+        tree.onDidChangeTreeData(() => {
```
EricJizbaMSFT: This is an interesting approach, but I'm worried about timing issues. For example, what if I want to deploy to "WebApp1" which happens to be stopped, and I do the following:
1. Right click WebApp1 -> Start
1. Select Deploy
1. Select Subscription
&lt;WebApp1 is started and tree refreshes so that the label is updated&gt;
1. Select WebApp1

Expected:
I'm prompted to confirm deployment since WebApp1 already exists

Actual:
I'm not prompted because the label refresh triggered onDidChangeTreeData

nturinski: Yeah, you're right.  I had trouble reproducing with those steps (since stop => start happens pretty quickly on my machine) but if I do this:

1.) Stop WebApp 1
2.) Start a deploy
3.) Before I pick a node, start WebApp1
4.) Deploy to WebApp1

I was thinking alternatively I could check the number of web apps before and after the node select action.  Because creating a new web app opens up a new wizard (and closes the deploy wizard), we shouldn't run into the same issue.

Thoughts?

EricJizbaMSFT: I would advocate for changing something inside the shared package. I feel like you're hesitant to change it, but I think that's the only concrete way to make sure we don't have timing issues. Even with your alternative above, it would be complicated if I create multiple web apps at the same time.

Here are a couple ideas off the top of my head:
1. Change the return value of showNodePicker from Promise<IAzureNode> to Promise<[IAzureNode, boolean]> (easy to explain, but annoying to implement because its a breaking change)
1. Add another event to AzureTreeDataProvider - something like 'onNodeCreated' that passes the node or node's id to the listener. Then you can check if that matches the node returned by showNodePicker

EricJizbaMSFT: Just saw this:
>  Because creating a new web app opens up a new wizard (and closes the deploy wizard), we shouldn't run into the same issue.

But there's still ways we could run into issues. For example I create a new web app while deploying, but also delete a web app during that time.

nturinski: Closing in lieu of option 2.

PR #363, opened 2018-03-26
Fix open in portal for children of slots

### LINKED ISSUE #362: Open in Portal for application settings on a Slot is broken
screen shot 2018-03-26 at 8 26 38 am

EricJizbaMSFT:
Sample tree for the sake of discussion:
screen shot 2018-03-26 at 8 35 08 am

There's an inconsistency with 'deploymentSlots' vs 'slots' in the portal and we have to special case _something_. However, rather than special-case the fourth node in the tree, I'm now special-casing the third node. This ensures that openInPortal works for any child of a slot, which fixes openInPortal for the fifth node in the tree. Fixes #362

### DIFF
--- a/src/explorer/DeploymentSlotsTreeItem.ts
+++ b/src/explorer/DeploymentSlotsTreeItem.ts
@@ -36,7 +36,7 @@ export class DeploymentSlotsTreeItem implements IAzureParentTreeItem {
     }
 
     public get id(): string {
-        return `${this.client.id}/deploymentSlots`;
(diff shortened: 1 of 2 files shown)


StephenWeatherford:
You can mark testcases as P2 or P3. But if it isn't likely to regress and isn't high pri, that's fine.

EricJizbaMSFT:
Yeah IMO the "likelihood to regress" vs "priority" doesn't make the cut for this one

PR #366, opened 2018-03-26
Leverage eventListener to stifle deploy prompt

nturinski:
Depends on this PR https://github.com/Microsoft/vscode-azuretools/pull/130

Not sure if the logic added by my last commit was necessary since the event is only fired after a node was created from the node picker, but I might just not be creative enough to think of a way to produce a bug.

Fixes #361


src/extension.ts
```diff
@@ -164,23 +171,29 @@ export function activate(context: vscode.ExtensionContext): void {
...
+            }
+            await node.treeItem.deploy(fsPath, outputChannel, ui, reporter, extensionPrefix, confirmDeployment, this.properties);
         } catch (err) {
             if (parseError(err).isUserCancelledError) {
                 throw err;
             }
             const appServicePlan = await node.treeItem.client.getAppServicePlan();
             this.properties.servicePlan = appServicePlan.sku.size;
             throw err;
+        } finally {
```
EricJizbaMSFT: I liked that you used try/finally, but it only works if onNodeCreatedFromQuickPickDisposable is created right before the 'try' starts. For example, if the user cancels in showNodePicker, this would never get disposed in the current implementation

src/extension.ts
```diff
@@ -154,6 +154,13 @@ export function activate(context: vscode.ExtensionContext): void {
     actionHandler.registerCommand('appService.Deploy', async function (this: IActionContext, target?: vscode.Uri | IAzureNode<WebAppTreeItem> | undefined): Promise<void> {
         let node: IAzureNode<WebAppTreeItem>;
         let fsPath: string;
+        let newApp: IAzureNode<WebAppTreeItem>;
+        let confirmDeployment: boolean = true;
+        const onNodeCreatedFromQuickPickDisposable: vscode.Disposable = tree.onNodeCreatedFromQuickPick((node: IAzureNode<WebAppTreeItem>) => {
```
EricJizbaMSFT: How about we name this something shorter, like 'newNodeListener'

src/extension.ts
```diff
@@ -164,23 +171,29 @@ export function activate(context: vscode.ExtensionContext): void {
...
                 if (parseError(err2).isUserCancelledError) {
                     this.properties.cancelStep = `showNodePicker:${WebAppTreeItem.contextValue}`;
                 }
                 throw err2;
             }
         }
-
         try {
-            await node.treeItem.deploy(fsPath, outputChannel, ui, reporter, extensionPrefix, true, this.properties);
+            if (newApp && newApp.id === node.id) {
```
EricJizbaMSFT: I like this logic

src/extension.ts
nturinski: Does that mean I should call dispose twice, once in the finally and once in the catch here
`catch (err2) {
                if (parseError(err2).isUserCancelledError) {
                    this.properties.cancelStep = `showNodePicker:${WebAppTreeItem.contextValue}`;
                }
                throw err2;
            }`

EricJizbaMSFT: No that means you need another try/finally that wraps _all_ of the code between `const onNodeCreatedFromQuickPickDisposable: vscode.Disposable = ...` and when you call dispose

EricJizbaMSFT: Yep you got it

PR #370, opened 2018-03-27
Check if state is undefined

sunmorgus:
An error in the explorer pane if the state is returned as undefined.

### DIFF
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -161,7 +161,7 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
 
     private createLabel(state: string): string {
         return (this.client.slotName ? this.client.slotName : this.client.siteName) +    // Site/slot name
-            (state && typeof state !== 'undefined' && state.toLowerCase() !== 'running' ? ` (${state})` : '');  // Status (if site/slot not running)
+            (state && state != undefined && state.toLowerCase() !== 'running' ? ` (${state})` : '');  // Status (if site/slot not running)
     }
 }
 
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -161,7 +161,7 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
 
     private createLabel(state: string): string {
         return (this.client.slotName ? this.client.slotName : this.client.siteName) +    // Site/slot name
-            (state && state.toLowerCase() !== 'running' ? ` (${state})` : '');  // Status (if site/slot not running)
+            (state && typeof state !== 'undefined' && state.toLowerCase() !== 'running' ? ` (${state})` : '');  // Status (if site/slot not running)
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/explorer/SiteTreeItem.ts
```diff
@@ -161,7 +161,7 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
 
     private createLabel(state: string): string {
         return (this.client.slotName ? this.client.slotName : this.client.siteName) +    // Site/slot name
-            (state && state.toLowerCase() !== 'running' ? ` (${state})` : '');  // Status (if site/slot not running)
+            (state && state !== undefined && state.toLowerCase() !== 'running' ? ` (${state})` : '');  // Status (if site/slot not running)
```
EricJizbaMSFT: `undefined` is a false-y value, so `state && state !== undefined ...` should be equivalent to just `state && ...`.

Do you have repro steps for this bug? I'm wondering if there's something else that needs to be fixed

sunmorgus: I do, yes. When connecting to a subscription where I don't have full permissions on all the web app resources, the "state" will not be available for some items. When this occurs, I only get "Error: Cannot read property 'toLowerCase' of undefined" in the explorer window:

VS-Code explorer window:
image

Web Apps in the Portal (note how I can see the Status on some, but not all):
image

EricJizbaMSFT: Do you mind sharing what role you have on those resources? We tried 'reader' and 'wesite contributor' but I can still see the status

Here's what I see for the options:
screen shot 2018-03-27 at 3 39 59 pm

It should be listed if you check 'Access Control' under the resource

sunmorgus: So, it's a custom role, not one of the built in ones, which, interestingly enough, is the same role for App Services that I can see the status for...

sunmorgus: A little more info... if I click on all the App Services in the list, there are 3 that appear like the below... think that may be the issue?

image

image

EricJizbaMSFT: That certainly could be a part of the problem. I'm not sure how familiar you are with VS Code extension development - but can you F5 this repo and find out where the exception is being thrown? Again - I'm skeptical that it's at this location, but there could easily be another 'toLowerCase' either in this repo or in one of our dependencies

I would turn on 'All Exceptions' in the debug view _right_ before you expand the subscription node in the tree. If you do it too soon, you'll see a bunch of VS Code exceptions that aren't relevant
screen shot 2018-03-27 at 4 20 12 pm

sunmorgus: Ok... running from the repo changed the error message a bit ("Cannot read property 'match' of undefined')... (sorry for all the screenshots... not sure what the best way to get this to you is)

image

image

image

image

image

image

image

image

image

sunmorgus: So, apparently, the serverFarmId for some of the resources in our subscription is blank. Not sure how that happened or if we need to clean those resources up or not. That being said, I'll play around with the code here to see if there is any way I can get it to return something valid (maybe skip those invalid resources?) when this occurs, so that I can still use the extension.

If it's ok with you, I'll close this PR and start a new one after I try to find a better solution.

EricJizbaMSFT: Yeah I think you're right. We've changed the code since 0.6.1, but in both cases its the serverFarmId that's undefined. Without a serverFarmId, you'll definitely have limited functionality for those resources. That being said, we should still display the resources that you _do_ have a serverFarmId for. 

Please file an issue on this repo to track the work. If you want to tackle it, add a note in the issue so that we know. Otherwise we can take a look

Issue #371, opened 2018-03-28
Web Apps Aren't Listed When serverFarmId Is Undefined


sunmorgus:
Per conversation in #370, the serverFarmId in the SiteClient constructor can return as undefined, causing the Web App List in the explorer pane to return an error and not display the available web apps.

Going to review the code and submit a PR when a better solution is found (other than the solution submitted in #370).

EricJizbaMSFT:
@sunmorgus FYI - we just released an update to the extension that includes this fix.

Thanks again for the contribution!

PR #372, opened 2018-03-28
Fix for when SiteClient constructor throws ArgumentError;

sunmorgus:
Fix for #371


sunmorgus:
@EricJizbaMSFT I had looked into maybe trying to get it to still return a result... that looks like it would be part of the https://github.com/Microsoft/vscode-azuretools as opposed to this one (possibly). I'll look into submitting a PR there as well so that we can display them in the list (maybe with a flag or label next to it of some sort indicating there was a problem).

PR #373, opened 2018-03-28
Show labels for invalid app services

sunmorgus:
Updated code from #372 to still allow the invalid resources to be listed along with the valid ones.


sunmorgus:
Here's what I've got so far...

image

1. Converted to "InvalidWebAppTreeItem" with no children
1. Added grayscale webapp icon

Couple of questions/input needed...

1. Can't decide whether to keep the "(invalid)" tag at the end of the label... I'd say having no children and the gray icon would be enough to indicate that there was some sort of problem, but I do like the extra little bit of info. 
   1. I did attempt to add a child item using the "InvalidTreeItem", but that won't work without having to build the fake Site object like I did before.
1. Considering attempting to include one or two context menu items for the invalid WebApps... 
   1. "Open WebApps Portal" (since I can't do "Open In Portal" without a valid id). Not even sure I could do that though, without having full site info
   1.  A static string item that says "App service configuration is invalid"

src/explorer/WebAppTreeItem.ts
```diff
@@ -126,3 +127,20 @@ export class WebAppTreeItem extends SiteTreeItem {
         return await fs.readFile(templatePath, 'utf8');
     }
 }
+
+export class InvalidWebAppTreeItem implements IAzureTreeItem {
```
EricJizbaMSFT: Please move this to a separate file "InvalidWebAppTreeItem.ts"

src/explorer/WebAppTreeItem.ts
```diff
@@ -126,3 +127,20 @@ export class WebAppTreeItem extends SiteTreeItem {
...
+    public static contextValue: string = 'file';
```
EricJizbaMSFT: I assume you copied this from the FileTreeItem haha? Please make the context value something like 'invalidWebApp' or 'invalidAppService'

src/explorer/WebAppTreeItem.ts
```diff
@@ -126,3 +127,20 @@ export class WebAppTreeItem extends SiteTreeItem {
...
+    public readonly contextValue: string = InvalidWebAppTreeItem.contextValue;
+    public readonly commandId: string = 'appService.showFile';
```
EricJizbaMSFT: Remove the commandId property - it was only used in the FileTreeItem

src/explorer/WebAppTreeItem.ts
```diff
@@ -24,6 +24,7 @@ export class WebAppTreeItem extends SiteTreeItem {
     public readonly appSettingsNode: IAzureTreeItem;
     public readonly webJobsNode: IAzureTreeItem;
     public readonly folderNode: IAzureTreeItem;
+    public readonly invalidAppServiceNode: IAzureTreeItem;
```
EricJizbaMSFT: Remove this property - its never used

src/explorer/WebAppProvider.ts
```diff
@@ -31,22 +31,18 @@ export class WebAppProvider implements IChildProvider {
...
+        await Promise.all(webAppCollection
+            .map(async (s: Site) => {
                 try {
-                    const siteClient = new SiteClient(s, node);
-                    return siteClient !== undefined;
-                    // tslint:disable-next-line:no-unsafe-any
-                } catch (err) {
-                    return false;
+                    const siteClient: SiteClient = new SiteClient(s, node);
+                    const appServicePlan: AppServicePlan = await siteClient.getAppServicePlan();
```
EricJizbaMSFT: Wrap `getAppServicePlan` and `treeItems.push` in `if (!siteClient.isFunctionApp)`

We have a completely separate extension that handles function apps: https://github.com/Microsoft/vscode-azurefunctions

src/explorer/WebAppProvider.ts
```diff
@@ -31,22 +31,18 @@ export class WebAppProvider implements IChildProvider {
...
+                    treeItems.push(new WebAppTreeItem(siteClient, appServicePlan));
+                } catch {
+                    treeItems.push(new InvalidWebAppTreeItem(`${s.name} (invalid)`));
```
EricJizbaMSFT: I like 'invalid'. However, rather than doing the formatting yourself, please specify `s.name` as the `label` property and `Invalid` as the `description` property

That will ensure its displayed in the same format as our other description properties (which as of today is the same format that you did, but could change in the future)

src/explorer/WebAppTreeItem.ts
```diff
@@ -126,3 +127,20 @@ export class WebAppTreeItem extends SiteTreeItem {
         return await fs.readFile(templatePath, 'utf8');
     }
 }
+
+export class InvalidWebAppTreeItem implements IAzureTreeItem {
+    public static contextValue: string = 'file';
+    public readonly contextValue: string = InvalidWebAppTreeItem.contextValue;
+    public readonly commandId: string = 'appService.showFile';
+
+    constructor(readonly label: string) {
```
EricJizbaMSFT: If you didn't know, the contextValue is used to determine which commands are shown in the context menu. See the 'viewItem' property in package.json:
https://github.com/Microsoft/vscode-azureappservice/blob/master/package.json#L294

If you want to support OpenInPortal, just do the same thing in package.json except use this treeItem's contextValue as the viewItem instead of 'appService'. You also have to specify the `id` property on this treeItem, which is used under the covers [here](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/src/treeDataProvider/AzureNode.ts#L125) for openInPortal. I don't think you'll have to completely construct the fake Site object - it should just be Site.id (which should probably exist even if serverFarmId does not - but idk)

src/explorer/WebAppTreeItem.ts
sunmorgus: I included it here, as that's the same way the `DeploymentSlotsNATreeItem` is defined in the `DeploymentSlotsTreeItem.js`... there's a precedent for doing it like so... should I still move it to a different file?

EricJizbaMSFT: Yeah @nturinski did that, not me 😊 I generally prefer one exported class per file, since you never know how much a class will grow as time goes on

That being said - this is kind of a stylistic preference and I wouldn't block a PR just for this

src/explorer/WebAppTreeItem.ts
sunmorgus: Yeah, site.id doesn't exist for these either, so that's not going to work. I'd love for it to just open to the "WebApps" blade in the portal, but I'm not sure how to get a direct link for that (when I go there in the portal, my tenant id is in the URL).

EricJizbaMSFT: I wouldn't worry about it then. Judging by these screenshots, the Portal wasn't that helpful anyways: https://github.com/Microsoft/vscode-azureappservice/pull/370#discussion_r177598043

sunmorgus: Roger that!

src/explorer/WebAppTreeItem.ts
sunmorgus: Went ahead and moved it to it's own file (I prefer it that way as well, lol).

src/explorer/WebAppProvider.ts
sunmorgus: Ugh, fixed... forgot about that.

src/explorer/WebAppProvider.ts
sunmorgus: Description property added.

src/explorer/WebAppTreeItem.ts
sunmorgus: Fixed

src/explorer/WebAppTreeItem.ts
sunmorgus: Fixed.

src/explorer/WebAppTreeItem.ts
sunmorgus: Lol, yeah, my bad... copy/paste got me. Renamed to 'invalidAppService' (matches the WebAppTreeItem better).

src/explorer/InvalidWebAppTreeItem.ts
```diff
@@ -0,0 +1,25 @@
+import * as path from 'path';
```
EricJizbaMSFT: Please add the Microsoft copyright header - I'm not an expert on the legalese but you can check out this section for more info: https://github.com/Microsoft/vscode-azureappservice#contributing

src/explorer/InvalidWebAppTreeItem.ts
```diff
@@ -0,0 +1,25 @@
...
+import { } from 'vscode-azureappservice';
+import { IAzureTreeItem } from 'vscode-azureextensionui';
+import { } from '../constants';
+import { } from './DeploymentSlotsTreeItem';
```
EricJizbaMSFT: Just remove the import lines without any variables

src/extension.ts
```diff
@@ -214,6 +215,13 @@ export function activate(context: vscode.ExtensionContext): void {
 
         node.treeItem.openCdInPortal(node);
     });
+    actionHandler.registerCommand('invalidAppService.Invalid', async (node?: IAzureNode<InvalidWebAppTreeItem>) => {
```
EricJizbaMSFT: What's the purpose of this command? Is it just to show the message? I think this would be confusing for users - a context menu command isn't the best place to show messages. Users expect context menu commands to actually do something

I don't think the message is necessary, but I can help you show the message as a child of the invalidAppService if you want

sunmorgus: Yeah, I get what you mean. Went ahead and removed it... I don't think any further indicator is necessary :)

Issue #389, opened 2018-04-06
Deploying .NET Core WebApp Should Run Publish First


sunmorgus:
When deploying a .net core WebApp, the extension should fist run a `dotnet publish` command and use the default location of `bin/Debug/netcoreapp2.0/publish` directory as the directory to deploy. Currently, it deploys the working directory, which does not result in a working WebApp.

Perhaps this could be implemented with a "pre-deploy" command in the settings? Will investigate over the weekend.

EricJizbaMSFT:
I actually worked on [something similar](https://github.com/Microsoft/vscode-azurefunctions/issues/223) in the functions extension recently. Rather than a "pre-deploy" command setting, I chose to look for a 'publish' task in your 'tasks.json' file and just run that before every deploy. That being said, I had to depend on an experimental api in VS Code that won't be generally available for at least a month. See [here](https://github.com/Microsoft/vscode/issues/45664) for more info, but it basically provides a way to list and run tasks from an extension.

Feel free to investigate, but right now waiting for that API seems like the best option. I think  VS Code's task system has a lot of benefits and I would prefer to leverage that if at all possible. I definitely don't want to end up re-implementing our own custom task system.

That being said, you (or we) could certainly fix the other issue by adding some sort of user setting for the default path. We actually have a 'deploySubPath' setting in the functions extension that does exactly what you want - if that works for you, you could just use the same pattern https://github.com/Microsoft/vscode-azurefunctions/blob/master/package.json#L539

sunmorgus:
Yeah, I agree waiting for that API seems like the best way to go... no sense in re-inventing the wheel lol.

I'll take a look at the 'deploySubPath' setting and get something submitted for review :)

JeroenVogelpoel:
Is there any progress on this issue yet?

Out-of-the-box deployments of ASP.Net Core projects to Azure App Services using this VS Code extension still results in faulty deployments using the wrong artifacts. The API referenced should/would/might be available now?

Also, there seems to be no implicit publish action at all when deploying; just a regular Debug build being copied over to App Services.

EricJizbaMSFT:
Yeah that api is ready - I'll stick this in the current milestone

PR #393, opened 2018-04-10
Add 'deploySubpath' setting

sunmorgus:
Added 'deploySubpath' setting, related to #389


sunmorgus:
@EricJizbaMSFT I'm not sure if all of my previous commits should really be showing in the PR or not... wasn't expecting that. If they shouldn't be, I can close this PR, and try and get just the commits for #389 to show (I suppose that means re-forking?).

src/extension.ts
```diff
@@ -173,6 +175,12 @@ export function activate(context: vscode.ExtensionContext): void {
                 node = target;
             }
 
+            const deploySubpath: string = workspace.getConfiguration(extensionPrefix).get(configurationSettings.deploySubpath);
+
+            if (deploySubpath) {
+                fsPath = path.join(fsPath, deploySubpath);
```
EricJizbaMSFT: So this won't quite work. We have several entry points for deploying and one of them is right clicking on a folder to deploy. If you right click on the 'bin/Debug/netcoreapp2.0/publish' folder, then you don't want to append 'bin/Debug/netcoreapp2.0/publish' _again_.

I also prefer to make settings as obvious as possible. That's why in the functions extension you see your sub path when you're selecting a workspace, rather than "secretly" appending the path _after_ the prompt:
screen shot 2018-04-11 at 11 57 36 am
(I have two workspace folders open '3' and '4', and '4' has deploySubPath set to 'example/path')

EricJizbaMSFT: You can honestly copy over most of the code in [this file](https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/utils/workspace.ts) and replace the code in `util.showWorkspaceFoldersQuickPick`

The only thing that we would lose is some telemetry stuff, but I can add that back after the fact if you're not familiar with our telemetry stuff

src/constants.ts
```diff
@@ -1,31 +1,32 @@
...
-            return undefined;
-    }
-}
-
-export enum configurationSettings {
-    zipIgnorePattern = 'zipIgnorePattern',
-    showBuildDuringDeployPrompt = 'showBuildDuringDeployPrompt',
-    showRemoteFiles = 'showRemoteFiles'
-}
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
sunmorgus: So... I think the whitespace change in this is from an earlier commit, that got overwritten by my change (below)... I did a compare on the file in vscode and adding the "deploySubpath" setting was the only thing that change...

src/util.ts
```diff
@@ -43,19 +45,21 @@ export function parseAzureResourceId(resourceId: string): { [key: string]: strin
...
-                    description: value.uri.fsPath,
-                    data: value
-                };
-            }
-        }) :
-        [];
-    folderQuickPickItems.unshift(browse);
+    const folderQuickPickItems = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders.map((value) => {
+        {
+            const subpath: string = vscode.workspace.getConfiguration(extensionPrefix).get(configurationSettings.deploySubpath);
```
EricJizbaMSFT: So currently we only use `showWorkspaceFoldersQuickPick` for deploying in this repo, but it's a generic function that we could use in other places. In those cases, we don't want to add `deploySubpath`. Please add an optional parameter like 'subPathSetting' where you pass in configurationSettings.deploySubpath

sunmorgus: I attempted merging your master over mine (following [this guide](https://help.github.com/articles/merging-an-upstream-repository-into-your-fork/), doesn't look like it effected the whitespace in that file though. I did merge your master over to mine prior to working on this as well, which has me wondering if that's not where the whitespace changes came from in the first place...

sunmorgus: `subPathSetting` param added to `showWorkspaceFoldersQuickPick`.

src/constants.ts
```diff
@@ -1,31 +1,32 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
```
EricJizbaMSFT: Maybe it's the line-endings that's causing the problem? You can check the status bar in VS Code:
screen shot 2018-04-19 at 9 32 04 am

Looks like it should be 'CLRF' for this file (we don't have a strict LF vs CLRF rule - it should just remain constant per file after it was created). If yours doesn't say CLRF, you can click that in the status bar and it should prompt you to switch line endings:
screen shot 2018-04-19 at 9 34 50 am

src/extension.ts
```diff
@@ -233,7 +235,8 @@ export function activate(context: vscode.ExtensionContext): void {
         // prompt user to deploy to newly created web app
         if (await vscode.window.showInformationMessage('Deploy to deployment slot?', yesButton, noButton) === yesButton) {
             this.properties[deployingToDeploymentSlot] = 'true';
-            const fsPath = await util.showWorkspaceFoldersQuickPick("Select the folder to deploy", this.properties);
+            const subpath: string = vscode.workspace.getConfiguration(extensionPrefix).get(configurationSettings.deploySubpath);
```
EricJizbaMSFT: Hmm still not quite right. The second parameter to `get` is missing:
```
vscode.workspace.getConfiguration(extensionPrefix).get(configurationSettings.deploySubpath, **FOLDERPATH**)
```

This is important for multi-root workspaces where the setting might apply on a per-folder basis. Rather than passing in `subpath` to `showWorkspaceFoldersQuickPick`, you could just pass in `configurationSettings.deploySubpath`. Then you can get `subPath` inside `vscode.workspace.workspaceFolders.map`

src/constants.ts
sunmorgus: So, it does say LF at the bottom; however, when I change it to CRLF and click save, it just switches back to LF and doesn't show any changes in the SCM...

EricJizbaMSFT: There's probably some VS Code setting controlling that. At least now we know what the change is. I think we've spent enough time on the whitespace stuff - fix the other comment and we should be good to merge this

src/extension.ts
sunmorgus: So... looking at the API documentation... there doesn't appear to be a way to do the folder path on the get: https://code.visualstudio.com/docs/extensionAPI/vscode-api#WorkspaceConfiguration... I believe that's something that's only part of the "Update" for configurations...

I did test this using single folders, multiple folders without a workspace, and using workspaces, and I don't remember running into an issue where I would need to specify a folder path on the get...

EricJizbaMSFT: Sorry - I meant to say 'getConfiguration', not 'get'

Here's more info about multi-root workspaces (if you're not familiar): https://code.visualstudio.com/docs/editor/multi-root-workspaces
Make sure you 'Add folder to workspace' as described

sunmorgus: Ok, I think I finally see what you mean... sorry to be argumentative... this was easier in the branch that I created for adding the subpath from the context menu...

I moved the get configuration to the `showWorkspaceFoldersQuickPick`, as requested. There's still one tiny issue though that I'm not sure how to fix... if you set the subpath setting in the folder settings, it works great... but if you set it in the workspace settings, then it still applies it to all the folders in the quick pick. I think I might be able to use the .inspect<T> method of getConfiguration to work around this... thoughts?

EricJizbaMSFT: I think that's by design - sometimes people want all of their workspace folders to have the same settings

src/extension.ts
```diff
@@ -158,8 +158,8 @@ export function activate(context: vscode.ExtensionContext): void {
         if (await vscode.window.showInformationMessage('Deploy to web app?', yesButton, noButton) === yesButton) {
             this.properties[deployingToWebApp] = 'true';
 
-            const fsPath = await util.showWorkspaceFoldersQuickPick("Select the folder to deploy", this.properties);
-            await createdApp.treeItem.deploy(createdApp, fsPath, outputChannel, ui, reporter, extensionPrefix, false, this.properties);
+            const fsPath = await util.showWorkspaceFoldersQuickPick("Select the folder to deploy", this.properties, configurationSettings.deploySubpath);
+            await createdApp.treeItem.deploy(node, fsPath, outputChannel, ui, reporter, extensionPrefix, false, this.properties);
```
EricJizbaMSFT: I believe this is a bad merge - the two calls in this file should not use 'node' anymore. We just changed this yesterday: https://github.com/Microsoft/vscode-azureappservice/pull/413/files

EricJizbaMSFT: Otherwise this looks good though :)

sunmorgus: Ok, should be fixed now... sorry for the wait!

EricJizbaMSFT:
Thanks @sunmorgus!

sunmorgus:
Thank you!

Issue #384, opened 2018-04-04
Remote debug node-based app services


EricJizbaMSFT:

EricJizbaMSFT:
@auchenberg  you're working on this correct?

auchenberg:
Yes @EricJizbaMSFT there's a branch for this with a working prototype: https://github.com/Microsoft/vscode-azureappservice/tree/node-remote-debug

This will be owned by an engineering team going forward.

fiveisprime:
This will be in preview and behind a feature flag for Build with only Node.js support.

EricJizbaMSFT:
Closed in #407. Remaining work #416 #417

PR #407, opened 2018-04-17
Add remote debugging support for node on linux

mrcrane:
Adding initial support for remote debugging for node on linux. #384 

This work assumes that changing the "remoteDebuggingEnabled" site config value will signal to the blessed node container to start the container appropriately. So setting this single config value should start the node process with the correct parameters and also set the APPSVC_TUNNEL_PORT appropriately.

These changes are completed but not currently deployed to Azure, so it is not possible to truly test this scenario end-to-end. Instead, I have tested that the "remoteDebuggingEnabled" setting is applied correctly, and also that I can attach the debugger to a webapp that is manually configured to setup remote debugging. Once the azure changes are deployed (in the next week or so) this change will be ready for a more thorough testing.


package.json
```diff
@@ -180,6 +183,21 @@
                 "title": "Create New Deployment Slot",
                 "category": "Azure App Service"
             },
+            {
+                "command": "diagnostics.EnableRemoteDebug",
```
EricJizbaMSFT: Please also add these to the commandPalette section with feature flag logic

package.json
```diff
@@ -180,6 +183,21 @@
...
+                "title": "Enable Remote Debugging",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "diagnostics.DisableRemoteDebug",
+                "title": "Disable Remote Debugging",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "diagnostics.StartRemoteDebug",
```
EricJizbaMSFT: nit: I'd prefer if the command name was more similar to the title "diagnostics.AttachDebugger"

src/diagnostics/RemoteDebugCommands.ts
```diff
@@ -0,0 +1,174 @@
...
+        this.reportMessage('Updating site configuration to set remote debugging...');
+        await this._siteClient.updateConfiguration(siteConfig);
+        this.reportMessage('Updating site configuration done...');
+    }
+
+    private async checkForRemoteDebugSupport(): Promise<void> {
+        const siteConfig: SiteConfigResource = await this.getSiteConfig();
+
+        // So far only node on linux is supported
+        if (!siteConfig.linuxFxVersion.toLowerCase().startsWith('node')) {
```
EricJizbaMSFT: This will get a null ref exception on windows. Suggest:
```
if (siteConfig.linuxFxVersion && !siteConfig.linuxFxVersion.toLowerCase().startsWith('node'))
```

src/diagnostics/RemoteDebugCommands.ts
```diff
@@ -0,0 +1,174 @@
...
+            throw new Error('Azure Remote Debugging is not supported for this instance type');
```
EricJizbaMSFT: nit: How about: 'Azure Remote Debugging is currently only supported for node on Linux.'

src/diagnostics/RemoteDebugCommands.ts
```diff
@@ -0,0 +1,174 @@
...
+
+        // So far only node is supported
+        return {
+            name: sessionId,
+            type: 'node',
+            protocol: 'inspector',
+            request: 'attach',
+            address: 'localhost',
+            port: portNumber,
+            localRoot: vscode.workspace.rootPath,
```
EricJizbaMSFT: `vscode.workspace.rootPath` is deprecated. What's the purpose of 'localRoot'? Can you leave it undefined? Otherwise you should pick something from vsocde.workspaceFolders

src/diagnostics/RemoteDebugCommands.ts
```diff
@@ -0,0 +1,174 @@
...
+    private async needUpdateSiteConfig(isRemoteDebuggingToBeEnabled: boolean): Promise<boolean> {
+        const siteConfig: SiteConfigResource = await this.getSiteConfig();
+        return isRemoteDebuggingToBeEnabled !== siteConfig.remoteDebuggingEnabled;
+    }
+
+    private async updateSiteConfig(isRemoteDebuggingToBeEnabled: boolean): Promise<void> {
+        const siteConfig: SiteConfigResource = await this.getSiteConfig();
+
+        siteConfig.remoteDebuggingEnabled = isRemoteDebuggingToBeEnabled;
+        this._siteConfig = siteConfig;
```
EricJizbaMSFT: nit: Rather than:
```
this._siteConfig = siteConfig;
```
suggest
```
this._siteConfig = await this._siteClient.updateConfiguration(siteConfig);
```

It's unlikely, but just in case updateConfiguration had any side effects or didn't work completely correctly

src/diagnostics/RemoteDebugCommands.ts
```diff
@@ -0,0 +1,174 @@
...
+    private async getSiteConfig(): Promise<SiteConfigResource> {
+        if (!this._siteConfig) {
+            this.reportMessage('Fetching site configuration...');
+            this._siteConfig = await this._siteClient.getSiteConfig();
+        }
+
+        return this._siteConfig;
+    }
+
+    private async needUpdateSiteConfig(isRemoteDebuggingToBeEnabled: boolean): Promise<boolean> {
```
EricJizbaMSFT: nit: IMO it would be more readable if these two lines of code weren't in their own method (which is only used once)

src/diagnostics/RemoteDebugCommands.ts
```diff
@@ -0,0 +1,174 @@
...
+            const confirmEnableMessage: string = 'The app configuration will be updated to enable remote debugging and restarted. Would you like to continue?';
+            await this.setRemoteDebug(true, confirmEnableMessage);
+
+            this.reportMessage('Starting tunnel proxy...');
+
+            const publishCredential: User = await this._siteClient.getWebAppPublishCredential();
+            tunnelProxy = new TunnelProxy(portNumber, this._siteClient, publishCredential, this._outputChannel);
+            await tunnelProxy.startProxy();
+
+            this.reportMessage('Starting debugging...');
```
EricJizbaMSFT: nit: 'Attaching debugger...'

mrcrane:
Good feedback! I'm happy to get this PR to match the code organization and style that you expect. I saw several different approaches and wasn't sure what to emulate. I have a few comments/questions first, as I start making changes:

1. Probably the biggest reason I used a class was so that I could gather all of the common variables as private members so that I didn't need to continually pass all of them to every inner function. IMO this helps a lot with readability, but I'll match whatever typescript/this repo expects. If I don't use a fire-and-forget class, would you rather that I moved all of those private members to parameters everywhere? Or is there another approach that I should follow?
2. Most of this code is reusable for the Functions scenario, with some potential differences around detecting if the app is debuggable, detecting the language (for now we assume node, but eventually we would detect that), and debug configurations. I would hope that this can eventually move to vscode-azuretools with a bit of refactoring, but I wasn't quite ready to commit to that yet. I only say that to keep it in mind if it impacts decisions here.

mrcrane:
Another question, do you prefer that I add more commits for PR feedback or should I squash them as I go?

package.json
```diff
@@ -349,10 +367,25 @@
                     "when": "config.appService.logpoints == true && view == azureAppService && viewItem == appService",
                     "group": "4_appServiceDiagnosticsCommands@3"
                 },
+                {
+                    "command": "diagnostics.EnableRemoteDebug",
```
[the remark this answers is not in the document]
mrcrane: I hadn't thought about that but I think yes, that makes a lot of sense. I'll try testing it out.

src/diagnostics/RemoteDebugCommands.ts
mrcrane: The purpose is only to map sources from the remote path to the local path. Documentation (e.g. https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_remote-debugging) led me to believe that I needed to set both localRoot and remoteRoot.

EricJizbaMSFT: Yeah in their example they just use "${workspaceFolder}". Can you try that?

rootPath was deprecated because of multi-root workspaces. For example, the user might have their source code in the second folder in the workspace, but rootPath always maps to the first folder

mrcrane: This ended up being a bigger problem. "${workspaceFolder}" will not evaluate without the context of a folder to begin with (in launch.json, this would be the folder that the launch.json is in). Because I'm starting the debugger from the app services tree, there is no easy way to know which folder to use.

I put in a check so that it will only try to map sources if there is one folder open. Otherwise it will give up. This works for now, and I will see if there is a better solution.

mrcrane:
All feedback is addressed. Let me know if the common remote debug code is organized in a way you would expect. I just made a file called remoteDebugCommon.ts and exported multiple functions from there.

src/commands/remoteDebug/remoteDebugCommon.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ext } from '../../extensionVariables';
+
+export function reportMessage(message: string, progress: vscode.Progress<{}>): void {
+    ext.outputChannel.appendLine(message);
+    progress.report({ message: message });
+}
+
+export function checkForRemoteDebugSupport(siteConfig: SiteConfigResource): void {
+    // So far only node on linux is supported
+    if (!siteConfig.linuxFxVersion.toLowerCase().startsWith('node')) {
```
EricJizbaMSFT: Missed one of my previous comments:

> This will get a null ref exception on windows. Suggest:
> ```
> if (siteConfig.linuxFxVersion && !siteConfig.linuxFxVersion.toLowerCase().startsWith('node'))
> ```

mrcrane:
Thanks! I added a few issues to track some future work: #416 #417 

Please feel free to merge when you are ready (I don't have permissions).

PR #428, opened 2018-05-01
Move explorer to Azure viewContainer

EricJizbaMSFT:
See https://github.com/Microsoft/vscode-azuretools/issues/167

Hopefully VS Code releases this update this week, but otherwise we can just ship the vsix that doesn't include this change

screen shot 2018-05-01 at 4 15 03 pm

Edit: Updated screenshot after discussion with Matt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
 # Change Log
 All notable changes to the "azure-appservice" extension will be documented in this file.
 
-## 0.8.0 - 2018-05-01
+## 0.8.0 - 2018-05-03
 ### Added
 - Remote debugging support for Node.js on Linux (feature flag required)
 - Modal confirmation dialogs for more visibility
+
+### Changed
+- Moved App Service Explorer to Azure view container
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
Also removed 'Azure' from explorer names. This is what it now looks like:

screen shot 2018-05-01 at 9 14 47 am

Also if you only have one extension installed, this is what it now looks like:
screen shot 2018-05-01 at 9 16 08 am

CHANGELOG.md
```diff
@@ -1,11 +1,14 @@
 # Change Log
 All notable changes to the "azure-appservice" extension will be documented in this file.
 
-## 0.8.0 - 2018-05-01
+## 0.8.0 - 2018-05-03
```
StephenWeatherford: This means people won't get updated to the new extension version until they've updated VS Code (which people often put off when they're working on stuff).

EricJizbaMSFT: I agree with you, but I don't feel strongly about it. Chris and Matt don't seem to care. Feel free to add your opinion to the discussion here: https://github.com/Microsoft/vscode-azuretools/issues/167

PR #438, opened 2018-05-11
Turn on strictNullChecks

### LINKED ISSUE #364: Turn on strictNullChecks
Originally when we did tslint for this repo we decided to turn off strictNullChecks (I don't remember the full discussion). After the recent bug in the Cosmos repo that potentially could've been caught by this, I say we turn it on. Even if we end up using the non-null assertion operator in a lot of cases, I think its worth it.

EricJizbaMSFT:
Fixes #364

### DIFF
--- a/src/explorer/DeploymentSlotsTreeItem.ts
+++ b/src/explorer/DeploymentSlotsTreeItem.ts
@@ -4,12 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 import WebSiteManagementClient = require('azure-arm-website');
-import { ResourceNameAvailability } from 'azure-arm-website/lib/models';
-import { NameValuePair, Site, WebAppCollection } from 'azure-arm-website/lib/models';
+import { NameValuePair, ResourceNameAvailability, Site, WebAppCollection } from 'azure-arm-website/lib/models';
 import * as path from 'path';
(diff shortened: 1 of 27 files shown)


nturinski:
😢

src/explorer/DeploymentSlotsTreeItem.ts
```diff
@@ -4,12 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 import WebSiteManagementClient = require('azure-arm-website');
-import { ResourceNameAvailability } from 'azure-arm-website/lib/models';
-import { NameValuePair, Site, WebAppCollection } from 'azure-arm-website/lib/models';
+import { NameValuePair, ResourceNameAvailability, Site, WebAppCollection } from 'azure-arm-website/lib/models';
 import * as path from 'path';
 import { window } from 'vscode';
```
nturinski: Could get rid of this if you got rid of line 139 window.showQuickPick and replaced it with our ext.ui

EricJizbaMSFT: Discovered a bug due to this comment, so I submitted a separate PR with the fix: https://github.com/Microsoft/vscode-azureappservice/pull/440

Issue #394, opened 2018-04-10
Add Java Tomcat support


fiveisprime:
Add support for Tomcat 8.5 and 9.0

1. Project creation experience: add Tomcat 8.5/9.0 runtime to creation options
1. Support `wardeploy` Kudu endpoint

The Java team is expecting to tackle 1 and provide an experience using the [Maven plugin](https://github.com/Microsoft/azure-maven-plugins/blob/master/azure-webapp-maven-plugin/README.md#web-app-on-linux) to handle deployments in the near term.

Support for `wapdeploy` coming later (after Build).

jdneo:
@EricJizbaMSFT, @fiveisprime I'm thinking that some of the experience might be changed here. So let's discuss it first.

Currently, when we are going to deploy a project to Azure Web App, the steps are:

1. Select folder to be deployed
2. Select target Web App
3. Go deploy

Since `wardeploy` is different from `zipdeploy`: 

- The target is a war file, not a folder
- Currently only suitable for Tomcat image on Linux Web App

So could we change the steps to:

1. Select target Web App
2. According to the selection, we can get the `LinuxFXVersion` of the target Web App, then decide whether we go `zipdeploy` or `wardeploy`
3. Select folder/war file
4. Go deploy

Will this experience change acceptable to you?

EricJizbaMSFT:
As long as the experience doesn't change for JavaScript/C#, then I'm fine with this

PR #456, opened 2018-06-05
add support to war deploy

jdneo:
resolve: #394 

gif record:
tomcat

### DIFF
--- a/package.json
+++ b/package.json
@@ -695,7 +695,7 @@
         "portfinder": "^1.0.13",
         "request": "^2.83.0",
         "request-promise": "^4.2.2",
-        "vscode-azureappservice": "~0.16.2",
+        "vscode-azureappservice": "0.16.3",
         "vscode-azureextensionui": "~0.13.0",
         "vscode-azurekudu": "~0.1.6",
         "vscode-debugadapter": "^1.24.0",
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -99,7 +99,7 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
 
     public async deploy(
         node: IAzureNode,
-        fsPath: string,
+        fsPath: string | undefined,
         outputChannel: OutputChannel,
         ui: IAzureUserInput,
         configurationSectionName: string,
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -695,7 +695,7 @@
         "portfinder": "^1.0.13",
         "request": "^2.83.0",
         "request-promise": "^4.2.2",
-        "vscode-azureappservice": "~0.16.2",
+        "vscode-azureappservice": "0.16.3",
```
EricJizbaMSFT: You can update the version on this, but leave the '~' in so that we automatically get bug fix updates

src/explorer/SiteTreeItem.ts
```diff
@@ -109,19 +109,30 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         const correlationId = getRandomHexString(10);
         telemetryProperties.correlationId = correlationId;
 
-        const workspaceConfig: WorkspaceConfiguration = workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath));
+        const siteConfig: WebSiteModels.SiteConfigResource = await this.client.getSiteConfig();
+        let deploymentFilePath: string;
+        if (fsPath) {
+            deploymentFilePath = fsPath;
+        } else {
+            if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.toLowerCase().startsWith('tomcat')) {
```
EricJizbaMSFT: Add 'tomcat' to the runtimes in constants.ts
https://github.com/Microsoft/vscode-azureappservice/blob/master/src/constants.ts

src/explorer/SiteTreeItem.ts
```diff
@@ -109,19 +109,30 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         const correlationId = getRandomHexString(10);
         telemetryProperties.correlationId = correlationId;
 
-        const workspaceConfig: WorkspaceConfiguration = workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath));
+        const siteConfig: WebSiteModels.SiteConfigResource = await this.client.getSiteConfig();
+        let deploymentFilePath: string;
+        if (fsPath) {
+            deploymentFilePath = fsPath;
```
EricJizbaMSFT: nit: Not sure I see the point of deploymentFilePath. You could just do:

```
if (!fsPath) {
    if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.toLowerCase().startsWith('tomcat')) {
        fsPath = ...
    } else {
        fsPath = ...
    }
}
```

src/util.ts
```diff
@@ -90,6 +90,44 @@ export async function showWorkspaceFoldersQuickPick(placeHolderString: string, t
     }
 }
 
+export async function showWarQuickPick(placeHolderString: string, telemetryProperties: TelemetryProperties): Promise<string> {
```
EricJizbaMSFT: We're trying to get rid of this generic util file. Please move this somewhere else (maybe into SiteTreeItem since that's the only place it's used? Also no need to export it in that case).

src/util.ts
```diff
@@ -90,6 +90,44 @@ export async function showWorkspaceFoldersQuickPick(placeHolderString: string, t
...
+    const warFiles: vscode.Uri[] = await vscode.workspace.findFiles('**/*.war');
+    const warQuickPickItems: IAzureQuickPickItem<string | undefined>[] = warFiles.map((uri: vscode.Uri) => {
+        return {
+            label: path.basename(uri.fsPath),
+            description: uri.fsPath,
+            data: uri.fsPath
+        };
+    });
+
+    warQuickPickItems.unshift({ label: '$(package) Browse...', description: '', data: undefined });
```
EricJizbaMSFT: I'd prefer if browse was the last option - I think normally user's will just pick the first item from the workspace, right?

src/explorer/SiteTreeItem.ts
jdneo: Find that if I remove the `deploymentFilePath` the method `appservice.deploy()` will have an error said:
`Argument of type 'string | undefined' is not assignable to parameter of type 'string'`

Other places are fine.

Maybe we need change the method signature in `appservice.deploy()` and throw an error when the fsPath is `undefined`?

jdneo:
@EricJizbaMSFT The code has been updated except for the `fsPath` related.

And please notice that now for all the language type project, the deployment experience will be: 
select target web app -> select file -> deploy.

Before this PR, it's:
select file -> select target web app -> deploy.

PR #467, opened 2018-06-13
Improved Creation Process setting

### LINKED ISSUE #453: Improved getting started experience
When deploying an app, try to determine the type of app being deployed; based on the type of app, use the following defaults. These defaults follow the `$ az webapp up` experience.

### All stacks
There are some places where the extension can be a little smarter and remove some prompts.

- Only prompt for workspace if more than one is present (need to discuss the implications here...)
- Only prompt for subscription when more than one sub is available (factor filtering into this)

### Node.js
There will only be two prompts when creating then deploying a new node app: name and location. Everything will be created based on the name and location. 

- Linux
- Resource group name to `appsvc_rg_Linux_<location>`
- Set plan to B1 with name of `appsvc_asp_Linux_<location>`
- Set `SCM_DO_BUILD_DURING_DEPLOYMENT` to 1
- Add VS Code setting to ignore node_modules directory
- Use latest LTS version of node (whatever the recommended version is at the time)

### .NET Core
Since .NET Core doesn't support zip deploy on Linux, we should default to the Windows experience. 

- Windows
- Resource group name to `appsvc_rg_Windows_<location>`
- Set plan to B1 with name of `appsvc_asp_Windows_<location>`

nturinski:
Fixes #453 

New creation process will be by default and users can alter a setting to return to the older, advanced creation process.

New creation also applies to web apps created through deploy.


src/explorer/WebAppProvider.ts
```diff
@@ -51,7 +53,10 @@ export class WebAppProvider implements IChildProvider {
     }
 
     public async createChild(node: IAzureNode<IAzureTreeItem>, showCreatingNode: (label: string) => void, actionContext: IActionContext): Promise<IAzureTreeItem> {
-        const newSite: Site | undefined = await createWebApp(util.getOutputChannel(), node.ui, actionContext, node.credentials, node.subscriptionId, node.subscriptionDisplayName, showCreatingNode);
+        // get workspace fspath here, but only if one is opened
```
EricJizbaMSFT: Can you explain this comment? It doesn't make sense to me

package.json
```diff
@@ -589,6 +589,11 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Enables remote debugging of a Web App. This experimental preview only supports node on Linux."
+                },
+                "appService.advancedCreation": {
```
EricJizbaMSFT: If you haven't already, I'd run this name by @fiveisprime

package.json
```diff
@@ -589,6 +589,11 @@
                     "type": "boolean",
                     "default": false,
```
EricJizbaMSFT: Also same deal from the other PR - udpate the reference to the appservice package version since this PR directly relies on it

src/explorer/WebAppProvider.ts
nturinski: Ah, crap, that was a remnant from my old PR when I was retrieving the workspace in the extension and passing it to createWebApp.

PR #477, opened 2018-06-20
Update ui to 0.15.0

StephenWeatherford:
Verified that Open in Portal still works (tried subscription, app service and app service deployment slot)


StephenWeatherford:
Do I need to update the ref to ui in the appservice common package?

PR #478, opened 2018-06-21
Remove not now button and only enable logging if app logging is enabled

### LINKED ISSUE #391: Clicking 'Not Now' in the popped dialog displays '…connected to log-streaming service'after starting streaming logs for one deployment slot
Repro steps:
 1. Create a App and choose Deployment Slots node.
 2. Right click Deployment Slots and create a deployment slot.
 3. Right click the created deployment slot and choose 'Start Streaming Logs'.
 4. Select 'Not Now' in the popped dialog.
 5. Check the log message in output window.

Expect:
The deployment slot isn't connected to log-streaming service.

Actual:
The deployment slot displays '…connected to log-streaming service' in output window.
actual

nturinski:
Fixes #391 
Uses ui.showWarningMessage so that cancels will be handled appropriately.


nturinski:
Link in PR editted.

PR #486, opened 2018-06-25
Change home prefix based on app kind

### LINKED ISSUE #481: Display an error after expanding 'kudu' node under 'Log Files'
Repro Steps:
1. Click '+' to create a new web app.
2. Make sure the new web app is created successfully, then drill down from new web app > 'Log Files' > 'kudu'.
3. Check whether there is an error after expanding 'kudu' node.

Expect:
There is no error.

Actual:
Display an error 'Error: 'D:\home\ome\LogFiles\kudu' not found'.
compare

nturinski:
Fixes #481 

Based on the app kind, the kudu url path uses either `/home` or `\home`.

I tested with Linux, Windows, and custom domains.

Linux app kind = 'linux'
Windows app kind = 'app'
Linux custom domain = 'app,linux'

The reason that I used include('linux') rather than === 'linux' is that custom domains will use `/home`.

### DIFF
--- a/src/explorer/FolderTreeItem.ts
+++ b/src/explorer/FolderTreeItem.ts
@@ -34,7 +34,7 @@ export class FolderTreeItem implements IAzureParentTreeItem {
         const httpResponse: kuduIncomingMessage = <kuduIncomingMessage>(await kuduClient.vfs.getItemWithHttpOperationResponse(this.folderPath)).response;
         // response contains a body with a JSON parseable string
         const fileList: kuduFile[] = <kuduFile[]>JSON.parse(httpResponse.body);
-        const home: string = 'home/';
+        const home: string = this.client.kind.includes('linux') ? 'home/' : 'home\\';
         return fileList.map((file: kuduFile) => {
             return file.mime === 'inode/directory' ?
                 // truncate the home/ of the path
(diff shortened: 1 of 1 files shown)


src/explorer/FolderTreeItem.ts
```diff
@@ -34,7 +34,7 @@ export class FolderTreeItem implements IAzureParentTreeItem {
         const httpResponse: kuduIncomingMessage = <kuduIncomingMessage>(await kuduClient.vfs.getItemWithHttpOperationResponse(this.folderPath)).response;
         // response contains a body with a JSON parseable string
         const fileList: kuduFile[] = <kuduFile[]>JSON.parse(httpResponse.body);
-        const home: string = 'home/';
+        const home: string = this.client.kind.includes('linux') ? 'home/' : 'home\\';
```
EricJizbaMSFT: Just as an alternative: what if you make this just 'home' without the path separator and add `+ 1` to the length below? Then you don't have to check `this.client.kind`. That being said, I can't think of an example where your existing code wouldn't work so I don't care too much.

nturinski: 💡 That seems like a smarter idea.

PR #491, opened 2018-06-28
Nturinski/deploy metadata

### LINKED ISSUE #454: Store deployment metadata
Store information about the app once deployed so that subsequent deploys require no prompts (or only a confirmation)
 
Maybe something like

```json
"appservice.deploy-config" : {
  "app": " SUB_GUID/maherna-node-todo",
  "slot": "production",
  "path": "./dist"
}
```

nturinski:
Fixes #454 

This PR has a few different scenarios:

- When a user deploys through the Deploy button (the blue arrow), we check if they have deploymentConfigurations in their workspace.  If there are no workspaces or multi-root workspaces, we ignore the setting, but Matt says the numbers indicate there are not that many multi-root workspace users.
 - If there are deployConfigs, automatically deploy to the workspace to the app in the config
 - If the app or workspace (I'd guess the user would have to delete the workspace during the deployment) do not exist, then delete the configuration completely.
 - If the user doesn't have a config yet, prompt them to ask if they'd like to save these configurations
- If the user deployed through the context menu
 - verify that the app being deployed matches the config.  If it does, automatically deploy the project without prompting for a workspace.
 - If it doesn't, then ignore the settings.


src/explorer/SiteTreeItem.ts
```diff
@@ -170,6 +175,29 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
+                    app: node.id,
+                    path: fsPath
+                });
+            } else if (result === neverShowAgain) {
+                workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.deploymentConfigurations, {
+                    dontShowAgain: true
+                });
+            } else if (result === DialogResponses.learnMore) {
+                // tslint:disable-next-line:no-unsafe-any
+                opn('https://aka.ms/AA1pq88');
```
EricJizbaMSFT: I'm not sure a 'Learn more' link is needed for this. You're just saving some defaults - that seems pretty obvious to me

src/explorer/SiteTreeItem.ts
```diff
@@ -170,6 +175,29 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
+    private async promptToSaveDeployConfigs(node: IAzureNode<SiteTreeItem>, fsPath: string): Promise<void> {
+        const saveDeploymentConfig: string = `Always deploy the workspace "${fsPath}" to "${node.treeItem.client.fullName}"?`;
+        const neverShowAgain: MessageItem = { title: "Never" };
+        let result: MessageItem = DialogResponses.learnMore;
+        while (result === DialogResponses.learnMore) {
+            result = await ext.ui.showWarningMessage(saveDeploymentConfig, DialogResponses.yes, neverShowAgain, DialogResponses.skipForNow, DialogResponses.learnMore);
+            if (result === DialogResponses.yes) {
+                workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.deploymentConfigurations, {
+                    app: node.id,
+                    path: fsPath
```
EricJizbaMSFT: You can't use the full fsPath here. Workspace settings get checked in to GitHub and one user's full fsPath won't match another user's full fsPath.

We already have a 'deploySubpath' setting. I recommend leveraging that instead of adding a new setting

src/explorer/SiteTreeItem.ts
```diff
@@ -170,6 +175,29 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
         }
     }
+
+    private async promptToSaveDeployConfigs(node: IAzureNode<SiteTreeItem>, fsPath: string): Promise<void> {
+        const saveDeploymentConfig: string = `Always deploy the workspace "${fsPath}" to "${node.treeItem.client.fullName}"?`;
```
EricJizbaMSFT: I would change this wording to "Default to "${node.treeItem.client.fullName}" for future deployments in this workspace?"

package.json
```diff
@@ -595,6 +595,11 @@
                     "default": false,
                     "description": "Enables remote debugging of a Web App. This experimental preview only supports node on Linux."
                 },
+                "appService.deploymentConfigurations": {
+                    "scope": "resource",
+                    "type": "object",
```
EricJizbaMSFT: You should make the type more specific than 'object'. Users should be able to manually set their deployment configuration just by reading the setting descriptions/types

src/explorer/SiteTreeItem.ts
nturinski: I was thinking a new user wouldn't be sure what it means to save deployment configurations.  If they hit yes and couldn't deploy to anything but their first app, it could be kind of frustrating when you don't have a clear understanding as to what workspace settings are.

src/explorer/SiteTreeItem.ts
nturinski: This phrasing comes directly from big boss man @fiveisprime, but I have no real preference on how we want to express this.

src/logPoints/LogPointsDebugAdapter.ts
```diff
@@ -8,8 +8,8 @@ import {
     DebugSession, Event, InitializedEvent, Logger, logger,
     LoggingDebugSession, Source,
     TerminatedEvent, Thread
-} from 'vscode-debugadapter';
-import { DebugProtocol } from 'vscode-debugprotocol';
+} from 'vscode-debugadapter/lib/main';
+import { DebugProtocol } from 'vscode-debugprotocol/lib/debugProtocol';
```
nturinski: I'm not sure where these and a few other of the imports got changed from...

src/explorer/SiteTreeItem.ts
```diff
@@ -124,11 +123,25 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             }
         }
         cancelWebsiteValidation(this);
+        // tslint:disable-next-line:strict-boolean-expressions
+        if (workspace.workspaceFolders && workspace.workspaceFolders.length === 1) {
+            const currentWorkspace: WorkspaceFolder = workspace.workspaceFolders[0];
+            if (currentWorkspace.uri.fsPath === fsPath || this.isSubpath(currentWorkspace.uri.fsPath, fsPath)) {
```
EricJizbaMSFT: You can't compare paths directly (it won't work in several edge cases - like `\example\path` and `\example\..\example\path`). You have to do something like this: https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/utils/fs.ts#L92

src/explorer/SiteTreeItem.ts
```diff
@@ -167,6 +180,34 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
         }
     }
+
+    private async promptToSaveDeployConfigs(node: IAzureNode<SiteTreeItem>, currentWorkspacePath: string, fsPath: string): Promise<void> {
```
EricJizbaMSFT: What's the difference between 'currentWorkspacePath' and 'fsPath'? Please change the names to be more descriptive

src/explorer/SiteTreeItem.ts
EricJizbaMSFT: I think this is a pretty complex dialog already, though, and I'm not sure it helps them when they're the most confused (aka the _next_ time they go to deploy). 

We already have a confirmation dialog for deploying, correct? What if that dialog has a 'Change defaults' button added if we detect they're using this deploy config. Or even simpler - you could just print the 'learn more' link in the deploy logs

src/logPoints/LogPointsDebugAdapter.ts
EricJizbaMSFT: Well can ya change it back?

package.json
```diff
@@ -595,6 +595,15 @@
                     "default": false,
                     "description": "Enables remote debugging of a Web App. This experimental preview only supports node on Linux."
                 },
+                "appService.deploymentConfigurations": {
+                    "scope": "resource",
+                    "type": {
+                        "app": "string",
+                        "subpath": "string",
```
EricJizbaMSFT: I think you should use the existing subpath setting. I think it's confusing to have two separate settings

src/extension.ts
```diff
@@ -198,6 +200,31 @@ export function activate(context: vscode.ExtensionContext): void {
...
+                if (deployConfig && deployConfig.app && deployConfig.subpath !== undefined) {
+                    const pathExists: boolean = await fs.pathExists(join(vscode.workspace.workspaceFolders[0].uri.fsPath, deployConfig.subpath));
+                    const nodeFromConfig: IAzureNode<WebAppTreeItem> | undefined = <IAzureNode<WebAppTreeItem>>await ext.tree.findNode(deployConfig.app); // resolves to undefined if app can't be found
+                    // tslint:disable-next-line:strict-boolean-expressions
+                    if (pathExists && nodeFromConfig) {
+                        // tslint:disable-next-line:strict-boolean-expressions
+                        if (!fsPath && (!node || node && node.id === nodeFromConfig.id)) {
+                            /*
+                            * only use the deployConfig in the following situation:
+                            * if there is no fsPath and no node, then the entry point was the deploy button
```
EricJizbaMSFT: > if there is a target node and it matches the id in the deployConfig

It feels like you should do the same thing, but for the fsPath as well (aka check to see if it matches fsPathFromConfig)

src/explorer/SiteTreeItem.ts
```diff
@@ -167,6 +178,22 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
         }
     }
+
+    private async promptToSaveDeployConfigs(node: IAzureNode<SiteTreeItem>, activeWorkspacePath: string, deployedProjectPath: string): Promise<void> {
```
nturinski: Worth passing the reporter to get some more telemetry on this?

src/extension.ts
```diff
@@ -187,17 +190,52 @@ export function activate(context: vscode.ExtensionContext): void {
         const newNodes: IAzureNode<WebAppTreeItem>[] = [];
         let fsPath: string | undefined;
         let confirmDeployment: boolean = true;
+        this.properties.deployedWithConfigs = 'false';
```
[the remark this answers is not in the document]
EricJizbaMSFT: I don't think so

src/explorer/SiteTreeItem.ts
EricJizbaMSFT: You shouldn't have to pass the reporter! :) That's the whole point of this work: https://github.com/Microsoft/vscode-azureappservice/pull/495

src/explorer/SiteTreeItem.ts
```diff
@@ -167,6 +178,27 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
         }
     }
+
+    private async promptToSaveDeployConfigs(node: IAzureNode<SiteTreeItem>, activeWorkspacePath: string, deployedProjectPath: string, telemetryProperties: TelemetryProperties): Promise<void> {
+        const saveDeploymentConfig: string = `Always deploy the workspace "${path.basename(activeWorkspacePath)}" to "${node.treeItem.client.fullName}"?`;
+        const neverSaveDeploymentConfiguration: MessageItem = { title: "Never" };
+        let result: MessageItem = DialogResponses.learnMore;
+        result = await ext.ui.showWarningMessage(saveDeploymentConfig, DialogResponses.yes, neverSaveDeploymentConfiguration, DialogResponses.skipForNow);
```
EricJizbaMSFT: nit: get rid of `let result: MessageItem = DialogResponses.learnMore;` and just do `const result: MessageItem = await ext.ui.showWarningMessage(...`

src/explorer/SiteTreeItem.ts
```diff
@@ -167,6 +178,27 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
         }
     }
+
```
EricJizbaMSFT: I'm still a little confused by the naming because I generally think of the 'project' as the same folder as the 'workspace', but I assume `deployedProjectPath` is the full path being deployed? How about just `workspacePath` and `deployPath`

src/explorer/SiteTreeItem.ts
```diff
@@ -167,6 +178,27 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
+    private async promptToSaveDeployConfigs(node: IAzureNode<SiteTreeItem>, activeWorkspacePath: string, deployedProjectPath: string, telemetryProperties: TelemetryProperties): Promise<void> {
+        const saveDeploymentConfig: string = `Always deploy the workspace "${path.basename(activeWorkspacePath)}" to "${node.treeItem.client.fullName}"?`;
+        const neverSaveDeploymentConfiguration: MessageItem = { title: "Never" };
+        let result: MessageItem = DialogResponses.learnMore;
+        result = await ext.ui.showWarningMessage(saveDeploymentConfig, DialogResponses.yes, neverSaveDeploymentConfiguration, DialogResponses.skipForNow);
+        if (result === DialogResponses.yes) {
+            workspace.getConfiguration(constants.extensionPrefix, Uri.file(deployedProjectPath))
```
EricJizbaMSFT: Create a variable to contain `workspace.getConfiguration(constants.extensionPrefix, Uri.file(deployedProjectPath))` rather than copying it a bunch

src/explorer/SiteTreeItem.ts
```diff
@@ -167,6 +178,27 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
+                .update(constants.configurationSettings.deployToWebAppId, node.id);
+            workspace.getConfiguration(constants.extensionPrefix, Uri.file(deployedProjectPath))
+                .update(constants.configurationSettings.deploySubpath, path.relative(activeWorkspacePath, deployedProjectPath)); // '' is a falsey value
+            telemetryProperties.promptToSaveDeployConfigs = 'Yes';
+        } else if (result === neverSaveDeploymentConfiguration) {
+            workspace.getConfiguration(constants.extensionPrefix, Uri.file(deployedProjectPath))
+                .update(constants.configurationSettings.deployToWebAppId, constants.neverSaveDeploymentConfiguration);
+            telemetryProperties.promptToSaveDeployConfigs = 'Never';
+        }
+        if (!telemetryProperties.promptToSaveDeployConfigs) {
```
EricJizbaMSFT: You did this again - why not just make this an 'else' case for the above if statements?

src/constants.ts
```diff
@@ -29,5 +30,12 @@ export enum configurationSettings {
     zipIgnorePattern = 'zipIgnorePattern',
     showBuildDuringDeployPrompt = 'showBuildDuringDeployPrompt',
     deploySubpath = 'deploySubpath',
-    advancedCreation = 'advancedCreation'
+    advancedCreation = 'advancedCreation',
+    deployToWebAppId = 'deployToWebAppId'
+}
+
+export enum deploymentEntryPoint {
```
EricJizbaMSFT: So we don't want to just throw EVERY constant into this file. I would only put it in here if it meets either of the following:
1. It's used multiple times
1. It's only used once, but we expect to change it (i.e. maybe a version number) and we want it to be easy to find

I don't think this meets either of those qualifications

src/extension.ts
```diff
@@ -187,17 +190,52 @@ export function activate(context: vscode.ExtensionContext): void {
...
         const onNodeCreatedFromQuickPickDisposable: vscode.Disposable = tree.onNodeCreate((newNode: IAzureNode<WebAppTreeItem>) => {
             // event is fired from azure-extensionui if node was created during deployment
             newNodes.push(newNode);
         });
         try {
             if (target instanceof vscode.Uri) {
                 fsPath = target.fsPath;
+                this.properties.deploymentEntryPoint = constants.deploymentEntryPoint.fileExplorerContextMenu;
             } else {
+                this.properties.deploymentEntryPoint = target ? constants.deploymentEntryPoint.webAppContextMenu : constants.deploymentEntryPoint.deployButton;
```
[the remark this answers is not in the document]
nturinski: I don't really care since the behaviors are the same.  I'm not even sure how we would be able to differentiate the two entry points.

src/explorer/SiteTreeItem.ts
nturinski: Yeah, so the `activeWorkspacePath` is the active workspace that the user currently has opened vs. `deployedProjectPath` is the full path that is being deployed (the fsPath retrieved from the `appService.deploy`).  I'm fine with changing it though I'm not sure if it makes it clearer 😕

src/constants.ts
```diff
@@ -6,6 +6,7 @@
 export const deploymentFileName: string = '.deployment';
 export const deploymentFile: string = `[config]
 SCM_DO_BUILD_DURING_DEPLOYMENT=true`;
+export const neverSaveDeploymentConfiguration: string = 'neverSaveDeploymentConfiguration';
```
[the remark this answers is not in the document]
nturinski: 'None' wouldn't really mean the same thing as 'Never ask again' though.  Are you okay with just 'Never'?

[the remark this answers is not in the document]
EricJizbaMSFT: So when I say 'None', I'm thinking about it with my suggested name above. I think 'None' works better with 'defaultWebAppToDeploy' than 'Never'. In the 'None' case, I technically have a default defined (which means you don't need to ask me for defaults again), but that default just happens to be nothing.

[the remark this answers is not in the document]
EricJizbaMSFT: But anyways yes I'm more fine with 'Never' than 'neverSaveDeploymentConfiguration'

src/explorer/SiteTreeItem.ts
```diff
@@ -124,11 +124,22 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             }
         }
         cancelWebsiteValidation(this);
+        // tslint:disable-next-line:strict-boolean-expressions
+        if (workspace.workspaceFolders && workspace.workspaceFolders.length === 1) {
```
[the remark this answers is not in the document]
nturinski: It's tricky to combine the logic here.
1. Either way, I have to check for `workspace.workspaceFolders` or TS will complain about possibily undefined unless I pass the activeWorkspace, but that feels wrong.
1. I need to check if deployed path is either the active workspace or a subpath of the workspace, but before this point, we don't definitive fspath since it is prompted for later.
1. We could move the logic, but it might affect other things 😕

src/explorer/SiteTreeItem.ts
```diff
@@ -167,6 +178,23 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
         }
     }
+
+    private async promptToSaveDeployDefaults(node: IAzureNode<SiteTreeItem>, workspacePath: string, deployPath: string, telemetryProperties: TelemetryProperties): Promise<void> {
+        const saveDeploymentConfig: string = `Always deploy the workspace "${path.basename(workspacePath)}" to "${node.treeItem.client.fullName}"?`;
+        const neverSaveDeploymentConfiguration: MessageItem = { title: "Never" };
```
EricJizbaMSFT: Change this to "Don't Ask Again"

src/extension.ts
```diff
@@ -176,8 +180,7 @@ export function activate(context: vscode.ExtensionContext): void {
         // prompt user to deploy to newly created web app
         if (await vscode.window.showInformationMessage('Deploy to web app?', yesButton, noButton) === yesButton) {
             this.properties[deployingToWebApp] = 'true';
-
-            await createdApp.treeItem.deploy(createdApp, undefined, extensionPrefix, false, this.properties);
+            await vscode.commands.executeCommand('appService.Deploy', createdApp);
```
EricJizbaMSFT: Do this same thing for CreateSlot

PR #492, opened 2018-06-29
Get rid of prompt and call enableScm for node projects

nturinski:
With the new quickCreation, we want to enableScmDeploy automatically if the newly created app has a node runtime.

This removes the prompt that we used to show users to enableScmDeploy.

Though I'd actually prefer to make an App Setting rather than using the .deployment file

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -27,7 +27,6 @@ export function getIgnoredFoldersForDeployment(runtime: string): string[] {
 
 export enum configurationSettings {
     zipIgnorePattern = 'zipIgnorePattern',
-    showBuildDuringDeployPrompt = 'showBuildDuringDeployPrompt',
     deploySubpath = 'deploySubpath',
     advancedCreation = 'advancedCreation'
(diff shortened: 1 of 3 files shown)


nturinski:
Well, with this PR, their project will only be changed if the linuxFxVersion resolves to node|x.x so we're not relying on the `package.json` to figure that out-- thought that does make me want to revisit the logic in this PR: https://github.com/Microsoft/vscode-azuretools/pull/202

I think if we don't put any folders into the ignoreGlobPattern, we can mitigate breaking anybody's projects.  Worst case, they have a node_modules folder and we zip that as well (which is what they were already doing), best case, they never `npm install` locally and their deployment would just work.

I don't think the change was necessary because the prompt was confusing, but more because the answer to do you want a better deployment experience should always be yes, so let's just optimize it for them.

nturinski:
Closing this for now.  We would like to collect more telemetry on this before deciding to enableScmDeploy automatically.

PR #500, opened 2018-07-09
Change wording of scmPrompt and add telemetry

nturinski:
New wording for scmEnable prompt and added telemetry for created web app's OS and runtime and for scmEnable prompt.

### DIFF
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -144,7 +144,7 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         const dontShowAgainButton: MessageItem = { title: "No, and don't show again" };
         const learnMoreButton: MessageItem = { title: 'Learn More' };
         const zipIgnoreFolders: string[] = constants.getIgnoredFoldersForDeployment(runtime);
-        const buildDuringDeploy: string = `Would you like to update your workspace configuration to run npm install on the target server? This will improve deployment performance.`;
+        const buildDuringDeploy: string = `Would you like to update your workspace configuration to run npm install on the target server? This should improve deployment performance.`;
         let input: MessageItem | undefined = learnMoreButton;
         while (input === learnMoreButton) {
             input = await window.showInformationMessage(buildDuringDeploy, yesButton, dontShowAgainButton, learnMoreButton);
@@ -163,11 +163,15 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             }
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.zipIgnorePattern, oldSettings.concat(zipIgnoreFolders));
             await fse.writeFile(path.join(fsPath, constants.deploymentFileName), constants.deploymentFile);
+            telemetryProperties.enableScmInput = "Yes";
         } else if (input === dontShowAgainButton) {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
+            telemetryProperties.enableScmInput = "No, and don't show again";
         }
 
-        telemetryProperties.enableScmInput = input ? input.title : 'Canceled';
+        if (!telemetryProperties.enableScmInput) {
+            telemetryProperties.enableScmInput = "Canceled";
+        }
     }
 }
 
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -120,7 +120,7 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
             if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.startsWith(constants.runtimes.node) && siteConfig.scmType === 'None' && !(await fse.pathExists(path.join(fsPath, constants.deploymentFileName)))) {
                 // check if web app has node runtime, is being zipdeployed, and if there is no .deployment file
                 // tslint:disable-next-line:no-unsafe-any
-                await this.enableScmDoBuildDuringDeploy(fsPath, constants.runtimes[siteConfig.linuxFxVersion.substring(0, siteConfig.linuxFxVersion.indexOf('|'))]);
+                await this.enableScmDoBuildDuringDeploy(fsPath, constants.runtimes[siteConfig.linuxFxVersion.substring(0, siteConfig.linuxFxVersion.indexOf('|'))], telemetryProperties);
             }
         }
         cancelWebsiteValidation(this);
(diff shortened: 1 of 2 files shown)


src/explorer/SiteTreeItem.ts
```diff
@@ -139,12 +139,12 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
     }
 
-    private async enableScmDoBuildDuringDeploy(fsPath: string, runtime: string): Promise<void> {
+    private async enableScmDoBuildDuringDeploy(fsPath: string, runtime: string, telemetryProperties: TelemetryProperties): Promise<void> {
         const yesButton: MessageItem = { title: 'Yes' };
         const dontShowAgainButton: MessageItem = { title: "No, and don't show again" };
         const learnMoreButton: MessageItem = { title: 'Learn More' };
         const zipIgnoreFolders: string[] = constants.getIgnoredFoldersForDeployment(runtime);
-        const buildDuringDeploy: string = `Would you like to configure your project for faster deployment?`;
+        const buildDuringDeploy: string = `Would you like to update your workspace configuration to run npm install on the target server? This will improve deployment performance.`;
```
StephenWeatherford: Suggest "should" instead of "will"

src/explorer/SiteTreeItem.ts
```diff
@@ -166,6 +166,8 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
         } else if (input === dontShowAgainButton) {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
         }
+
+        telemetryProperties.enableScmInput = input ? input.title : 'Canceled';
```
StephenWeatherford: Note that that will send localized titles to telemetry...

src/extension.ts
```diff
@@ -172,7 +173,9 @@ export function activate(context: vscode.ExtensionContext): void {
         }
 
         const createdApp = <IAzureNode<WebAppTreeItem>>await node.createChild(this);
-
+        const createdAppConfig: SiteConfigResource = await createdApp.treeItem.client.getSiteConfig();
+        this.properties.os = createdAppConfig.kind ? createdAppConfig.kind : 'undefined';
```
StephenWeatherford: Not sure why you're not just leaving it undefined (which will result in a null entry in telemetry).

nturinski: Does it automatically convert the string "undefined" to a null entry?  Definitely didn't want that.

src/explorer/SiteTreeItem.ts
nturinski: Good point...

src/extension.ts
```diff
@@ -172,7 +173,9 @@ export function activate(context: vscode.ExtensionContext): void {
         }
 
         const createdApp = <IAzureNode<WebAppTreeItem>>await node.createChild(this);
-
+        const createdAppConfig: SiteConfigResource = await createdApp.treeItem.client.getSiteConfig();
```
EricJizbaMSFT: We don't want to await this just for telemetry. Do something like this:
https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/deploy/deploy.ts#L33

src/extension.ts
EricJizbaMSFT: Perhaps we should change the type of `TelemetryProperties` [here](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L240) to `string | undefined` instead of just `string`.

That way you could do `this.properties.os = createdAppConfig.kind` without ts yelling at you

src/extension.ts
```diff
@@ -172,7 +173,9 @@ export function activate(context: vscode.ExtensionContext): void {
...
+        this.properties.os = createdAppConfig.kind ? createdAppConfig.kind : 'undefined';
+        this.properties.runtime = createdAppConfig.linuxFxVersion ? createdAppConfig.linuxFxVersion : 'undefined';
```
EricJizbaMSFT: This doesn't cover telemetry for all the 'create' entry points. We also want to track telemetry if they create it as a part of deploy or other actions

src/explorer/SiteTreeItem.ts
```diff
@@ -163,8 +163,14 @@ export abstract class SiteTreeItem implements IAzureParentTreeItem {
...
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.zipIgnorePattern, oldSettings.concat(zipIgnoreFolders));
             await fse.writeFile(path.join(fsPath, constants.deploymentFileName), constants.deploymentFile);
+            telemetryProperties.enableScmInput = "Yes";
         } else if (input === dontShowAgainButton) {
             workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).update(constants.configurationSettings.showBuildDuringDeployPrompt, false);
+            telemetryProperties.enableScmInput = "No, and don't show again";
+        }
+
+        if (!telemetryProperties.enableScmInput) {
+            telemetryProperties.enableScmInput = "Canceled";
```
EricJizbaMSFT: nit: Couldn't you just make this an 'else' case for the above if/elseif blocks?

src/extension.ts
```diff
@@ -215,6 +224,15 @@ export function activate(context: vscode.ExtensionContext): void {
                     if (newApp.id === node.id) {
                         // if the node selected for deployment is the same newly created nodes, stifle the confirmDeployment dialog
                         confirmDeployment = false;
+                        newApp.treeItem.client.getSiteConfig().then(
+                            (createdAppConfig: SiteConfigResource) => {
+                                this.properties.os = createdAppConfig.linuxFxVersion ? 'Linux' : 'Windows';
+                                this.properties.runtime = createdAppConfig.linuxFxVersion ? createdAppConfig.linuxFxVersion : 'undefined';
```
EricJizbaMSFT: nit: I'd prefer properties.linuxFxVersion to be consistent with other telemetry: https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/deploy/deploy.ts#L27

src/extension.ts
```diff
@@ -215,6 +224,15 @@ export function activate(context: vscode.ExtensionContext): void {
                     if (newApp.id === node.id) {
                         // if the node selected for deployment is the same newly created nodes, stifle the confirmDeployment dialog
                         confirmDeployment = false;
+                        newApp.treeItem.client.getSiteConfig().then(
+                            (createdAppConfig: SiteConfigResource) => {
+                                this.properties.os = createdAppConfig.linuxFxVersion ? 'Linux' : 'Windows';
```
EricJizbaMSFT: nit: if "os" is just based on linuxFxVersion and we're already collecting linuxFxVersion, do we need a separate 'os' field?

nturinski: Agreed that we don't need it, but might make it more intuitive to read the telemetry especially for people without context.  But if you think it's redundant, I'm cool with removing it.

src/extension.ts
nturinski: Since this happens for every deploy action, having another property might make it easier to determine if it's the runtime for creation (so we can count created web apps) vs. just the runtime from a deployment-- not sure if that matters or not as we could filter on the createFromDeploy property?

EricJizbaMSFT: Yeah I was assuming we'd filter on your 'createdFromDeploy' property

src/extension.ts
EricJizbaMSFT: Yeah I guess I prefer to mess with the query in the telemetry dashboard to make it more readable, rather than trying to do extra stuff directly in the code here

PR #526, opened 2018-07-12
Add confirm boolean for modal

### LINKED ISSUE #523: Create Web App and then deploy prompts for confirmation
Repro steps:
1. Create new Web App or slot
1. When it asks "Do you want to deploy?" select yes

Expected: It doesn't confirm if I want to deploy since this is a new web app
Actual: It asks me if I'm sure I want to deploy

NOTE: This only affects the 'create' entry point. If you use the 'deploy' entry point and create a new web app as a part of the node-picker, you will not be prompted for confirmation.

nturinski:
Fixes #523


nturinski:
You were right-- when the target is from the File Explorer, there are more arguments.  I moved confirmDeployment to be the 2nd parameter to avoid any bugs.

PR #561, opened 2018-08-14
Shortcut Key for Azure View

### LINKED ISSUE #457: [Suggestion]Add keyboard shortcuts for 'Azure' tab
Repro Steps:
Hover the mouse on the Azure, then check whether there is a keyboard shortcuts for it.

Suggestion:
Add keyboard shortcuts for 'Azure' tab.
keyboard

nturinski:
Fixes #457 

I decided on ctrl+shift+a for [A]zure, but am open to any suggestions.

I suppose we should do this for all other Azure extensions?

### DIFF
--- a/package.json
+++ b/package.json
@@ -218,10 +218,7 @@
                 {
                     "id": "azure",
                     "title": "Azure",
-                    "icon": "resources/azure.svg",
-                    "command": {
-                        "id": "workbench.view.azure"
-                    }
+                    "icon": "resources/azure.svg"
                 }
             ]
         },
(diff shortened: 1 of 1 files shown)


StephenWeatherford:
Just a little concerned that this will cause a conflict eventually since vscode uses CTRL+SHIFT+letter for all its own commands.  Almost every other one is already taken.

nturinski:
Definitely understand that concern, but I think it's worth taking the shortcut key now before it gets taken since all other view shortcuts use the ctrl+shift+{$letter} format.  At least then when there is a conflict, we can say we called dibs.

PR #562, opened 2018-08-14
Add `Enable File Logging` context item in App/Files

### LINKED ISSUE #519: Add context menu to enable file logging
By default, the log files under the Log Files node only have the docker output along with this message

>Logging is not enabled for this container. Please use https://aka.ms/linux-diagnostics to enable logging to see container logs here.

That led me to automatically start right-clicking things to see if I could enable this without starting up streaming and I couldn't find anything. I'm thinking that the context menu should be at the app and Log Files level, but I'm open to discussion on this.

nturinski:
Fixes #519.

Entry points include context menu of Web App, Files, and Log Files nodes and Command Pallette.

### DIFF
--- a/package.json
+++ b/package.json
@@ -493,9 +493,14 @@
                     "group": "5_appServiceRemoteDebugCommands@2"
                 },
                 {
-                    "command": "appService.Refresh",
+                    "command": "appService.EnableLogging",
                     "when": "view == azureAppService && viewItem == folder",
                     "group": "1_folderGeneralCommands@1"
+                },
+                {
+                    "command": "appService.Refresh",
+                    "when": "view == azureAppService && viewItem == folder",
+                    "group": "1_folderGeneralCommands@2"
                 },
                 {
                     "command": "appService.OpenInPortal",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
         "onCommand:appService.StartRemoteDebug",
         "onCommand:appService.DisableRemoteDebug",
         "onCommand:appService.CreateSlot",
+        "onCommand:appService.EnableFileLogging",
         "onView:azureAppService"
     ],
     "main": "./out/src/extension",
@@ -210,6 +211,11 @@
             {
                 "command": "appService.ScaleUp",
                 "title": "Scale up App Service Plan...",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "appService.EnableFileLogging",
+                "title": "Enable File Logging",
                 "category": "Azure App Service"
             }
         ],
@@ -493,7 +499,7 @@
                     "group": "5_appServiceRemoteDebugCommands@2"
                 },
                 {
-                    "command": "appService.EnableLogging",
+                    "command": "appService.EnableFileLogging",
                     "when": "view == azureAppService && viewItem == folder",
                     "group": "1_folderGeneralCommands@1"
                 },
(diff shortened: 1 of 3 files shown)


src/commands/enableFileLogging.ts
```diff
@@ -0,0 +1,32 @@
...
+        await window.withProgress({ location: ProgressLocation.Notification, title: enablingLogging }, async (): Promise<void> => {
+            ext.outputChannel.appendLine(enablingLogging);
+            // tslint:disable-next-line:no-non-null-assertion
+            await node.treeItem.enableHttpLogs();
+            await commands.executeCommand('appService.Restart', node);
+            window.showInformationMessage(enabledLogging);
+            ext.outputChannel.appendLine(enabledLogging);
+        });
+    } else {
+        window.showInformationMessage(`File logging has already been enabled for ${node.treeItem.client.fullName}.`);
```
EricJizbaMSFT: So since `OpenLogStream` calls `enableFileLogging` - doesn't that mean this message would pop up every time you start the log stream? That seems annoying

src/extension.ts
```diff
@@ -334,6 +316,18 @@ export function activate(context: vscode.ExtensionContext): void {
     });
 
     registerEvent('appService.fileEditor.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, async function (this: IActionContext, doc: vscode.TextDocument): Promise<void> { await fileEditor.onDidSaveTextDocument(this, context.globalState, doc); });
+    registerCommand('appService.EnableFileLogging', async (node?: IAzureNode<SiteTreeItem> | IAzureNode<FolderTreeItem> | IAzureParentNode<IAzureTreeItem>) => {
```
EricJizbaMSFT: Do you need `| IAzureParentNode<IAzureTreeItem>`? Feels like the other two describe all entry points

src/commands/enableFileLogging.ts
```diff
@@ -0,0 +1,32 @@
+import { commands, ProgressLocation, window } from "vscode";
+import { DialogResponses, IAzureNode } from "vscode-azureextensionui";
+import { SiteTreeItem } from "../explorer/SiteTreeItem";
+import { ext } from '../extensionVariables';
+
+/*---------------------------------------------------------------------------------------------
```
EricJizbaMSFT: nit: Move the copyright header above the imports

package.json
```diff
@@ -493,10 +499,15 @@
                     "group": "5_appServiceRemoteDebugCommands@2"
                 },
                 {
-                    "command": "appService.Refresh",
+                    "command": "appService.EnableFileLogging",
```
EricJizbaMSFT: So this would show up on every folder, even sub folders, correct? I feel like we really only need it on the "Log Files" node. Also the web app node - which you mentioned in the description, but I don't see here.

nturinski: Yeah, good point.  I think I'll create a new Subfolder type for all the other treeItems.  As far as being on the "Files" node, that was on Matt's request though I agree with you that it seems a bit excessive (though easier to have it on both).

src/extension.ts
nturinski: Without `IAzureParentNode<IAzureTreeItem>`, I was getting type conflictions because of line 326.  Specifically, it wouldn't let me cast node.parent to IAzureNode<SiteTreeItem>.

PR #591, opened 2018-08-30
Add button to Stream Logs after deploy

### LINKED ISSUE #530: view logs (after deploy) doesn't take me to the app logs
Testing out the new improved e2e experience...  Once deployment is done I get a notification that the site is deployed with two options: "View Logs" and "Browse Website".

image

My expectation when clicking no "View Logs" was that I would see my Application's logs, not the extension deployment logs (Output -> Azure App Service). I'm not sure why I would want to see those logs??

EricJizbaMSFT:
Fixes #530 

Based on this PR in the shared package: https://github.com/Microsoft/vscode-azuretools/pull/251

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -10,6 +10,7 @@ import { pathExists } from 'fs-extra';
 import * as path from 'path';
 import { join } from 'path';
 import * as vscode from 'vscode';
+import { MessageItem } from 'vscode';
 import * as appservice from 'vscode-azureappservice';
 import { DialogResponses, IActionContext, IAzureNode, IAzureTreeItem, parseError } from 'vscode-azureextensionui';
 import * as constants from '../constants';
@@ -20,6 +21,7 @@ import * as util from '../util';
 import * as javaUtil from '../utils/javaUtils';
 import { isPathEqual, isSubpath } from '../utils/pathUtils';
 import { cancelWebsiteValidation, validateWebSite } from '../validateWebSite';
+import { startStreamingLogs } from './startStreamingLogs';
 
 // tslint:disable-next-line:max-func-body-length cyclomatic-complexity
 export async function deploy(context: IActionContext, confirmDeployment: boolean, target?: vscode.Uri | IAzureNode<SiteTreeItem> | undefined): Promise<void> {
@@ -79,6 +81,8 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         }
     }
 
+    const treeItem: SiteTreeItem = node.treeItem;
+
     if (newNodes.length > 0) {
         for (const newApp of newNodes) {
             if (newApp.id === node.id) {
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT:
Screenshot:
screen shot 2018-08-30 at 11 44 47 am

src/extension.ts
```diff
@@ -256,25 +256,7 @@ export function activate(context: vscode.ExtensionContext): void {
...
-            });
-            if (!isEnabled) {
-                // tslint:disable-next-line:no-non-null-assertion
-                await enableFileLogging(node!);
-            }
-        };
-
-        await startStreamingLogs(node.treeItem.client, verifyLoggingEnabled, node.treeItem.logStreamLabel);
-    });
+    registerCommand('appService.OpenLogStream', startStreamingLogs);
```
nturinski: You didn't have to import startStreamingLogs?

nturinski:
I'm fine with it, but wanted to verify-- The reason we don't put `startStreamingLogs` in the shared package is because of the pretty big implementation differences between App Service sand Functions?

src/extension.ts
EricJizbaMSFT: Nice catch - fixed

src/commands/deploy.ts
```diff
@@ -126,13 +130,30 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
     }
-    cancelWebsiteValidation(node.treeItem);
+    cancelWebsiteValidation(treeItem);
     await node.runWithTemporaryDescription("Deploying...", async () => {
-        // tslint:disable-next-line:no-non-null-assertion
-        await appservice.deploy(node!.treeItem.client, <string>fsPath, constants.extensionPrefix, context.properties);
+        await appservice.deploy(treeItem.client, <string>fsPath, constants.extensionPrefix, context.properties);
     });
+
+    const deployComplete: string = `Deployment to "${treeItem.client.fullName}" completed.`;
```
nturinski: Oh yeah, also you're not displaying the url anymore.

EricJizbaMSFT:
So the majority of startStreamingLogs _is_ in the shared package here: https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/src/startStreamingLogs.ts

It's a little confusing because this repo has a function `startStreamingLogs` which calls the shared package function, which is also called `startStreamingLogs`. I'm open to changing the name of one, but I don't care too much. Anyways the shared package has 115 lines of code and this repo only has 32 lines of code - so most of it is still in the shared package.

src/commands/deploy.ts
EricJizbaMSFT: So I did that intentionally because we now have 3 buttons and it felt cluttered with the url. I also feel like you don't need to see the url since you have 'Browse Website'. Thoughts?

nturinski: I agree with you.  I think I brought that up before, but I thought @fiveisprime or Chris wanted to keep it for visibility? Don't quote me on that though...

Issue #524, opened 2018-07-12
Allow user to 'change defaults' directly from deploy confirmation dialog


EricJizbaMSFT:
We discussed this offline already, but just filing an issue to keep track of it. If the user has defaults, they will only see the confirmation dialog before deploying, but they might not know how to change those defaults in the future. A 'change defaults' button on the confirmation dialog might make it more discoverable

PR #593, opened 2018-08-31
Add 'reset default' button for deploy warning 

Omelianenko:
One of the possible fixes for [#524](https://github.com/Microsoft/vscode-azureappservice/issues/524)

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -120,7 +120,26 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         const warning: string = `Are you sure you want to deploy to "${node.treeItem.client.fullName}"? This will overwrite any previous deployment and cannot be undone.`;
         context.properties.cancelStep = 'confirmDestructiveDeployment';
         const deployButton: vscode.MessageItem = { title: 'Deploy' };
-        await ext.ui.showWarningMessage(warning, { modal: true }, deployButton, DialogResponses.cancel);
+        if (defaultWebAppToDeploy && defaultWebAppToDeploy !== constants.none) {
+            const changeDefault: vscode.MessageItem = { title: 'Change default' };
+            const result: vscode.MessageItem = await ext.ui.showWarningMessage(warning, { modal: true }, deployButton, DialogResponses.cancel, changeDefault);
+            if (result.title === 'Change default') {
+                const localRoot = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0].uri.path : "";
+                //next line(s) should be CHANGED somehow (different OS, looks not healthy)
+                const settingsPath: string = localRoot.concat("\\.vscode\\settings.json");
+                const openPath = vscode.Uri.file(settingsPath);
+                vscode.workspace.openTextDocument(openPath).then(doc => {
+                    vscode.window.showTextDocument(doc);
+                });
+                const workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(constants.extensionPrefix, vscode.Uri.file(fsPath));
+                workspaceConfiguration.update(constants.configurationSettings.defaultWebAppToDeploy, '').then(async () => {
+                    await vscode.commands.executeCommand('appService.Deploy');
+                });
+                return;
+            }
+        } else {
+            await ext.ui.showWarningMessage(warning, { modal: true }, deployButton, DialogResponses.cancel);
+        }
         context.properties.cancelStep = '';
     }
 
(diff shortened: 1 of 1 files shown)


src/commands/deploy.ts
```diff
@@ -120,7 +120,26 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
+                const settingsPath: string = localRoot.concat("\\.vscode\\settings.json");
+                const openPath = vscode.Uri.file(settingsPath);
+                vscode.workspace.openTextDocument(openPath).then(doc => {
+                    vscode.window.showTextDocument(doc);
+                });
+                const workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(constants.extensionPrefix, vscode.Uri.file(fsPath));
+                workspaceConfiguration.update(constants.configurationSettings.defaultWebAppToDeploy, '').then(async () => {
+                    await vscode.commands.executeCommand('appService.Deploy');
+                });
+                return;
```
PrashanthCorp: Like the above comment. 
This command can be written as a sequence of two await commands.

src/commands/deploy.ts
```diff
@@ -120,7 +120,22 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         const warning: string = `Are you sure you want to deploy to "${node.treeItem.client.fullName}"? This will overwrite any previous deployment and cannot be undone.`;
         context.properties.cancelStep = 'confirmDestructiveDeployment';
         const deployButton: vscode.MessageItem = { title: 'Deploy' };
-        await ext.ui.showWarningMessage(warning, { modal: true }, deployButton, DialogResponses.cancel);
+        if (defaultWebAppToDeploy) {
+            const changeDefault: vscode.MessageItem = { title: 'Change default' };
+            const result: vscode.MessageItem = await ext.ui.showWarningMessage(warning, { modal: true }, deployButton, DialogResponses.cancel, changeDefault);
+            if (result.title === 'Change default') {
+                const localRootPath = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0].uri.path : "";
```
nturinski: Rather than getting the `localRootPath` again, you can check the `currentWorkspace` variable above:
l45: `currentWorkspace = vscode.workspace.workspaceFolders[0];`

If it's undefined, you know that you don't have to worry about deployConfig.

src/commands/deploy.ts
```diff
@@ -120,7 +120,22 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
+                const settingsPath = path.join(localRootPath, '.vscode', 'settings.json');
+                const doc = await vscode.workspace.openTextDocument(vscode.Uri.file(settingsPath));
+                vscode.window.showTextDocument(doc);
+                const workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(constants.extensionPrefix, vscode.Uri.file(fsPath));
+                await workspaceConfiguration.update(constants.configurationSettings.defaultWebAppToDeploy, '');
+                await vscode.commands.executeCommand('appService.Deploy');
```
nturinski: It's obvious to us now, but I think a comment here explaining why there's a recursive `appService.Deploy` call would be useful for future reference. :)

src/commands/deploy.ts
```diff
@@ -119,8 +120,22 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
+        }
+        const result: vscode.MessageItem = await ext.ui.showWarningMessage(warning, { modal: true }, ...items);
+        if (result.title === 'Reset default') {
+            // tslint:disable-next-line:no-non-null-assertion
+            const localRootPath = currentWorkspace!.uri.fsPath;
+            const settingsPath = path.join(localRootPath, '.vscode', 'settings.json');
+            const doc = await vscode.workspace.openTextDocument(vscode.Uri.file(settingsPath));
+            vscode.window.showTextDocument(doc);
+            const workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(constants.extensionPrefix, vscode.Uri.file(fsPath));
+            await workspaceConfiguration.update(constants.configurationSettings.defaultWebAppToDeploy, '');
```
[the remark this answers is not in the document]
Omelianenko: Is it better to just delete line 134? I don't need it because of [this line](https://github.com/Microsoft/vscode-azureappservice/blob/da506f428c3e9b1dd4a8004912ea4bcb2ff74b5a/src/commands/deploy.ts#L111)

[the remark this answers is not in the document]
EricJizbaMSFT: Well that line should also be using this: https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/ProjectSettings.ts#L74

Omelianenko:
I've added [line 61](https://github.com/Microsoft/vscode-azureappservice/blob/83ee98b35c7249bd00228f21e19c3d7f18af4c35/src/commands/deploy.ts#L61) but now not sure whether this decision is good. How it works now: if you deploy somewhere else than defaultWayAppToDeploy it resets defaultWayAppToDeploy and ask you about making current way to deploy as defaultWayAppToDeploy, but without line 61 it will ask only next deploying(read [this comment](https://github.com/Microsoft/vscode-azureappservice/blob/83ee98b35c7249bd00228f21e19c3d7f18af4c35/src/commands/deploy.ts#L59)). What do you think is better?

PR #603, opened 2018-09-05
Added node for cosmos connections

### LINKED ISSUE #582: Add node for cosmos connections
Like this (except for just the cosmos node for now):
image

Things that still need to be decided:
* The name of the node
* The icon for the node

This needs to be added under a feature flag. See the `appService.enableRemoteDebugging` setting as an example. https://github.com/Microsoft/vscode-azureappservice#preview-features

Omelianenko:
fixes #582


package.json
```diff
@@ -655,6 +655,11 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Enables the Advanced Web App Creation which will prompt for app name, location, website OS, runtime, resource group, and app service plan."
+                },
+                "appService.enableConnectionsNode": {
```
EricJizbaMSFT: Please remove this section until the "Connections" feature actually provides value to the user. This is the correct thing to do if the feature was in preview, but this actually isn't even ready for "preview" support.

You should see this in your settings.json:
screen shot 2018-09-06 at 9 40 59 am

But it should still work - which is all we need for our own testing purposes.

resources/dark/CloudService_16x.svg
```diff
@@ -0,0 +1 @@
....466-1.568 1.854-2.659 3.477-2.659.813 0 1.6.282 2.231.792l.384-.025c1.435 0 2.658 1.027 2.984 2.42A3.209 3.209 0 0 1 16 9.816z" id="outline"/><path class="icon-vs-bg" d="M4 7H1V6h3v1zm0 1H2v1h2V8zm0 2H1v1h3v-1zm11-.205c0 1.121-.805 2.057-1.856 2.186l-.193.019H7.257c-1.266-.023-2.243-1.068-2.243-2.379 0-1.035.635-1.918 1.517-2.243.202-1.33 1.294-2.348 2.607-2.348a2.56 2.56 0 0 1 1.934.889c.216-.08.445-.123.682-.123 1.094 0 1.992.894 2.064 2.02A2.215 2.215 0 0 1 15 9.795zm-1 .002c0-.672-.493-1.217-1.102-1.217l-.258.037a1.28 1.28 0 0 0 .182-.663c0-.653-.479-1.181-1.069-1.181-.457 0-.843.316-.997.761-.125-.866-.799-1.531-1.618-1.531-.906 0-1.642.811-1.642 1.813 0 .157.023.306.058.451l-.28-.037c-.696 0-1.261.623-1.261 1.391 0 .766.556 1.381 1.243 1.393v.002h5.694v-.006c.585-.033 1.05-.561 1.05-1.213z" id="iconBg"/><path class="icon-vs-fg" d="M14 9.793c0 .652-.465 1.18-1.049 1.213v.006H7.257v-.002c-.688-.012-1.243-.627-1.243-1.393 0-.767.564-1.39 1.261-1.39l.279.037a1.947 1.947 0 0 1-.058-.451C7.496 6.811 8.231 6 9.138 6c.819 0 1.493.665 1.618 1.531.154-.444.54-.761.997-.761.591 0 1.069.527 1.069 1.181 0 .247-.067.473-.182.663l.258-.037c.609 0 1.102.544 1.102 1.216z" id="iconFg"/></svg>
```
EricJizbaMSFT: Please post screenshots in both dark theme and light theme

src/explorer/ConnectionTreeItem.ts
```diff
@@ -0,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
+import { SiteClient } from 'vscode-azureappservice';
+import { IAzureNode, IAzureParentTreeItem, IAzureTreeItem } from 'vscode-azureextensionui';
+
+export class ConnectionTreeItem implements IAzureParentTreeItem {
```
EricJizbaMSFT: nit: This class/file should be named Connection**s**TreeItem

src/explorer/WebAppTreeItem.ts
```diff
@@ -48,7 +51,16 @@ export class WebAppTreeItem extends SiteTreeItem {
         const tier: string = String(appServicePlan.sku!.tier);
         // tslint:disable-next-line:no-non-null-assertion
         this.deploymentSlotsNode = /^(basic|free|shared)$/i.test(tier) ? new DeploymentSlotsNATreeItem(tier, appServicePlan.id!) : new DeploymentSlotsTreeItem(this.client);
-        return [this.deploymentSlotsNode, this.folderNode, this.logFolderNode, this.webJobsNode, this.appSettingsNode];
+        const nodes = [this.deploymentSlotsNode, this.folderNode, this.logFolderNode, this.webJobsNode, this.appSettingsNode];
+        if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length === 1) {
+            const currentWorkspace = vscode.workspace.workspaceFolders[0];
+            const workspaceConfig = vscode.workspace.getConfiguration(constants.extensionPrefix, currentWorkspace.uri);
```
EricJizbaMSFT: You only need to pass in a uri to getConfiguration if the setting is unique per folder. For example, `defaultWebAppToDeploy` is unique for each project, so we listed the scope [here](https://github.com/Microsoft/vscode-azureappservice/blob/master/package.json#L650) as "resource".

This setting doesn't need to be unique per folder in my opinion and thus should have the default scope. Which means you don't need to pass in the uri

src/explorer/ConnectionTreeItem.ts
```diff
@@ -0,0 +1,33 @@
...
+    public get iconPath(): { light: string, dark: string } {
+        const iconName = 'Connections_16x.svg';
+        return {
+            light: path.join(__filename, '..', '..', '..', '..', 'resources', 'light', iconName),
+            dark: path.join(__filename, '..', '..', '..', '..', 'resources', 'dark', iconName)
+        };
+    }
+
+    public async loadMoreChildren(_node: IAzureNode<IAzureTreeItem>, _clearCache: boolean): Promise<IAzureTreeItem[]> {
+        throw new Error('error');
```
EricJizbaMSFT: This is typically what we do: `throw new Error("Method not implemented.");`

src/explorer/ConnectionTreeItem.ts
Omelianenko: What do you think about "Connection(s)Provider" instead of "ConnectionsTreeItem" and "CoonectionTreeItem" for nodes under this "ConnectionProvider"(similary to WebAppProvider and WebAppTreeItem classes).
1100am
Or it's better to use "ConnectionsTreeItem" and "ConnectionTreeItem"(similary to "DeploymentSlotsTreeItem" and "DeploymentSlotTreeItem"?

EricJizbaMSFT: WebAppProvider is different. You'll notice it only implements `ChildProvider`, not `IAzureParentTreeItem`

I think ConnectionsTreeItem makes the most sense. For nodes under that, I would suggest "CosmosDBTreeItem" and "StorageTreeItem"

Omelianenko:
Dark theme:
dark-theme
darktheme
Light theme:
light-theme
lighttheme

src/explorer/ConnectionsTreeItem.ts
```diff
@@ -24,7 +24,7 @@ export class ConnectionTreeItem implements IAzureParentTreeItem {
     }
 
     public async loadMoreChildren(_node: IAzureNode<IAzureTreeItem>, _clearCache: boolean): Promise<IAzureTreeItem[]> {
-        throw new Error('error');
+        throw new Error('Method not implemented.');
```
EricJizbaMSFT: This should return 'CosmosDBTreeItem' now, correct?

src/explorer/WebAppTreeItem.ts
```diff
@@ -48,7 +51,15 @@ export class WebAppTreeItem extends SiteTreeItem {
         const tier: string = String(appServicePlan.sku!.tier);
         // tslint:disable-next-line:no-non-null-assertion
         this.deploymentSlotsNode = /^(basic|free|shared)$/i.test(tier) ? new DeploymentSlotsNATreeItem(tier, appServicePlan.id!) : new DeploymentSlotsTreeItem(this.client);
-        return [this.deploymentSlotsNode, this.folderNode, this.logFolderNode, this.webJobsNode, this.appSettingsNode];
+        const nodes = [this.deploymentSlotsNode, this.folderNode, this.logFolderNode, this.webJobsNode, this.appSettingsNode];
+        if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length === 1) {
```
EricJizbaMSFT: Remove this check about `vscode.workspace.workspaceFolders` since you're not using it anymore

src/explorer/WebAppTreeItem.ts
```diff
@@ -48,7 +51,15 @@ export class WebAppTreeItem extends SiteTreeItem {
...
+            const workspaceConfig = vscode.workspace.getConfiguration(constants.extensionPrefix);
+            if (workspaceConfig.get(constants.configurationSettings.enableConnectionsNode)) {
+                this.connectionsNode = new ConnectionsTreeItem(this.client);
```
EricJizbaMSFT: You only need to create this node once in the constructor (like the other nodes are doing), rather than creating it every time "loadMoreChildren" is called

src/explorer/WebAppTreeItem.ts
```diff
@@ -25,13 +27,19 @@ export class WebAppTreeItem extends SiteTreeItem {
...
 
     constructor(client: SiteClient) {
         super(client);
         this.folderNode = new FolderTreeItem(this.client, 'Files', "/site/wwwroot");
         this.logFolderNode = new FolderTreeItem(this.client, 'Log Files', '/LogFiles', 'logFolder');
         this.webJobsNode = new WebJobsTreeItem(this.client);
         this.appSettingsNode = new AppSettingsTreeItem(this.client);
+
+        const workspaceConfig = vscode.workspace.getConfiguration(constants.extensionPrefix);
+        if (workspaceConfig.get(constants.configurationSettings.enableConnectionsNode)) {
```
EricJizbaMSFT: So I wanted you to move _creating_ the node into the constructor, but checking against the workspaceConfig still needs to happen inside of `loadMoreChildren`. The node should always be defined, whether or not it's returned in the array of children below

For example, follow these repro steps:
1. Make sure `enableConnectionsNode` is false
1. Expand a web app
1. Set `enableConnectionsNode` to true
1. Refresh that web app node

The node should show up after the refresh, but I don't think it will with your current code.

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { SiteClient } from 'vscode-azureappservice';
+import { IAzureNode, IAzureParentTreeItem, IAzureTreeItem } from 'vscode-azureextensionui';
+
+export class CosmosDBTreeItem implements IAzureParentTreeItem {
+    public static contextValue: string = 'Connection';
```
EricJizbaMSFT: I suggest making this 'CosmosDBConnection'

src/explorer/ConnectionsTreeItem.ts
```diff
@@ -0,0 +1,36 @@
...
+    public get iconPath(): { light: string, dark: string } {
+        const iconName = 'Connections_16x.svg';
+        return {
+            light: path.join(__filename, '..', '..', '..', '..', 'resources', 'light', iconName),
+            dark: path.join(__filename, '..', '..', '..', '..', 'resources', 'dark', iconName)
+        };
+    }
+
+    public async loadMoreChildren(_node: IAzureNode<IAzureTreeItem>, _clearCache: boolean): Promise<IAzureTreeItem[]> {
+        this.cosmosDBNode = new CosmosDBTreeItem(this.client);
```
EricJizbaMSFT: Same pattern here as in the WebAppTreeItem. You only need to create `cosmosDBNode` once in the constructor. loadMoreChildren should just be `return [this.cosmosDBNode];`

PR #604, opened 2018-09-06
Added refresh buttons to own groups

Omelianenko:
before:
picture
now:
now

Fixed for Deployment Slots, WebJobs, and LogFiles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -516,7 +516,7 @@
                 {
                     "command": "appService.Refresh",
                     "when": "view == azureAppService && viewItem == subFolder",
-                    "group": "1_folderGeneralCommands@1"
+                    "group": "1_subFolderRefreshCommands@1"
                 },
                 {
                     "command": "appService.Refresh",
--- a/package.json
+++ b/package.json
@@ -426,7 +426,7 @@
                 {
                     "command": "appService.Refresh",
                     "when": "view == azureAppService && viewItem == deploymentSlots",
-                    "group": "1_deploymentSlotsGeneralCommands@3"
+                    "group": "2_deploymentSlotsGeneralCommands@1"
                 },
                 {
                     "command": "appService.OpenInPortal",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -516,12 +516,12 @@
                 {
                     "command": "appService.Refresh",
                     "when": "view == azureAppService && viewItem == subFolder",
-                    "group": "1_folderGeneralCommands@2"
+                    "group": "1_folderGeneralCommands@1"
```
Omelianenko: Not sure about this

EricJizbaMSFT: `subFolder` is any folder below the "Files" node. For example "HttpTriggerJS" in my case below:
screen shot 2018-09-06 at 9 25 45 am

I believe "subFolderRefreshCommand" might be more accurate. But honestly the numbers are the only part you _really_ need. The "_subFolderRefreshCommand" part is just for our own readability and I find it more annoying to maintain than helpful. In the the functions repo, I don't use that part at all: https://github.com/Microsoft/vscode-azurefunctions/blob/master/package.json#L373

In other words, I don't really care what put as long as it works :) Please post one screenshot for each case that you changed

package.json
```diff
@@ -531,7 +531,7 @@
                 {
                     "command": "appService.Refresh",
                     "when": "view == azureAppService && viewItem == webJobs",
-                    "group": "1_webJobsGeneralCommands@2"
+                    "group": "2_webJobsGeneralCommands@1"
```
Omelianenko: webjobsgeneral

package.json
```diff
@@ -516,12 +516,12 @@
...
                     "command": "appService.Refresh",
                     "when": "view == azureAppService && viewItem == subFolder",
-                    "group": "1_folderGeneralCommands@2"
+                    "group": "1_subFolderRefreshCommands@1"
                 },
                 {
                     "command": "appService.Refresh",
                     "when": "view == azureAppService && viewItem == logFolder",
-                    "group": "1_folderGeneralCommands@2"
+                    "group": "2_folderGeneralCommands@1"
```
Omelianenko: logfoldergeneral

package.json
```diff
@@ -516,12 +516,12 @@
                 {
                     "command": "appService.Refresh",
                     "when": "view == azureAppService && viewItem == subFolder",
-                    "group": "1_folderGeneralCommands@2"
+                    "group": "1_subFolderRefreshCommands@1"
```
Omelianenko: subfolderrefresh

package.json
```diff
@@ -426,7 +426,7 @@
                 {
                     "command": "appService.Refresh",
                     "when": "view == azureAppService && viewItem == deploymentSlots",
-                    "group": "1_deploymentSlotsGeneralCommands@3"
+                    "group": "2_deploymentSlotsGeneralCommands@1"
```
Omelianenko: deployslotgeneral

PR #616, opened 2018-09-18
Added command to install Cosmos extension

### LINKED ISSUE #583: Add command to install Cosmos Extension
If the user does not have the Cosmos DB extension installed, we should add a command under the node described here: https://github.com/Microsoft/vscode-azureappservice/issues/582

Something like "Install Cosmos DB extension...". It will either link them to the VS Code marketplace or just install it directly for them

Omelianenko:
Fixes #583


src/extension.ts
```diff
@@ -312,6 +312,7 @@ export function activate(context: vscode.ExtensionContext): void {
             vscode.window.showInformationMessage(`File logging has already been enabled for ${siteTreeItem.client.fullName}.`);
         }
     });
+    registerCommand('appService.OpenCosmosDBExtension', () => vscode.commands.executeCommand('extension.open', 'ms-azuretools.vscode-cosmosdb'));
```
EricJizbaMSFT: This works without editing the package.json? I feel like you need to add the `OpenCosmosDBExtension` command to package.json...

src/extension.ts
```diff
@@ -312,6 +312,7 @@ export function activate(context: vscode.ExtensionContext): void {
...
 }
```
EricJizbaMSFT: VS Code _just_ added the 'extension.open' command, right? We need to handle the situation when that command doesn't exist for people on older versions of VS Code.

There's two options:
1. Bump the minimum VS Code version here: https://github.com/Microsoft/vscode-azureappservice/blob/master/package.json#L10
1. Have backup logic if the command doesn't exist (maybe just open the browser?)

I vote option 2 because this doesn't seem like a good enough reason to bump the min VS Code version. Although don't spend a lot of time on the "backup logic" because it really is just "backup" logic that shouldn't happen very often

src/extension.ts
Omelianenko: It works, will add it to package

package.json
```diff
@@ -217,6 +217,11 @@
                 "command": "appService.EnableFileLogging",
                 "title": "Enable File Logging",
                 "category": "Azure App Service"
+            },
+            {
+                "command": "appService.InstallCosmosDBExtension",
```
EricJizbaMSFT: Add this to the 'commandPalette' section with 'never' specified

src/extension.ts
```diff
@@ -312,7 +312,14 @@ export function activate(context: vscode.ExtensionContext): void {
             vscode.window.showInformationMessage(`File logging has already been enabled for ${siteTreeItem.client.fullName}.`);
         }
     });
-    registerCommand('appService.OpenCosmosDBExtension', () => vscode.commands.executeCommand('extension.open', 'ms-azuretools.vscode-cosmosdb'));
+    registerCommand('appService.InstallCosmosDBExtension', async () => {
+        const listOfCommands = await vscode.commands.getCommands();
+        if (listOfCommands.find((x: string) => x === 'extension.open')) {
```
EricJizbaMSFT: nit: Make a `const` for 'extension.open' instead of copying the same string in two different lines

src/extension.ts
```diff
@@ -312,7 +312,14 @@ export function activate(context: vscode.ExtensionContext): void {
...
+            vscode.commands.executeCommand('extension.open', 'ms-azuretools.vscode-cosmosdb');
+        } else {
+            vscode.commands.executeCommand('vscode.open', vscode.Uri.parse('https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-cosmosdb'));
```
EricJizbaMSFT: We usually just use 'opn' instead of 'vscode.Open'. See here for an example: https://github.com/Microsoft/vscode-azureappservice/blob/c97d81c60f985e629493ee3ecdc6741e67c4e7a7/src/explorer/SiteTreeItem.ts#L61

PR #626, opened 2018-09-28
Added CosmosDB to Connections node

### LINKED ISSUE #584: Add command to connect cosmos db account
Assuming the Cosmos DB extension is installed, this will prompt the user to pick their Cosmos DB account (or attached account). It will copy the connection string and insert it as an application setting on the web app. It will also insert a setting in the local VS Code workspace (This is how we persist the connection in the App Service view)

This needs to leverage the Cosmos DB extension, but we don't have an established pattern for working across extensions. This still needs to be determined.

Omelianenko:
Fixes #584

### DIFF
--- a/package.json
+++ b/package.json
@@ -225,7 +225,7 @@
             },
             {
                 "command": "appService.revealConnection",
-                "title": "Reveal into Cosmos DB",
+                "title": "Reveal in Cosmos DB extension",
                 "category": "Azure App Service"
             },
             {
@@ -235,7 +235,7 @@
             },
             {
                 "command": "appService.DeleteCosmosDBConnection",
-                "title": "Delete",
+                "title": "Remove Connection",
                 "category": "Azure App Service"
             }
         ],
@@ -605,27 +605,27 @@
                 },
                 {
                     "command": "appService.Refresh",
-                    "when": "view == azureAppService && viewItem == сonnectionsTreeItem",
+                    "when": "view == azureAppService && viewItem == connections",
                     "group": "1@1"
                 },
                 {
                     "command": "appService.AddCosmosDBConnection",
-                    "when": "view == azureAppService && viewItem == сosmosDBConnection",
+                    "when": "view == azureAppService && viewItem == сosmosDBConnections",
                     "group": "1@1"
                 },
                 {
                     "command": "appService.Refresh",
-                    "when": "view == azureAppService && viewItem == сosmosDBConnection",
+                    "when": "view == azureAppService && viewItem == сosmosDBConnections",
                     "group": "2@2"
                 },
                 {
                     "command": "appService.revealConnection",
-                    "when": "view == azureAppService && viewItem == connectionAccountDatabase",
+                    "when": "view == azureAppService && viewItem == cosmosDBDatabase",
                     "group": "1@1"
                 },
                 {
                     "command": "appService.DeleteCosmosDBConnection",
-                    "when": "view == azureAppService && viewItem == connectionAccountDatabase",
+                    "when": "view == azureAppService && viewItem == cosmosDBDatabase",
                     "group": "2@1"
                 }
             ],
@@ -720,6 +720,9 @@
                 "appService.connections": {
                     "type": "array",
                     "description": "Array of web apps with list of all theirs connections",
+                    "required": [
+                        "webAppId"
+                    ],
                     "items": {
                         "type": "object",
                         "description": "All connections of specific web app.",
@@ -727,11 +730,11 @@
                         "properties": {
                             "webAppId": {
                                 "type": "string",
-                                "description": "Web app identificator."
+                                "description": "Web app id."
                             },
                             "cosmosDB": {
                                 "type": "array",
-                                "description": "Array of connections identificators.",
+                                "description": "Array of Cosmos DB database ids.",
                                 "items": {
                                     "type": "string",
                                     "description": "Connection identificator"
--- a/package.json
+++ b/package.json
@@ -234,7 +234,7 @@
                 "category": "Azure App Service"
             },
             {
-                "command": "appService.DeleteCosmosDBConnection",
+                "command": "appService.RemoveCosmosDBConnection",
                 "title": "Remove Connection",
                 "category": "Azure App Service"
             }
@@ -327,7 +327,7 @@
                     "when": "never"
                 },
                 {
-                    "command": "appService.DeleteCosmosDBConnection",
+                    "command": "appService.RemoveCosmosDBConnection",
                     "when": "never"
                 }
             ],
@@ -624,7 +624,7 @@
                     "group": "1@1"
                 },
                 {
-                    "command": "appService.DeleteCosmosDBConnection",
+                    "command": "appService.RemoveCosmosDBConnection",
                     "when": "view == azureAppService && viewItem == cosmosDBDatabase",
                     "group": "2@1"
                 }
@@ -737,7 +737,7 @@
                                 "description": "Array of Cosmos DB database ids.",
                                 "items": {
                                     "type": "string",
-                                    "description": "Connection identificator"
+                                    "description": "Connection id"
                                 }
                             }
                         }
--- a/package.json
+++ b/package.json
@@ -222,6 +222,21 @@
                 "command": "appService.InstallCosmosDBExtension",
                 "title": "Install Cosmos DB Extension",
                 "category": "Azure App Service"
+            },
+            {
+                "command": "appService.revealConnection",
+                "title": "Reveal into Cosmos DB",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "appService.AddCosmosDBConnection",
+                "title": "Add New Connection...",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "appService.DeleteCosmosDBConnection",
+                "title": "Delete",
+                "category": "Azure App Service"
             }
         ],
         "viewsContainers": {
@@ -301,6 +316,18 @@
                 },
                 {
                     "command": "appService.InstallCosmosDBExtension",
+                    "when": "never"
+                },
+                {
+                    "command": "appService.revealConnection",
+                    "when": "never"
+                },
+                {
+                    "command": "appService.AddCosmosDBConnection",
+                    "when": "never"
+                },
+                {
+                    "command": "appService.DeleteCosmosDBConnection",
                     "when": "never"
                 }
             ],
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -575,6 +602,31 @@
...
+                    "group": "1@1"
+                },
+                {
+                    "command": "appService.AddCosmosDBConnection",
+                    "when": "view == azureAppService && viewItem == сosmosDBConnection",
+                    "group": "1@1"
+                },
+                {
+                    "command": "appService.Refresh",
+                    "when": "view == azureAppService && viewItem == сosmosDBConnection",
```
EricJizbaMSFT: Can we rename 'сosmosDBConnection' to 'сosmosDBConnections' and 'connectionAccountDatabase' to 'cosmosDBDatabase'

package.json
```diff
@@ -664,6 +716,29 @@
...
+                        "description": "All connections of specific web app.",
+                        "additionalProperties": false,
+                        "properties": {
+                            "webAppId": {
+                                "type": "string",
+                                "description": "Web app identificator."
+                            },
+                            "cosmosDB": {
+                                "type": "array",
+                                "description": "Array of connections identificators.",
```
EricJizbaMSFT: `Array of Cosmos DB database ids`

package.json
```diff
@@ -664,6 +716,29 @@
...
+                    "type": "array",
+                    "description": "Array of web apps with list of all theirs connections",
+                    "items": {
+                        "type": "object",
+                        "description": "All connections of specific web app.",
+                        "additionalProperties": false,
+                        "properties": {
+                            "webAppId": {
+                                "type": "string",
+                                "description": "Web app identificator."
```
EricJizbaMSFT: I don't think 'identificator' is a word - it's just 'identifier'. In any case, I would just use 'id'

package.json
```diff
@@ -664,6 +716,29 @@
...
                     "default": false,
                     "description": "Enables the Advanced Web App Creation which will prompt for app name, location, website OS, runtime, resource group, and app service plan."
+                },
+                "appService.connections": {
+                    "type": "array",
+                    "description": "Array of web apps with list of all theirs connections",
+                    "items": {
+                        "type": "object",
+                        "description": "All connections of specific web app.",
+                        "additionalProperties": false,
```
EricJizbaMSFT: Make sure 'webAppId' is required. I think you just need to set `"required": ["webAppId"]`

src/extension.ts
```diff
@@ -322,6 +325,11 @@ export function activate(context: vscode.ExtensionContext): void {
             opn('https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-cosmosdb');
         }
     });
+    registerCommand('appService.AddCosmosDBConnection', async (node: IAzureNode<CosmosDBTreeItem>) => {
+        const connectionToAdd = <string>await vscode.commands.executeCommand('cosmosDB.api.getDatabase');
```
EricJizbaMSFT: Move everything into `addCosmosDBConnection.ts`, including this line

src/commands/connections/deleteCosmosDBConnection.ts
```diff
@@ -0,0 +1,29 @@
...
+import { IConnections } from './IConnections';
+
+export async function deleteCosmosDBConnection(node: IAzureNode<ConnectionAccountDatabaseTreeItem>): Promise<void> {
+    const connectionToDelete = node.treeItem.connectionId;
+    const workspaceConfig = vscode.workspace.getConfiguration(constants.extensionPrefix);
+    const allConnections = workspaceConfig.get<IConnections[]>(constants.configurationSettings.connections, []);
+
+    let connectionsUnit = allConnections.find((x: IConnections) => x.webAppId === node.treeItem.client.id);
+    // tslint:disable-next-line:strict-boolean-expressions
+    connectionsUnit = connectionsUnit || <IConnections>{};
```
EricJizbaMSFT: If connectionsUnit isn't defined, then you don't need to remove anything. Same for connectionsUnit.cosmosDB

I would just wrap the rest of this code in `if (connectionsUnit && connectionsUnit.cosmosDB) {`

src/explorer/ConnectionsTreeItem.ts
```diff
@@ -9,13 +9,14 @@ import { IAzureNode, IAzureParentTreeItem, IAzureTreeItem } from 'vscode-azureex
...
 export class ConnectionsTreeItem implements IAzureParentTreeItem {
-    public static contextValue: string = 'Connections';
+    public static contextValue: string = 'сonnectionsTreeItem';
     public readonly contextValue: string = ConnectionsTreeItem.contextValue;
     public readonly label: string = 'Connections';
     public readonly cosmosDBNode: IAzureTreeItem;
 
     constructor(readonly client: SiteClient) {
         this.cosmosDBNode = new CosmosDBTreeItem(this.client);
+        //client.updateApplicationSettings();
```
EricJizbaMSFT: Remove this comment

src/explorer/ConnectionsTreeItem.ts
```diff
@@ -9,13 +9,14 @@ import { IAzureNode, IAzureParentTreeItem, IAzureTreeItem } from 'vscode-azureex
 import { CosmosDBTreeItem } from './CosmosDBTreeItem';
 
 export class ConnectionsTreeItem implements IAzureParentTreeItem {
-    public static contextValue: string = 'Connections';
+    public static contextValue: string = 'сonnectionsTreeItem';
```
EricJizbaMSFT: I liked the old contextValue better. We could add 'TreeItem' to every single context value if we wanted and it would just get out of hand

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -6,26 +6,36 @@
...
             return [{
-                contextValue: 'InstallcosmosDBExtension',
+                contextValue: 'InstallCosmosDBExtension',
                 label: 'Install Cosmos DB Extension...',
                 commandId: 'appService.InstallCosmosDBExtension',
                 isAncestorOf: () => { return false; }
             }];
         }
-        throw new Error('Method not implemented.');
+        const workspaceConfig = vscode.workspace.getConfiguration(constants.extensionPrefix);
```
EricJizbaMSFT: If there are no connections, you should return a custom node that says "Add Cosmos DB Connection..." similar to your custom node above for 'Install Cosmos DB Extension...'

src/explorer/ConnectionAccountDatabaseTreeItem.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    constructor(readonly client: SiteClient, readonly connectionId: string) {
+        this.label = this.getLabel(connectionId);
+    }
+
+    private getLabel(id: string): string {
+        // tslint:disable-next-line:strict-boolean-expressions
+        const items = this.parseCosmos(id) || this.parseAttached(id) || undefined;
+        if (!items) {
+            throw new Error('Id doesn\'t match any of known connection patterns');
```
EricJizbaMSFT: How about "Failed to parse connection id."

src/explorer/ConnectionAccountDatabaseTreeItem.ts
```diff
@@ -0,0 +1,42 @@
...
+    public readonly contextValue: string = ConnectionAccountDatabaseTreeItem.contextValue;
+    public readonly label: string;
+
+    constructor(readonly client: SiteClient, readonly connectionId: string) {
+        this.label = this.getLabel(connectionId);
+    }
+
+    private getLabel(id: string): string {
+        // tslint:disable-next-line:strict-boolean-expressions
+        const items = this.parseCosmos(id) || this.parseAttached(id) || undefined;
```
EricJizbaMSFT: Change the tslint rule here to have 'allow-null-union' and you shouldn't need `|| undefined`

https://github.com/Microsoft/vscode-azureappservice/blob/master/tslint.json#L93

I've been meaning to change that rule anyways.

package.json
```diff
@@ -664,6 +716,32 @@
...
+                            "webAppId": {
+                                "type": "string",
+                                "description": "Web app id."
+                            },
+                            "cosmosDB": {
+                                "type": "array",
+                                "description": "Array of Cosmos DB database ids.",
+                                "items": {
+                                    "type": "string",
+                                    "description": "Connection identificator"
```
EricJizbaMSFT: nit: You missed one 'identificator'

package.json
```diff
@@ -222,6 +222,21 @@
...
+                "title": "Reveal in Cosmos DB extension",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "appService.AddCosmosDBConnection",
+                "title": "Add New Connection...",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "appService.DeleteCosmosDBConnection",
```
EricJizbaMSFT: nit: Rename the command to 'RemoveCosmosDBConnection' as well (and the underlying code)

src/commands/connections/addCosmosDBConnection.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    // tslint:disable-next-line:strict-boolean-expressions
+    connectionsUnit.cosmosDB = connectionsUnit.cosmosDB || [];
+    if (!connectionsUnit.cosmosDB.find((x: string) => x === connectionToAdd)) {
+        connectionsUnit.cosmosDB.push(connectionToAdd);
+        workspaceConfig.update(constants.configurationSettings.connections, allConnections);
+        if (node.treeItem.contextValue === 'AddCosmosDBConnection') {
+            // tslint:disable-next-line:no-non-null-assertion
+            await node.parent!.refresh();
```
Omelianenko: Is this a good way to resolve the situation when there are no Cosmos DB connections and it provides node as 'Add Cosmos DB Connection' button?

EricJizbaMSFT: Soooo I completely forgot, but you shouldn't be handling refresh yourself. You need to implement the `createChild` method inside `CosmosDBTreeItem`. See here as an example: https://github.com/Microsoft/vscode-azureappservice/blob/master/src/explorer/DeploymentSlotsTreeItem.ts#L59 It will handle refreshing and other UI stuff for you. Once you've implemented that, you just need to call `node.createChild()`

Similarly, the remove code should be implemented in `deleteTreeItem` in `CosmosDBDatabase` tree item. See here for an example: https://github.com/Microsoft/vscode-azureappservice/blob/master/src/explorer/SiteTreeItem.ts#L64

You can fix that stuff now (and let me review the change before merging) or you can merge your existing code as-is and fix in a separate PR.

Omelianenko: Working on the other thing now, so I'll merge it and will fix it after tomorrow's demo

Issue #623, opened 2018-09-25
Many notifications when deploying a web app


chrisdias:
I get *a lot* of prompts when deploying an app to App Service.  Do we need all of these?

Here' what I get on first deployment:

![image](https://user-images.githubusercontent.com/1487073/46043312-b1554780-c0e5-11e8-91ad-e645cb226147.png)

Then I get the next 3 prompts:

![image](https://user-images.githubusercontent.com/1487073/46043341-c03bfa00-c0e5-11e8-910b-cea7338bd6c1.png)

Then when done i get one more prompt:

![image](https://user-images.githubusercontent.com/1487073/46043433-01340e80-c0e6-11e8-8665-93add8792e3b.png)

So that's at least 7 notifications when deploying an app.

Why not 1 while creating as an info, 1 when it is done, and then a final one that asks if you want to "Always deploy the website..."

nturinski:
Closed via: https://github.com/Microsoft/vscode-azuretools/pull/301

PR #637, opened 2018-10-10
Remove additional showInformationWindow calls

nturinski:
Too many prompts that were cluttering up the UI


nturinski:
After looking at Chris's issue: https://github.com/Microsoft/vscode-azureappservice/issues/623, I thought maybe we should clean up the rest.  But if you're fine with it, I am too :)

PR #655, opened 2018-10-15
Added reveal command

### LINKED ISSUE #568: Connecting AppSevice web apps to CosmosDB
Target scenario for MVP: Connecting a CosmosDB account includes any of the CosmosDB APIs along with external connections. It's important to note that the connection needs to be as open as possible with the database connection to enable users to choose a service other than CosmosDB. However, if the user already has the CosmosDB extension installed and is connected to external accounts, we should include it in the connection selection.  Some early thoughts:

1. Expand Connected Experience (a place hold name) node then expand CosmosDB node
2. When no accounts are connected, only the option to connect is present "Connect CosmosDB Account"
3. Clicking the message begins the connection process
    - If the user doesn't have the CosmosDB extension installed, they are prompted to install it
    - If they do, the account selection quick-pick opens to show existing accounts
4. User chooses account and database
5. The connection string for the CosmosDB account/database is configured as an app setting for the Web App

Once services are connected, there are some other help we can provide. For example:
- Picking the correct SDK module
- Scaffolding authentication code for the SDK if necessary
- Scaffolding common code (e.g. connecting to a Mongo DB using the connection string, etc)

Omelianenko:
Fix #568

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -20,6 +20,7 @@ import { disableRemoteDebug } from './commands/remoteDebug/disableRemoteDebug';
 import { startRemoteDebug } from './commands/remoteDebug/startRemoteDebug';
 import { startStreamingLogs } from './commands/startStreamingLogs';
 import { swapSlots } from './commands/swapSlots';
+import { CosmosDBDatabase } from './explorer/CosmosDBDatabase';
 import { DeploymentSlotsNATreeItem, DeploymentSlotsTreeItem, ScaleUpTreeItem } from './explorer/DeploymentSlotsTreeItem';
 import { DeploymentSlotTreeItem } from './explorer/DeploymentSlotTreeItem';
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -323,9 +324,7 @@ export function activate(context: vscode.ExtensionContext): void {
     });
     registerCommand('appService.AddCosmosDBConnection', addCosmosDBConnection);
     registerCommand('appService.RemoveCosmosDBConnection', removeCosmosDBConnection);
-    registerCommand('appService.revealConnection', () => {
-        throw new Error('Not implemented yet.');
-    });
+    registerCommand('appService.RevealConnection', async (node: CosmosDBDatabase) => vscode.commands.executeCommand('cosmosDB.revealTreeItem', node.connectionId));
```
EricJizbaMSFT: The cosmos command should have `.api.` in the name just like the other api calls

PR #658, opened 2018-10-16
Refactored Add and Delete Connection functions

Omelianenko:


src/explorer/CosmosDBDatabase.ts
```diff
@@ -25,26 +27,39 @@ export class CosmosDBDatabase extends AzureTreeItem<ISiteTreeRoot> {
...
+        const workspaceConfig = vscode.workspace.getConfiguration(constants.extensionPrefix);
+        const connections = workspaceConfig.get<IConnections[]>(constants.configurationSettings.connections, []);
+        const connectionsUnit = connections.find((x: IConnections) => x.webAppId === this.root.client.id);
+        if (connectionsUnit && connectionsUnit.cosmosDB) {
+            const indexToDelete = connectionsUnit.cosmosDB.findIndex((x: string) => x === connectionToDelete);
+            if (indexToDelete > -1) {
+                connectionsUnit.cosmosDB.splice(indexToDelete, 1);
+                workspaceConfig.update(constants.configurationSettings.connections, connections);
+                // tslint:disable-next-line:no-non-null-assertion
+                await this.parent!.refresh();
```
EricJizbaMSFT: You're refreshing the parent node in `extension.ts`, so you don't need to do it here as well. Same feedback for createChildImpl

src/commands/connections/updateWebAppSettings.ts
```diff
@@ -0,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AppSettingsTreeItem } from "vscode-azureappservice";
+import { ext } from "../../extensionVariables";
+
+export async function updateWebAppSetting(webAppId: string, appSettingToUpdate: string, value: string): Promise<void> {
+    const appSettItem = <AppSettingsTreeItem | undefined>await ext.tree.findTreeItem(webAppId + String('/application'));
```
EricJizbaMSFT: I don't like using `findTreeItem` with `/application` hard-coded. Instead, you should use the node's parent to eventually get the app settings tree item. The default type for a parent is `AzureParentTreeItem | undefined`, but you can make that type more specific like this:
```
export class CosmosDBDatabase extends AzureTreeItem<ISiteTreeRoot> {
    ...
    public readonly parent: CosmosDBTreeItem;

    constructor(parent: CosmosDBTreeItem, ...) {
        super(parent);
        ...
    }
}
```

Once you get to the WebAppTreeItem parent, you can access `appSettingsNode` directly: https://github.com/Microsoft/vscode-azureappservice/blob/master/src/explorer/WebAppTreeItem.ts#L26

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -48,6 +49,54 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+            // Don't wait
+            vscode.window.showInformationMessage(`Database "${createdDatabase.label}" connected to Web App "${this.root.client.fullName}". Created "${appSettingsToUpdate}" App Setting.`, ok, showDatabase).then(async (result: vscode.MessageItem | undefined) => {
+                if (result === showDatabase) {
+                    vscode.commands.executeCommand('appService.RevealConnection', createdDatabase);
+                }
+            });
+
+            return createdDatabase;
+        }
+        throw new Error("Impossible to have more than one CosmosDB connection for one web app!");
```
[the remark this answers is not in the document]
Omelianenko: Ok, I'll make a different PR for this, but in case If we continue to allow to have multiple connections should we:
1. just ignore if a user wants to add a connection
2. throw an error that for now only one connection possible to add
3. add this connection:
- MONGO_URL stays the same
- MONGO_URL has the value of the last added connection
And what should we do in case if a user deleted a connection that was the value of MONGO_URL, should connection stay the same or we need to delete it or replace with some other connection?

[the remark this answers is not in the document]
Omelianenko: Actually, it was a bug. I go there only if the database is already connected and exists in workspace settings. 

Anyway, I need to return/throw something in this func(it was void before refactoring), so I don't really can leave it without any changes.

So, what do you think is the best way to resolve this? Here are options within I choose:
1. Show notification that database connected successfully
2. Show notification(or error) that database already connected
3. Just do nothing (throw new UserCancelledError) 

For now, I just going to leave 3 variant, I think it's equivalent to 'no major behavior change'

src/commands/connections/updateWebAppSettings.ts
```diff
@@ -0,0 +1,17 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { WebAppTreeItem } from "src/explorer/WebAppTreeItem";
+import { ext } from "../../extensionVariables";
+
+export async function updateWebAppSetting(webAppId: string, appSettingToUpdate: string, value: string): Promise<void> {
+    const webApp = <WebAppTreeItem | undefined>await ext.tree.findTreeItem(webAppId);
```
EricJizbaMSFT: This is better, but I don't think you quite understood what I meant. You shouldn't need to do `findTreeItem` at all. I used `CosmosDBDatabase` as an example, but it's really `CosmosDBTreeItem.ts` and `ConnectionsTreeItem.ts` where you need to specify the type of the parent. Then you can just do `this.parent.parent.appSettingsNode` within CosmosDBTreeItem.ts

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -48,6 +49,47 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+            // Don't wait
+            vscode.window.showInformationMessage(`Database "${createdDatabase.label}" connected to Web App "${this.root.client.fullName}". Created "${appSettingsToUpdate}" App Setting.`, ok, showDatabase).then(async (result: vscode.MessageItem | undefined) => {
+                if (result === showDatabase) {
+                    vscode.commands.executeCommand('appService.RevealConnection', createdDatabase);
+                }
+            });
+
+            return createdDatabase;
+        }
+        throw new UserCancelledError();
```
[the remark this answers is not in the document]
Omelianenko: Looks like for this I'll need to change the return type of createChildImpl [here](https://github.com/Microsoft/vscode-azuretools/blob/17f4cbd719c1646b63134213aedb18e3fa78d2c6/ui/src/treeDataProvider/AzureParentTreeItem.ts#L47).

The error I get:
```
[ts]
Property 'createChildImpl' in type 'CosmosDBTreeItem' is not assignable to the same property in base type 'AzureParentTreeItem<ISiteTreeRoot>'.
  Type '(showCreatingTreeItem: (label: string) => void) => Promise<AzureTreeItem<ISiteTreeRoot> | undefined>' is not assignable to type '(showCreatingTreeItem: (label: string) => void, userOptions?: any) => Promise<AzureTreeItem<ISiteTreeRoot>>'.
    Type 'Promise<AzureTreeItem<ISiteTreeRoot> | undefined>' is not assignable to type 'Promise<AzureTreeItem<ISiteTreeRoot>>'.
      Type 'AzureTreeItem<ISiteTreeRoot> | undefined' is not assignable to type 'AzureTreeItem<ISiteTreeRoot>'.
        Type 'undefined' is not assignable to type 'AzureTreeItem<ISiteTreeRoot>'.
(method) CosmosDBTreeItem.createChildImpl(showCreatingTreeItem: (label: string) => void): Promise<AzureTreeItem<ISiteTreeRoot> | undefined>
Implement this if you want the 'create' option to show up in the tree picker. Should not be called directly

```

src/commands/connections/updateWebAppSettings.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AppSettingsTreeItem } from "vscode-azureappservice";
+
+export async function updateWebAppSetting(appSettItem: AppSettingsTreeItem, appSettingToUpdate: string, value: string): Promise<void> {
+    await appSettItem.editSettingItem(appSettingToUpdate, appSettingToUpdate, value);
```
EricJizbaMSFT: IMO this isn't worth it's own file anymore. You only did it because you didn't want to copy the code in the future when you support storage right? But it's only two lines of code now, so copying seems fine

Omelianenko: I did it also because of a future using this method for others type of connections. For now, I'm planning to use it only for SQL databases and not sure that I actually will need this as the method, so it's not hard to put it inside of the CosmosDBTreeItem and move into the method back when it's really needed.

PR #659, opened 2018-10-16
Improved connection settings description

Omelianenko:


package.json
```diff
@@ -228,11 +228,6 @@
                 "title": "Reveal in Cosmos DB extension",
                 "category": "Azure App Service"
             },
-            {
```
EricJizbaMSFT: I don't see how this is redundant. If I already have a connection, this is the only entry point for adding another connection

Omelianenko: We've decided to support only one cosmosDB connection, so now this UI button does nothing. Do you think that showing some notification is a good way to resolve this?

EricJizbaMSFT: I've mentioned multiple times that I don't agree with that decision and I'm still waiting for a response. Copying my questions from [this comment](https://github.com/Microsoft/vscode-azureappservice/pull/629#issuecomment-428685178):

> I can understand "only one connection" if it was just a question of time/prioritization, but you already support multiple so this feels like it's extra work to _remove_ a feature.
> 
> Even if you argue that Cosmos DB will _never_ have multiple connections (which I disagree with) - what about other connection types like storage? Will the the user also never have mutiple storage connections? I want this to set a good pattern for all connection types.

Omelianenko: I agree with you that others type of connections can be multiple, possibly someday even Cosmos can be. That's why I changed the structure of workspace settings. But for now, while we're not having the answer about how to store multiple CosmosDB connections in AppSettings, I have doubts that we still should have this button and possibility to do that.

PR #665, opened 2018-10-20
Using cosmos api object instead of executing cosmos commands

Omelianenko:
Based on changes in [this Cosmos PR](https://github.com/Microsoft/vscode-cosmosdb/pull/886).
Fix #665

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -36,7 +36,6 @@ import { LogPointsSessionWizard } from './logPoints/LogPointsSessionWizard';
 import { RemoteScriptDocumentProvider, RemoteScriptSchema } from './logPoints/remoteScriptDocumentProvider';
 import { LogpointsCollection } from './logPoints/structs/LogpointsCollection';
 import { getPackageInfo, IPackageInfo } from './utils/IPackageInfo';
-import { VscodeCosmos } from './vscode-cosmos.api';
 
 // tslint:disable-next-line:export-name
 // tslint:disable-next-line:max-func-body-length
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -36,14 +36,20 @@ import { LogPointsSessionWizard } from './logPoints/LogPointsSessionWizard';
...
 // tslint:disable-next-line:max-func-body-length
-export function activate(context: vscode.ExtensionContext): void {
+export async function activate(context: vscode.ExtensionContext): Promise<void> {
     registerUIExtensionVariables(ext);
     registerAppServiceExtensionVariables(ext);
     ext.context = context;
 
+    const cosmosExtension = vscode.extensions.getExtension('ms-azuretools.vscode-cosmosdb');
+    if (cosmosExtension) {
+        ext.cosmosAPI = <VscodeCosmos>await cosmosExtension.activate();
```
EricJizbaMSFT: We don't want to get the cosmos extension in the activate method for app service. It will slow down app service and most users don't care about cosmos db.

You need to delay getting the api until you know it's actually necessary. I can discuss offline if you need more detail

Omelianenko: What do you think about doing this in loadMoreChildrenImpl CosmosDBTreeItem?

Omelianenko: Moved it to CosmosDBTreeItem loadMoreChildrenImpl

Omelianenko:
Going to synchronize *.api.d.ts after merge of [this Cosmos PR](https://github.com/Microsoft/vscode-cosmosdb/pull/886).

Omelianenko:
Not relevant now

PR #666, opened 2018-10-20
Add own app setting for docdb sql connections 

### LINKED ISSUE #663: Support SQL database connections
Strongly connected with [this](https://github.com/Microsoft/vscode-cosmosdb/issues/883). Now it works only with Mongo.

Omelianenko:
fix #663

### DIFF
--- a/src/explorer/CosmosDBTreeItem.ts
+++ b/src/explorer/CosmosDBTreeItem.ts
@@ -72,27 +72,39 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         // tslint:disable-next-line:strict-boolean-expressions
         connectionsUnit.cosmosDB = connectionsUnit.cosmosDB || [];
         if (!connectionsUnit.cosmosDB.find((x: string) => x === connectionToAdd)) {
+            const connectionStringValue = (<string>await vscode.commands.executeCommand('cosmosDB.api.getConnectionString', connectionToAdd));
+
+            let allSettingsNames: string;
+            const appSettingsToCreate = new Map<string, string>();
+            if (/^mongo/i.test(connectionStringValue)) {
+                appSettingsToCreate.set('MONGO_URL', connectionStringValue);
+
+                allSettingsNames = '"MONGO_URL"';
+            } else {
+                const parsedConnection = connectionStringValue.match(/^AccountEndpoint=(.*);AccountKey=(.*);Database=(.*)$/);
+                if (parsedConnection === null) {
+                    throw new Error(`Can't parse provided connection string: ${connectionStringValue}. Should match the next pattern: AccountEndpoint=...;AccountKey=...;Database=...`);
+                }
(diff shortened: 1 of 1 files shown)


src/explorer/CosmosDBTreeItem.ts
```diff
@@ -77,8 +77,14 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
             const createdDatabase = new CosmosDBDatabase(this, connectionToAdd);
             showCreatingTreeItem(createdDatabase.label);
 
-            const appSettingToUpdate = "MONGO_URL";
             const connectionStringValue = (<string>await vscode.commands.executeCommand('cosmosDB.api.getConnectionString', connectionToAdd));
+
+            let appSettingToUpdate: string;
+            if (connectionStringValue.match(/^mongodb(\+srv)?:\/\/[^\/]*/) !== null) {
```
EricJizbaMSFT: In general keep regexp's as simple as they can be while still accomplishing the necessary task. To me, `/^mongo/i` is all you need to distinguish between mongo and sql. Also (nit) use `test` instead of `match` which returns `boolean` instead of `RegExpMatchArray | null`. Something like `/^mongo/i.test(connectionStringValue)`

Omelianenko: What is the `i` in `^/mongo/i?`?

EricJizbaMSFT: the letters after the '/' are flags applied to the regexp. 'i' means case-insensitive and I almost always use it

Another common one is 'g' which means 'global' which means include every match even if there are multiple

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -72,21 +72,41 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         // tslint:disable-next-line:strict-boolean-expressions
         connectionsUnit.cosmosDB = connectionsUnit.cosmosDB || [];
         if (!connectionsUnit.cosmosDB.find((x: string) => x === connectionToAdd)) {
+            const connectionStringValue = (<string>await vscode.commands.executeCommand('cosmosDB.api.getConnectionString', connectionToAdd));
+
+            const allSettingsNames: string[] = [];
+            const appSettingsToCreate = new Map<string, string>();
+            if (/^mongo/i.test(connectionStringValue)) {
+                allSettingsNames.push('MONGO_URL');
+                appSettingsToCreate.set(allSettingsNames[0], connectionStringValue);
```
EricJizbaMSFT: I don't want you referencing `allSettingsNames[0]`. Just create a separate variable for mongoUrl and reference that explicitly. We don't know how allSettingsNames could change in the future and I don't want to rely on the ordering of the array. For example, maybe someone adds a default setting to both mongo AND docdb which means `allSettingsNames[0]` would resolve to that default setting instead of 'MONGO_URL'


Mostly I'm worried about a scenario in the future where someone adds a default value to `allSettingsNames` and then

PR #675, opened 2018-10-25
Add warning message for overwriting existing application settings.

### LINKED ISSUE #660: Don't allow to rewrite "MONGO_URL" application setting if already exist
Now if you add new mongo connection it silently rewrites "MONGO_URL" app setting. I think that some notification like ` "MONGO_URL" application setting is going to be rewritten, are you sure you want to do it?Yes|No` could help here.

Omelianenko:


Omelianenko:
Not relevant now

Issue #567, opened 2018-08-16
Add a tree node dedicated to connecting to log streaming


fiveisprime:
Continuing from #519 

In an effort to make log streaming (and enabling app logging) discoverable, let's add a new node either under "Log Files" (and potentially change that to "Logs") or directly under the web app.

![image](https://user-images.githubusercontent.com/1186948/44236242-276aa400-a161-11e8-9046-8406b8831d67.png)

PR #676, opened 2018-10-25
Logstream tree node

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/567
Relies on this PR: https://github.com/Microsoft/vscode-azuretools/pull/296

I'd still like to add icons and possibly change the wording/behavior.  I figured it made the most sense to change it from connect -> disconnect, but please let me know if you disagree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -627,6 +627,11 @@
                     "command": "appService.RemoveCosmosDBConnection",
                     "when": "view == azureAppService && viewItem == cosmosDBDatabase",
                     "group": "2@1"
+                },
+                {
+                    "command": "appService.Refresh",
+                    "when": "view == azureAppService && viewItem == logstream",
+                    "group": "1@1"
                 }
             ],
             "explorer/context": [
--- a/package.json
+++ b/package.json
@@ -627,11 +627,6 @@
                     "command": "appService.RemoveCosmosDBConnection",
                     "when": "view == azureAppService && viewItem == cosmosDBDatabase",
                     "group": "2@1"
-                },
-                {
-                    "command": "appService.Refresh",
-                    "when": "view == azureAppService && viewItem == logstream",
-                    "group": "1@1"
                 }
             ],
             "explorer/context": [
--- a/src/commands/startStreamingLogs.ts
+++ b/src/commands/startStreamingLogs.ts
@@ -5,27 +5,40 @@
 
 import * as vscode from 'vscode';
 import * as appservice from 'vscode-azureappservice';
+import { LogStreamTreeItem } from '../explorer/LogStreamTreeItem';
 import { SiteTreeItem } from "../explorer/SiteTreeItem";
 import { WebAppTreeItem } from '../explorer/WebAppTreeItem';
 import { ext } from '../extensionVariables';
 import { enableFileLogging } from './enableFileLogging';
 
-export async function startStreamingLogs(node?: SiteTreeItem): Promise<void> {
+export async function startStreamingLogs(node?: SiteTreeItem | LogStreamTreeItem): Promise<void> {
+    let logStreamTreeItem: LogStreamTreeItem | undefined;
+    let webAppTreeItem: WebAppTreeItem;
+
     if (!node) {
-        node = <WebAppTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+        webAppTreeItem = <WebAppTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+    } else if (node.contextValue === LogStreamTreeItem.contextValue) {
+        logStreamTreeItem = <LogStreamTreeItem>node;
+        // tslint:disable-next-line:no-non-null-assertion
+        webAppTreeItem = <WebAppTreeItem>node.parent!.parent;
+    } else {
+        webAppTreeItem = <WebAppTreeItem>node;
     }
 
     const verifyLoggingEnabled: () => Promise<void> = async (): Promise<void> => {
         const isEnabled = await vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, async p => {
             p.report({ message: 'Checking container diagnostics settings...' });
-            // tslint:disable-next-line:no-non-null-assertion
-            return await node!.isHttpLogsEnabled();
+            return await webAppTreeItem.isHttpLogsEnabled();
         });
         if (!isEnabled) {
-            // tslint:disable-next-line:no-non-null-assertion
-            await enableFileLogging(node!);
+            await enableFileLogging(webAppTreeItem);
         }
     };
 
-    await appservice.startStreamingLogs(node.root.client, verifyLoggingEnabled, node.logStreamLabel);
+    await appservice.startStreamingLogs(webAppTreeItem.root.client, verifyLoggingEnabled, webAppTreeItem.logStreamLabel);
+    if (logStreamTreeItem) {
+        // don't wait for this refresh
+        // tslint:disable-next-line:no-floating-promises
+        logStreamTreeItem.refresh();
+    }
 }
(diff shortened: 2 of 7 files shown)


src/extension.ts
```diff
@@ -248,6 +249,11 @@ export function activate(context: vscode.ExtensionContext): void {
         }
 
         await stopStreamingLogs(node.root.client);
+        if (node.contextValue === LogStreamTreeItem.contextValue) {
+            // don't wait for this refresh
+            // tslint:disable-next-line:no-floating-promises
+            node.refresh();
```
EricJizbaMSFT: Why don't wait?

src/commands/startStreamingLogs.ts
```diff
@@ -5,27 +5,40 @@
...
         });
         if (!isEnabled) {
-            // tslint:disable-next-line:no-non-null-assertion
-            await enableFileLogging(node!);
+            await enableFileLogging(webAppTreeItem);
         }
     };
 
-    await appservice.startStreamingLogs(node.root.client, verifyLoggingEnabled, node.logStreamLabel);
+    await appservice.startStreamingLogs(webAppTreeItem.root.client, verifyLoggingEnabled, webAppTreeItem.logStreamLabel);
```
EricJizbaMSFT: startStreamingLogs already returns the `ILogStream`. I'm pretty sure you can use that rather than searching for the log stream using your new `getLogStreams` thing

src/commands/startStreamingLogs.ts
```diff
@@ -5,27 +5,40 @@
...
 
-export async function startStreamingLogs(node?: SiteTreeItem): Promise<void> {
+export async function startStreamingLogs(node?: SiteTreeItem | LogStreamTreeItem): Promise<void> {
+    let logStreamTreeItem: LogStreamTreeItem | undefined;
+    let webAppTreeItem: WebAppTreeItem;
+
     if (!node) {
-        node = <WebAppTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+        webAppTreeItem = <WebAppTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+    } else if (node.contextValue === LogStreamTreeItem.contextValue) {
```
EricJizbaMSFT: You should be using `node instanceof LogStreamTreeItem` to benefit from typing

src/commands/startStreamingLogs.ts
```diff
@@ -5,27 +5,40 @@
...
+        logStreamTreeItem = <LogStreamTreeItem>node;
+        // tslint:disable-next-line:no-non-null-assertion
+        webAppTreeItem = <WebAppTreeItem>node.parent!.parent;
```
EricJizbaMSFT: If you define the type of the parent like this: https://github.com/Microsoft/vscode-azureappservice/blob/master/src/explorer/CosmosDBTreeItem.ts#L19

Than you won't have to do the non-null-assertion

src/commands/startStreamingLogs.ts
```diff
@@ -5,27 +5,40 @@
...
+    } else {
+        webAppTreeItem = <WebAppTreeItem>node;
```
EricJizbaMSFT: This could be a slot. Casting to WebAppTreeItem isn't accurate

package.json
```diff
@@ -627,6 +627,11 @@
                     "command": "appService.RemoveCosmosDBConnection",
                     "when": "view == azureAppService && viewItem == cosmosDBDatabase",
                     "group": "2@1"
+                },
+                {
+                    "command": "appService.Refresh",
```
EricJizbaMSFT: Do we really need this refresh?

src/extension.ts
nturinski: I figured we wouldn't want to hang up in case the refresh took a while.  Figured there was nothing to gain from awaiting it.

src/explorer/LogStreamTreeItem.ts
```diff
@@ -0,0 +1,35 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { getLogStreamId, getLogStreams, ILogStream, ISiteTreeRoot } from 'vscode-azureappservice';
+import { AzureParentTreeItem, AzureTreeItem } from 'vscode-azureextensionui';
+
+const openLogStreamCommand: string = 'appService.OpenLogStream';
+const stopLogStreamCommand: string = 'appService.StopLogStream';
+const disconnectLogStream: string = 'Disconnect log stream...';
```
EricJizbaMSFT: I was talking to Matt and he thought it would be more useful in the "isConnected" case if this node opened the output channel. Sometimes it's hard to find.

Honestly "Connect to log stream..." and "View log stream..." are so similar you might not even need to change the label if that simplifies the code

src/commands/startStreamingLogs.ts
nturinski: I'm having some issues with this.. :|

nturinski:
I'm running into an issue where the Log Streams node isn't the first child despite being the first element in the array.  Does VS Code's tree provider automatically sort children alphabetically or something?

EricJizbaMSFT:
Yes it's sorted by default but you can change that: https://github.com/Microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L250

EricJizbaMSFT:
To help with all the edge cases, I might suggest using an Event pattern on ILogStream. Similar to what you did here:
https://github.com/Microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L23

I imagine there would be an `onDisconnect` and `onConnect` event and you would just hook them both up to refresh the LogStreamTreeItem.

nturinski:
Yeah, I agree that the value seems very low for the amount of work needed to make sure that it handles all test cases.  I just had done some of the work for "Disconnect" and wanted to keep it... 😭

It's probably not worth waiting on an event for it.

src/commands/startStreamingLogs.ts
```diff
@@ -5,27 +5,32 @@
...
+export async function startStreamingLogs(node?: SiteTreeItem | LogStreamTreeItem): Promise<void> {
+    let siteTreeItem: SiteTreeItem;
+
     if (!node) {
-        node = <WebAppTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+        siteTreeItem = <WebAppTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+    } else if (node instanceof LogStreamTreeItem) {
+        siteTreeItem = <SiteTreeItem>node.parent.parent;
+    } else {
+        siteTreeItem = <SiteTreeItem>node;
```
EricJizbaMSFT: nit: I don't think this cast is necessary. Typescript should know based on the other cases that this is SiteTreeItem

src/explorer/LogStreamTreeItem.ts
```diff
@@ -0,0 +1,23 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ISiteTreeRoot } from 'vscode-azureappservice';
+import { AzureTreeItem } from 'vscode-azureextensionui';
+import { FolderTreeItem } from './FolderTreeItem';
+
+const openLogStreamCommand: string = 'appService.OpenLogStream';
+const connectLogStream: string = 'Connect to log stream...';
```
EricJizbaMSFT: nit: doesn't seem necessary to have a constant when you could just do this below:
```
public commandId: string = 'appService.OpenLogStream';
```

src/extension.ts
```diff
@@ -248,6 +249,9 @@ export function activate(context: vscode.ExtensionContext): void {
         }
 
         await stopStreamingLogs(node.root.client);
+        if (node instanceof LogStreamTreeItem) {
```
EricJizbaMSFT: Is this refresh still necessary?

PR #681, opened 2018-10-26
Mongo connections based on application settings. Using CosmosDB API object

### LINKED ISSUE #680: Now load connections from AppSettings
As we discussed, going to get rid of the workspace appService.connections variable. For now, going to support only settings that store MongoDB connection strings

Omelianenko:
Fix #680 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -846,7 +846,7 @@
     },
     "dependencies": {
         "azure-arm-resource": "^3.0.0-preview",
-        "azure-arm-website": "^4.0.0",
+        "azure-arm-website": "^5.3.0",
         "fs-extra": "^4.0.2",
         "opn": "^5.3.0",
         "portfinder": "^1.0.13",
--- a/package.json
+++ b/package.json
@@ -624,12 +624,12 @@
                 },
                 {
                     "command": "appService.RevealConnection",
-                    "when": "view == azureAppService && viewItem == cosmosDBDatabase",
+                    "when": "view == azureAppService && viewItem == cosmosDBConnection",
                     "group": "1@1"
                 },
                 {
                     "command": "appService.RemoveCosmosDBConnection",
-                    "when": "view == azureAppService && viewItem == cosmosDBDatabase",
+                    "when": "view == azureAppService && viewItem == cosmosDBConnection",
                     "group": "2@1"
                 }
             ],
--- a/package.json
+++ b/package.json
@@ -852,7 +852,7 @@
         "portfinder": "^1.0.13",
         "request": "^2.83.0",
         "request-promise": "^4.2.2",
-        "vscode-azureappservice": "^0.23.0",
+        "vscode-azureappservice": "^0.24.1",
         "vscode-azureextensionui": "^0.18.1",
         "vscode-azurekudu": "^0.1.8",
         "vscode-debugadapter": "^1.24.0",
--- a/src/explorer/CosmosDBDatabase.ts
+++ b/src/explorer/CosmosDBDatabase.ts
@@ -7,6 +7,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { ISiteTreeRoot } from 'vscode-azureappservice';
 import { AzureTreeItem } from 'vscode-azureextensionui';
+import { CosmosDBItem } from '../vscode-cosmos.api';
 import { CosmosDBTreeItem } from './CosmosDBTreeItem';
 
 export class CosmosDBDatabase extends AzureTreeItem<ISiteTreeRoot> {
@@ -18,67 +19,20 @@ export class CosmosDBDatabase extends AzureTreeItem<ISiteTreeRoot> {
         };
     }
     public static contextValue: string = 'cosmosDBDatabase';
-
-    private static mongoPref: string = '[a-zA-Z]+:\/\/';
-    private static mongoCredential: string = '(?:[^@]*@)?';
-    private static mongoAccount: string = '([^\/]*)';
-    private static mongoDatabase: string = '(\/[^\?]*)?';
-
     public readonly contextValue: string = CosmosDBDatabase.contextValue;
     public readonly label: string;
     public readonly parent: CosmosDBTreeItem;
 
-    constructor(parent: CosmosDBTreeItem, readonly connectionId: string) {
+    constructor(parent: CosmosDBTreeItem, readonly cosmosDBItem: CosmosDBItem, readonly appSettingName: string) {
         super(parent);
-        this.label = this.getLabel(connectionId);
+        if (cosmosDBItem.accountName && cosmosDBItem.databaseName) {
+            this.label = cosmosDBItem.accountName + String("/") + cosmosDBItem.databaseName;
+        } else {
+            throw new Error("Couldn't get correct database from CosmosDB.");
+        }
     }
 
     public async deleteTreeItemImpl(): Promise<void> {
         throw new Error('Not implemented yet!');
-    }
-
-    private getLabel(id: string): string {
-        const items = this.mongoGetAccountDatabase(id);
-        if (items) {
-            const accountName = items[1];
-            if (accountName.includes(',')) {
-                throw new Error(`Don't suppoort mongo connections with multiple hosts yet.`);
-            }
-            let label = this.accountNameShorter(accountName);
-
-            if (items.length === 3 && items[2].length > 1) {
-                const databaseName = items[2];
-                label = label.concat(databaseName);
-            }
-
-            return label;
-        }
-        throw new Error('Failed to parse connection id');
-    }
-
-    /**
-     * Mongo connection string that supported follows the following format:
-     * mongodb://[username:password@]account.documents.azure.com:10255[/[database][?ssl=true]]
-     */
-    private mongoGetAccountDatabase(id: string): RegExpMatchArray | undefined {
-        const regExp = new RegExp(CosmosDBDatabase.mongoPref + CosmosDBDatabase.mongoCredential + CosmosDBDatabase.mongoAccount + CosmosDBDatabase.mongoDatabase);
-        const matches: RegExpMatchArray | null = id.match(regExp);
-        if (matches === null || matches.length <= 1) {
-            return undefined;
-        }
-        return matches;
-    }
-
-    /**
-     * Make account name shorter: account.documents.azure.com:10255 -> account
-     * For now assume that only `.` can split
-     */
-    private accountNameShorter(accountName: string): string {
-        const regExp = /[^.]*/;
-        const matches = accountName.match(regExp);
-        if (matches) {
-            return matches[0];
-        }
-        return accountName;
     }
 }
--- a/src/explorer/CosmosDBDatabase.ts
+++ b/src/explorer/CosmosDBDatabase.ts
@@ -11,13 +11,6 @@ import { CosmosDBItem } from '../vscode-cosmos.api';
 import { CosmosDBTreeItem } from './CosmosDBTreeItem';
 
 export class CosmosDBDatabase extends AzureTreeItem<ISiteTreeRoot> {
-
-    public get iconPath(): string | vscode.Uri | { light: string | vscode.Uri; dark: string | vscode.Uri } {
-        return {
-            light: path.join(__filename, '..', '..', '..', '..', 'resources', 'light', 'Database.svg'),
-            dark: path.join(__filename, '..', '..', '..', '..', 'resources', 'dark', 'Database.svg')
-        };
-    }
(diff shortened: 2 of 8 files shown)


src/explorer/CosmosDBDatabase.ts
```diff
@@ -22,48 +33,52 @@ export class CosmosDBDatabase extends AzureTreeItem<ISiteTreeRoot> {
...
-            }
-        }
+        throw new Error('Not implemented yet!');
     }
 
     private getLabel(id: string): string {
-        const items = this.parseCosmos(id) || this.parseAttached(id);
-        if (!items) {
-            throw new Error('Failed to parse connection id');
+        const items = this.mongoGetAccountDatabase(id);
```
EricJizbaMSFT: The Cosmos DB extension already has logic to create a label from the connection string. See `getServerIdFromConnectionString` here:
https://github.com/Microsoft/vscode-cosmosdb/blob/master/src/tree/AttachedAccountsTreeItem.ts#L211

It's completely different than what you're doing and I don't want to duplicate that logic, so this should probably be included in the API. Maybe you pass in the connection string and it just returns the account name and database name

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -38,20 +38,38 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
         }
-        const workspaceConfig = vscode.workspace.getConfiguration(constants.extensionPrefix);
-        const connections = workspaceConfig.get<IConnections[]>(constants.configurationSettings.connections, []);
-        // tslint:disable-next-line:strict-boolean-expressions
-        const unit = connections.find((x: IConnections) => x.webAppId === this.root.client.id) || <IConnections>{};
-        if (!unit.cosmosDB || unit.cosmosDB.length === 0) {
+
+        const connectionStrings: string[] = [];
+        const appSettingsNode = this.parent.parent.appSettingsNode;
+        const allAppSettings = await appSettingsNode.ensureSettings();
```
EricJizbaMSFT: No need to use the appSettingsNode directly. Just call `this.root.client.listApplicationSettings()`

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -38,20 +38,38 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+        if (allAppSettings.properties) {
+            const dictionary: { [key: string]: string } = allAppSettings.properties;
+            Object.keys(dictionary).forEach((key) => {
+                connectionStrings.push(dictionary[key]);
+            });
+        }
+
+        const allConnectionItems = connectionStrings.map(connectionId => {
+            try {
```
EricJizbaMSFT: Avoid using a try/catch where you ignore errors. You might accidentally ignore the wrong errors. I suggest something like this:

```typescript
const treeItems: CosmosDBDatabase[] = [];
for (const connectionString of connectionStrings) {
    if (<connectionString is a mongo connection string>) {
        treeItems.push(new CosmosDBDatabase(...));
    }
}
...
return treeItems;
```

src/explorer/CosmosDBDatabase.ts
Omelianenko: I think it's better to use something like parent.label(in Cosmos API method). As I understand, for attached this will give us the same label as `getServerIdFromConnectionString `(the label is already created by that function), but for "internal" we'll get account labels that are actually in CosmosDB explorer. 
Oh, will I have those labels for attached databases if I didn't "loadmore"  "Attached Database Accounts" node? 

Wanted to parse connection string because wasn't sure that it's a good idea to make API calls for all connections when loadmore method is running. So, if we do not care a lot about API calls can we just get Cosmos treeItemId from getDatabase() method(when the loadmore method is running)? In this case, Reveal command will stay the same and we don't need to overcomplicate it.

src/vscode-cosmos.api.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    cosmosDBTreeItemId: string | undefined
+    databaseName: string | undefined
+}
+
+export interface VSCodeCosmosDB {
+    /**
+     * Founds the database in CosmosDB and return info about it including TreeItem id in CosmosDB Explorer
+     * @param connectionData The connection string and/or other info that provided for successfull database searching
+     */
+    getDatabase(connectionData: CosmosDBItem): Promise<CosmosDBItem | undefined>;
```
EricJizbaMSFT: connectionData should not be of type `CosmosDBItem`. It should be something like `{ connectionString: string }` since that's the only thing you support right now.

src/vscode-cosmos.api.d.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface CosmosDBItem {
+    accountName: string | undefined
+    connectionString: string | undefined
```
EricJizbaMSFT: None of these should be `| undefined`. getDatabase and pickDatabase should always return an item with these properties defined.

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -38,90 +44,82 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
-        const workspaceConfig = vscode.workspace.getConfiguration(constants.extensionPrefix);
-        const allConnections = workspaceConfig.get<IConnections[]>(constants.configurationSettings.connections, []);
-        let connectionsUnit = allConnections.find((x: IConnections) => x.webAppId === this.root.client.id);
-        if (!connectionsUnit) {
-            connectionsUnit = <IConnections>{};
-            allConnections.push(connectionsUnit);
-            connectionsUnit.webAppId = this.root.client.id;
-        }
+        const appSettingsNode = this.parent.parent.appSettingsNode;
+        const appSettingKeyToAdd = (<AppSettingTreeItem>await appSettingsNode.createChild(databaseToAdd.connectionString)).id;
```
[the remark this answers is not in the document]
Omelianenko: Should I copy-paste appSettingsNode.createChild & validateNewKeyInput methods in AppService then? They are almost exactly what I need, can't find other way to run those

src/vscode-cosmos.api.d.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface CosmosDBItem {
```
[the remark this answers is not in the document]
Omelianenko: I already have CosmosDBDatabase in AppService as the class for CosmosDB UI node

src/explorer/CosmosDBTreeItem.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yes you should implement your own prompt for the key. You should have a different message and default value anyways. I'm fine if you use `validateNewKeyInput` directly, but I would make it a static method on the class or just a function from that file.

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -38,87 +38,70 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+            ext.cosmosAPI = <VSCodeCosmosDB>cosmosDB.exports;
         }
-        return unit.cosmosDB.map(connectionId => {
-            return new CosmosDBDatabase(this, connectionId);
+
+        const mongoAppSettingsKeys: string[] = [];
+        // tslint:disable-next-line:strict-boolean-expressions
+        const appSettings = (await this.root.client.listApplicationSettings()).properties || {};
+        Object.keys(appSettings).forEach((key) => {
+            if (/^mongo/i.test(appSettings[key])) {
```
EricJizbaMSFT: Mmm maybe not quite this simple. I think you should use the regexp you previously had for just the prefix: `^mongodb[^\/]*:\/\/`

Although `^\/` seems like it should be `^:` to me since ':' comes before '/' in the connection string.

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -38,87 +38,70 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+            label: 'Add Cosmos DB Connection...'
+        })];
+    }
 
-            const appSettingsNode = this.parent.parent.appSettingsNode;
-            appSettingsToCreate.forEach(async (value, appSetting) => appSettingsNode.editSettingItem(appSetting, appSetting, value));
-            await appSettingsNode.refresh();
+    public async createChildImpl(showCreatingTreeItem: (label: string) => void): Promise<AzureTreeItem<ISiteTreeRoot>> {
+        const databaseToAdd = await ext.cosmosAPI.pickDatabase();
+        if (!databaseToAdd || !databaseToAdd.connectionString) {
```
EricJizbaMSFT: No need to do `!databaseToAdd.connectionString`

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -38,87 +38,70 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+            throw new UserCancelledError();
+        }
+        // tslint:disable-next-line:strict-boolean-expressions
+        const appSettings = await this.root.client.listApplicationSettings() || {};
+        const appSettingKeyToAdd: string = await ext.ui.showInputBox({
+            prompt: 'Enter new connection setting key',
```
EricJizbaMSFT: Make the default value 'MONGO_URL'

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -38,87 +38,70 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
-            const appSettingsNode = this.parent.parent.appSettingsNode;
-            appSettingsToCreate.forEach(async (value, appSetting) => appSettingsNode.editSettingItem(appSetting, appSetting, value));
-            await appSettingsNode.refresh();
+    public async createChildImpl(showCreatingTreeItem: (label: string) => void): Promise<AzureTreeItem<ISiteTreeRoot>> {
+        const databaseToAdd = await ext.cosmosAPI.pickDatabase();
+        if (!databaseToAdd || !databaseToAdd.connectionString) {
+            throw new UserCancelledError();
+        }
+        // tslint:disable-next-line:strict-boolean-expressions
+        const appSettings = await this.root.client.listApplicationSettings() || {};
```
EricJizbaMSFT: Do you need `|| {}`? can this really return undefined?

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -38,87 +38,70 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
-            // Don't wait
-            vscode.window.showInformationMessage(`Database "${createdDatabase.label}" connected to Web App "${this.root.client.fullName}". Created ${allSettingsNames.map((s) => `"${s}"`).join(', ')} App Settings.`, ok, showDatabase).then(async (result: vscode.MessageItem | undefined) => {
-                if (result === showDatabase) {
-                    vscode.commands.executeCommand('appService.RevealConnection', createdDatabase);
-                }
-            });
+        const ok: vscode.MessageItem = { title: 'OK' };
+        const showDatabase: vscode.MessageItem = { title: 'Show Database' };
+        // Don't wait
+        vscode.window.showInformationMessage(`Database "${createdDatabase.label}" connected to Web App "${this.root.client.fullName}". Created "${appSettingKeyToAdd}" App Settings.`, ok, showDatabase).then(async (result: vscode.MessageItem | undefined) => {
```
EricJizbaMSFT: nit: "Web App" and "App Settings" should all be lower case.

package.json
```diff
@@ -852,7 +852,7 @@
         "portfinder": "^1.0.13",
         "request": "^2.83.0",
         "request-promise": "^4.2.2",
-        "vscode-azureappservice": "^0.23.0",
+        "vscode-azureappservice": "^0.24.1",
```
EricJizbaMSFT: You need to also bump 'azure-arm-website' to '5.whatever'. We bumped it in the 0.24.0 release. See here: https://github.com/Microsoft/vscode-azuretools/blob/master/appservice/package.json#L37

PR #685, opened 2018-10-30
Remove connection method

### LINKED ISSUE #660: Don't allow to rewrite "MONGO_URL" application setting if already exist

### LINKED ISSUE #662: To clean or even to delete application settings after connection deleting
When you delete the connection, "MONGO_URL" application setting stays the same, so in fact, it's still `connected`.
image
image

### LINKED ISSUE #664: Use CosmosDB API object instead of executing commands

Omelianenko:
Fix #662 
Fix #660 fixed on the side of CosmosDB, it asks the appsettings key name for connection creating. ([link](https://github.com/Microsoft/vscode-cosmosdb/pull/886))
Fix #664 by this and [the last](https://github.com/Microsoft/vscode-azureappservice/pull/681) PRs

### DIFF
--- a/src/explorer/CosmosDBConnection.ts
+++ b/src/explorer/CosmosDBConnection.ts
@@ -53,7 +53,6 @@ export class CosmosDBConnection extends AzureTreeItem<ISiteTreeRoot> {
             keysToDelete.forEach((key) => {
                 delete properties[key];
             });
-            appSettings.properties = properties;
             await this.root.client.updateApplicationSettings(appSettings);
             await this.parent.parent.parent.appSettingsNode.refresh();
         }
--- a/src/explorer/CosmosDBConnection.ts
+++ b/src/explorer/CosmosDBConnection.ts
@@ -33,7 +33,6 @@ export class CosmosDBConnection extends AzureTreeItem<ISiteTreeRoot> {
         const valueToDelete = this.cosmosDBDatabase.connectionString;
 
         const appSettings = await this.root.client.listApplicationSettings();
-        // tslint:disable-next-line:strict-boolean-expressions
         const properties = appSettings.properties;
         if (properties) {
             const keysToDelete: string[] = [];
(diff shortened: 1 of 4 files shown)


src/explorer/CosmosDBConnection.ts
```diff
@@ -29,6 +30,34 @@ export class CosmosDBConnection extends AzureTreeItem<ISiteTreeRoot> {
     }
 
     public async deleteTreeItemImpl(): Promise<void> {
-        throw new Error('Not implemented yet!');
+        const valueToDelete = this.cosmosDBDatabase.connectionString;
+
+        const appSettings = await this.root.client.listApplicationSettings();
+        // tslint:disable-next-line:strict-boolean-expressions
+        const properties = appSettings.properties || {};
```
Omelianenko: Simple "if" is better here

src/explorer/CosmosDBConnection.ts
```diff
@@ -29,6 +30,32 @@ export class CosmosDBConnection extends AzureTreeItem<ISiteTreeRoot> {
...
+                const items: vscode.MessageItem[] = [DialogResponses.deleteResponse, DialogResponses.cancel];
+                const result: vscode.MessageItem = await ext.ui.showWarningMessage(warning, { modal: true }, ...items);
+                if (result === DialogResponses.cancel) {
+                    throw new UserCancelledError();
+                }
+            }
+            keysToDelete.forEach((key) => {
+                delete properties[key];
+            });
+            appSettings.properties = properties;
```
EricJizbaMSFT: nit: I don't think you have to do `appSettings.properties = properties;` since I think `properties` is a reference (aka not a copy)

PR #689, opened 2018-11-02
Don't try to show the connections with the same labels. Changed lint setting for static methods

### LINKED ISSUE #688: The error occurs when there are more than one AppSetting with the same value

Omelianenko:
Fix #688


src/explorer/CosmosDBTreeItem.ts
```diff
@@ -44,11 +44,13 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         }
 
         const mongoAppSettingsKeys: string[] = [];
+        const usedValue: { [key: string]: boolean } = {};
         // tslint:disable-next-line:strict-boolean-expressions
         const appSettings = (await this.root.client.listApplicationSettings()).properties || {};
         Object.keys(appSettings).forEach((key) => {
-            if (/^mongodb[^:]*:\/\//i.test(appSettings[key])) {
+            if (/^mongodb[^:]*:\/\//i.test(appSettings[key]) && !usedValue[appSettings[key]]) {
```
EricJizbaMSFT: I wouldn't base this logic on the connectionString. You can have different connection strings point to the same database (for example, if they use different keys). I would remove duplicates _after_ you get `const cosmosDBDatabase = await ext.cosmosAPI.getDatabase({ connectionString: appSettings[key] });` and check based on hostName/port/databaseName

Omelianenko: The problem occurs becuase it's impossible to show treeitems with the same labels(under Connections node). I can fix it just by deleting duplicates from `treeItems ` based on label. Also, it's possible to have connections with the same account and database name in other azure subscription, but I don't think we should care about it a lot as it's very rare and corner case

EricJizbaMSFT: > I can fix it just by deleting duplicates from treeItems based on labels

Yes I would prefer that

Omelianenko: > > I can fix it just by deleting duplicates from treeItems based on labels
> 
> Yes I would prefer that

It didn't help, the error occurs when it creates new CosmosDBConnection with existed id 😞

src/explorer/CosmosDBConnection.ts
```diff
@@ -18,7 +18,12 @@ export class CosmosDBConnection extends AzureTreeItem<ISiteTreeRoot> {
 
     constructor(parent: CosmosDBTreeItem, readonly cosmosDBDatabase: CosmosDBDatabase, readonly appSettingName: string) {
         super(parent);
-        this.label = `${cosmosDBDatabase.accountName}/${cosmosDBDatabase.databaseName}`;
+        this.label = CosmosDBConnection.makeLabel(cosmosDBDatabase);
+    }
+
+    // tslint:disable-next-line:function-name
```
EricJizbaMSFT: nit: Rather than disable the rule for this line, can you disable it for this repo? Just change tslint.json to have the same "function-name" stuff as this: 
https://github.com/Microsoft/vscode-azureeventgrid/blob/master/tslint.json#L20

PR #692, opened 2018-11-02
Reveal in App Settings button

### LINKED ISSUE #691: Make dependence between application settings and database connections more obvious to users
Now it's not very clear that automatically created database connections nodes were created based on application settings.

I prefer to add reveal in application settings button:
image
image

Omelianenko:
Fix #691


package.json
```diff
@@ -628,9 +637,14 @@
                     "group": "1@1"
                 },
                 {
-                    "command": "appService.RemoveCosmosDBConnection",
+                    "command": "appService.RevealConnectionInAppSettings",
                     "when": "view == azureAppService && viewItem == cosmosDBConnection",
                     "group": "2@1"
```
EricJizbaMSFT: I feel like both reveal commands should be in the same "group". Reveal in Cosmos should be "1@1" and reveal in app settings should be "1@2". Please comment with a screenshot after you're done

src/commands/connections/revealConnectionInAppSettings.ts
```diff
@@ -0,0 +1,14 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { CosmosDBConnection } from '../../explorer/CosmosDBConnection';
+import { ext } from "../../extensionVariables";
+
+export async function revealConnectionInAppSettings(node: CosmosDBConnection): Promise<void> {
+    const nodeToReveal = await ext.tree.findTreeItem(`${node.parent.parent.parent.appSettingsNode.fullId}/${node.appSettingKey}`);
+    if (nodeToReveal) {
```
EricJizbaMSFT: This should throw an error if it can't find the node, rather than silently failing. Something like `Failed to find app setting with key "...".`

src/extensionVariables.ts
```diff
@@ -17,4 +17,5 @@ export namespace ext {
     export let reporter: ITelemetryReporter;
     export let context: ExtensionContext;
     export let cosmosAPI: VSCodeCosmosDB;
+    export let treeView: TreeView<AzureTreeItem<ISubscriptionRoot>>;
```
EricJizbaMSFT: nit: `TreeView<AzureTreeItem>` should work

package.json
```diff
@@ -628,9 +637,14 @@
...
+                },
+                {
+                    "command": "appService.RemoveCosmosDBConnection",
```
EricJizbaMSFT: Adding 'remove' is not related to "Reveal in App Settings button". Please do a separate PR for that.

Omelianenko: It's already [in master](https://github.com/Microsoft/vscode-azureappservice/blob/3485c7b996f951969ccae5c3db5fac16dcffe83d/package.json#L630-L634). Github just shows changes not in the best way

Omelianenko:
As we discussed, going to add a button for database revealing before merging(the same as filter button for subscriptions).

EricJizbaMSFT:
Please merge ASAP. Any new features (no matter how minor) should be done separately so that it doesn't slow down or delay the existing stuff that's already approved

PR #695, opened 2018-11-09
Uses new Cosmos API

Omelianenko:
PR for the [version](https://github.com/Microsoft/vscode-cosmosdb/pull/886) of API


EricJizbaMSFT:
Please fix the Conflicting files before I review

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -55,13 +55,24 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
         const usedLabels: { [key: string]: boolean } = {};
         for (const key of mongoAppSettingsKeys) {
-            const cosmosDBDatabase = await ext.cosmosAPI.getDatabase({ connectionString: appSettings[key] });
-            if (cosmosDBDatabase) {
-                const label = CosmosDBConnection.makeLabel(cosmosDBDatabase);
-                if (!usedLabels[label]) {
-                    usedLabels[label] = true;
-                    treeItems.push(new CosmosDBConnection(this, cosmosDBDatabase, key));
+            try {
+                const databaseTreeItem = await ext.cosmosAPI.findTreeItem({ connectionString: appSettings[key] });
```
EricJizbaMSFT: Please use `createTreeItemsWithErrorHandling`, which handles errors in a common way

Here's an example: https://github.com/Microsoft/vscode-azureappservice/blob/master/src/explorer/WebAppProvider.ts#L48

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -77,7 +88,11 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
     }
 
     public async createChildImpl(showCreatingTreeItem: (label: string) => void): Promise<AzureTreeItem<ISiteTreeRoot>> {
-        const databaseToAdd = await ext.cosmosAPI.pickDatabase();
+        const options: PickTreeItemOptions & { resourceType: 'Database' } = {
```
EricJizbaMSFT: nit: You should be able to just do: `const options: PickTreeItemOptions = {`

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -40,7 +40,7 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         }
 
         if (ext.cosmosAPI === undefined) {
-            ext.cosmosAPI = <VSCodeCosmosDB>cosmosDB.exports;
+            ext.cosmosAPI = <CosmosDBExtensionApi>cosmosDB.exports;
```
EricJizbaMSFT: You're going to have to change this once you merge the cosmos db pr with master. It will return an [AzureExtensionApiProvider](https://github.com/Microsoft/vscode-azuretools/blob/master/ui/api.d.ts#L15) which you'll have to call `getApi` on and pass in `^1.0.0`

src/explorer/CosmosDBTreeItem.ts
Omelianenko: Compile error I get:
image

PR #699, opened 2018-11-14
Add support for account-level connection strings

EricJizbaMSFT:
I think database-level connection strings will more common, but you never know what users will put in their application settings. The only real difference is the label we show.

Also delete an unused file.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -321,7 +321,7 @@ export function activate(context: vscode.ExtensionContext): void {
     });
     registerCommand('appService.AddCosmosDBConnection', addCosmosDBConnection);
     registerCommand('appService.RemoveCosmosDBConnection', removeCosmosDBConnection);
-    registerCommand('appService.RevealConnection', async (node: CosmosDBConnection) => await node.databaseTreeItem.reveal());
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -321,7 +321,7 @@ export function activate(context: vscode.ExtensionContext): void {
     });
     registerCommand('appService.AddCosmosDBConnection', addCosmosDBConnection);
     registerCommand('appService.RemoveCosmosDBConnection', removeCosmosDBConnection);
-    registerCommand('appService.RevealConnection', async (node: CosmosDBConnection) => await node.databaseTreeItem.reveal());
+    registerCommand('appService.RevealConnection', async (node: CosmosDBConnection) => await node.cosmosExtensionItem.reveal());
```
nturinski: Excuse me, sir, we're trying to transition our `extension.ts` file to only contain the commands.

EricJizbaMSFT:
> how do users use an account to attach to a database

Idk but users can put anything in their app settings. It's possible they just forgot the database name

src/extension.ts
EricJizbaMSFT: I didn't add any new commands - just edited existing ones :P

Issue #610, opened 2018-09-11
[Suggestion]The description of setting 'appService.defaultWebAppToDeploy' is not clear in Workspace Settings


v-dantch:
Repro Steps:
1. Open Workspace Settings, type and select 'appService.defailtWebAppToDeploy'.
2. Check the description of this setting.

Actual:
Display 'The default web app to use when deploying. Every subequent deployment will deploy to this web app or slot'.
![actual](https://user-images.githubusercontent.com/34729092/45333681-fcf4e680-b566-11e8-948d-442bd8dd7401.PNG)

Suggestion:
The description is clearer and told the web app's id is required such as : 'The default web app's id to use when deploying...'.

PR #706, opened 2018-11-15
Reword description of defaultWebAppToDeploy to be clearer

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/610

### DIFF
--- a/package.json
+++ b/package.json
@@ -786,7 +786,7 @@
                 "appService.defaultWebAppToDeploy": {
                     "scope": "resource",
                     "type": "string",
-                    "description": "The default web app to use when deploying.  Every subsequent deployment of this project will deploy to this web app or slot.  The web app is represented with its full Azure id string.  Can be disabled by setting to \"None\""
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -786,7 +786,7 @@
                 "appService.defaultWebAppToDeploy": {
                     "scope": "resource",
                     "type": "string",
-                    "description": "The default web app to use when deploying.  Every subsequent deployment will deploy to this web app or slot."
+                    "description": "The default web app to use when deploying.  Every subsequent deployment of this project will deploy to this web app or slot.  The web app is represented with its full Azure id string.  Can be disabled by setting to \"None\""
```
EricJizbaMSFT: Can we combine `The default web app to use when deploying.` and `The web app is represented with its full Azure id string.`

Something like `The full Azure id of the default web app to use when deploying.`

EricJizbaMSFT: nit: 'workspace' instead of 'project'

Issue #708, opened 2018-11-15
Fix deleting docdb connections


EricJizbaMSFT:
It involves deleting three app settings instead of one and I didn't feel like implementing that as a part of the main 'docdb' PR

EricJizbaMSFT:
Fixed in #709

PR #709, opened 2018-11-15
Add support for DocDB connections

EricJizbaMSFT:
Currently it's pretty strict in terms of how it "detects" a DocDB connection. It has to find an app setting that ends in "_ENDPOINT" and an app setting that ends in "_MASTER_KEY" and they have to have the same prefix. I figure that's a good starting point and we can improve/change the detection in the future.

Relies on this PR in Cosmos: https://github.com/Microsoft/vscode-cosmosdb/pull/946

Also removed the feature flag since this was the biggest piece missing.

Only thing I didn't implement was "Delete connection", but I filed this issue to do it separately: https://github.com/Microsoft/vscode-azureappservice/issues/708

### DIFF
--- a/src/commands/connections/revealConnectionInAppSettings.ts
+++ b/src/commands/connections/revealConnectionInAppSettings.ts
@@ -7,9 +7,11 @@ import { CosmosDBConnection } from '../../explorer/CosmosDBConnection';
 import { ext } from "../../extensionVariables";
 
 export async function revealConnectionInAppSettings(node: CosmosDBConnection): Promise<void> {
-    const nodeToReveal = await ext.tree.findTreeItem(`${node.parent.parent.parent.appSettingsNode.fullId}/${node.appSettingKey}`);
+    // Ideally this reveals all appSettingKeys, but for now just reveal the first one
+    const firstKey: string = node.appSettingKeys[0];
+    const nodeToReveal = await ext.tree.findTreeItem(`${node.parent.parent.parent.appSettingsNode.fullId}/${firstKey}`);
     if (!nodeToReveal) {
-        throw new Error(`Failed to find app setting with "${node.appSettingKey}" key.`);
+        throw new Error(`Failed to find app setting with key "${firstKey}".`);
     }
     await ext.treeView.reveal(nodeToReveal);
 }
--- a/src/explorer/CosmosDBTreeItem.ts
+++ b/src/explorer/CosmosDBTreeItem.ts
@@ -20,9 +20,9 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
     public readonly label: string = 'Cosmos DB';
     public readonly parent: ConnectionsTreeItem;
 
-    private readonly _endpointSuffix: string = 'ENDPOINT';
-    private readonly _keySuffix: string = 'MASTER_KEY';
-    private readonly _databaseSuffix: string = 'DATABASE_ID';
+    private readonly _endpointSuffix: string = '_ENDPOINT';
+    private readonly _keySuffix: string = '_MASTER_KEY';
+    private readonly _databaseSuffix: string = '_DATABASE_ID';
 
     constructor(parent: ConnectionsTreeItem) {
         super(parent);
@@ -52,8 +52,7 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         // tslint:disable-next-line:strict-boolean-expressions
         const appSettings = (await this.root.client.listApplicationSettings()).properties || {};
         const connections: IDetectedConnection[] = this.detectMongoConnections(appSettings).concat(this.detectDocDBConnections(appSettings));
-        const usedLabels: { [key: string]: boolean } = {};
-        const treeItems = await createTreeItemsWithErrorHandling(
+        const treeItems = <CosmosDBConnection[]>await createTreeItemsWithErrorHandling(
             this,
             connections,
             'invalidCosmosDBConnection',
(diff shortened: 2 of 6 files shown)


src/explorer/CosmosDBTreeItem.ts
```diff
@@ -84,31 +86,31 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
-        appSettings.properties[appSettingKeyToAdd] = databaseToAdd.connectionString;
+        for (const [k, v] of newAppSettings) {
+            appSettings.properties[k] = v;
+        }
+
         await this.root.client.updateApplicationSettings(appSettings);
         await this.parent.parent.appSettingsNode.refresh();
 
-        const createdDatabase = new CosmosDBConnection(this, databaseToAdd, appSettingKeyToAdd);
+        const createdDatabase = new CosmosDBConnection(this, databaseToAdd, newAppSettings.keys()[0]);
```
nturinski: Clarifying comment as to what is going on here?

EricJizbaMSFT: I assume you're just confused by why I'm doing `[0]`?

Refactored how this works. Let me know if it's still confusing.

src/commands/connections/revealConnectionInAppSettings.ts
```diff
@@ -7,9 +7,11 @@ import { CosmosDBConnection } from '../../explorer/CosmosDBConnection';
 import { ext } from "../../extensionVariables";
 
 export async function revealConnectionInAppSettings(node: CosmosDBConnection): Promise<void> {
-    const nodeToReveal = await ext.tree.findTreeItem(`${node.parent.parent.parent.appSettingsNode.fullId}/${node.appSettingKey}`);
+    // Ideally this reveals all appSettingKeys, but for now just reveal the first one
+    const firstKey: string = node.appSettingKeys[0];
+    const nodeToReveal = await ext.tree.findTreeItem(`${node.parent.parent.parent.appSettingsNode.fullId}/${firstKey}`);
```
nturinski: lol what is this? Great-great-grandparent?

EricJizbaMSFT: What - are you ageist :P

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -43,72 +49,63 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
-        appSettings.properties[appSettingKeyToAdd] = databaseToAdd.connectionString;
-        await this.root.client.updateApplicationSettings(appSettings);
+        appSettingsDict.properties = appSettingsDict.properties || {};
+
+        const newAppSettings: Map<string, string> = databaseToAdd.docDBData ?
+            await this.promptForDocDBAppSettings(appSettingsDict, databaseToAdd) :
+            await this.promptForMongoAppSettings(appSettingsDict, databaseToAdd);
+
+        for (const [k, v] of newAppSettings) {
+            appSettingsDict.properties[k] = v;
```
nturinski: Do we need to alert users if they're overwriting an existing setting?

EricJizbaMSFT: `promptForMongoAppSettings` and `promptForDocDBAppSettings ` use `validateInput` to prevent them from getting in that situation

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -120,4 +117,89 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+                const masterKey: string | undefined = appSettings[keyKey];
+
+                if (documentEndpoint && masterKey) {
+                    const keys: string[] = [endpointKey, keyKey];
+                    let connectionString = `AccountEndpoint=${documentEndpoint};AccountKey=${masterKey};`;
+
+                    const databaseKey = prefix + this._databaseSuffix;
+                    if (Object.keys(appSettings).find((k) => k === databaseKey)) {
+                        keys.push(databaseKey);
+                        connectionString += `Database=${appSettings[databaseKey]}`;
```
nturinski: Is this required in the connectionString?

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -120,4 +117,89 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+                    }
+
+                    result.push({ keys, connectionString });
```
nturinski: Looks good to me except I'm confused by this line.  I don't quite understand why you save all the keys as an array.  You only need to add the connection string, right?

EricJizbaMSFT: So the mongo scenario uses a single app setting for the connection string. However, the DocDB scenario uses three app settings:
```
AZURE_COSMOS_ENDPOINT
AZURE_COSMOS_MASTER_KEY
AZURE_COSMOS_DATABASE_ID
```

It does _not_ use the connection string directly. See Matt's sample project here: https://github.com/fiveisprime/node-todo/blob/cosmos/app/models/todo.js#L3

src/explorer/CosmosDBTreeItem.ts
EricJizbaMSFT: Database name is not required. You can see in Matt's sample [here](https://github.com/fiveisprime/node-todo/blob/cosmos/app/models/todo.js#L5), he actually uses `process.env.AZURE_COSMOS_DATABASE_ID || 'todos'`

Issue #707, opened 2018-11-15
Azure Remote Debugging can be started multiple times


mrcrane:
It is possible to issue the "Start Remote Debugging" command multiple times, giving an experience that is confusing to the user. Executing the command a second time will start the process of creating a tunnel proxy again, and in some cases it will even create multiple concurrent debugging sessions.

In particular, this may be an issue when establishing the tunnel is taking a long time, and the user is not seeing positive results. In this case we should display an error message if the user tries to start it again, such as 'Azure Remote Debugging is already started.'

nturinski:
Closing as it is resolved by https://github.com/Microsoft/vscode-azureappservice/pull/710.

PR #710, opened 2018-11-15
Only allow one remote debugging session + other improvements

mrcrane:
Addressing #707 with a very simple approach. Is there any better way to do this?

Also:
- Move remote debugging messaging to ProgressLocation.Notification so that it is more obvious for the user to see that something is happening.
- Add version of AppService container to telemetry to give better insight into which users are having issues.

### DIFF
--- a/src/commands/remoteDebug/startRemoteDebug.ts
+++ b/src/commands/remoteDebug/startRemoteDebug.ts
@@ -17,7 +17,7 @@ let isRemoteDebugging = false;
 
 export async function startRemoteDebug(actionContext: IActionContext, node?: SiteTreeItem): Promise<void> {
     if (isRemoteDebugging) {
-        throw new Error('Azure Remote Debugging is already started.');
+        throw new Error('Azure Remote Debugging is currently starting or already started.');
     }
 
     isRemoteDebugging = true;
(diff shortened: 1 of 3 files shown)


src/commands/remoteDebug/startRemoteDebug.ts
```diff
@@ -13,16 +13,35 @@ import { WebAppTreeItem } from '../../explorer/WebAppTreeItem';
 import { ext } from '../../extensionVariables';
 import * as remoteDebug from './remoteDebugCommon';
 
-export async function startRemoteDebug(node?: SiteTreeItem): Promise<void> {
+let isRemoteDebugging = false;
+
+export async function startRemoteDebug(actionContext: IActionContext, node?: SiteTreeItem): Promise<void> {
+    if (isRemoteDebugging) {
+        throw new Error('Azure Remote Debugging is already started.');
```
EricJizbaMSFT: How long does it take to start remote debugging? If it's a non-negligible amount of time, I think the message should say something like this:
```
Azure Remote Debugging is currently starting or already started.
```

mrcrane: Yes it can take some time to start the debugging session, so this makes sense. I updated the message.

Issue #515, opened 2018-07-12
Log files can be edited and saved successfully 


v-dantch:
Repro Steps:
1. Drill down from Subscription -> web app -> Log Files, then open one log file.
2. Check whether the log file can be edited and saved or not.

Expect:
Fail to edit the log file and cannot save it directly but somewhere locally.

Actual:
Log files can be edited and saved directly.

nturinski:
I think what you're describing is by-design.

Sorry, I might have described it poorly in the test case!  But all Log Files can be edited and saved locally, but the save will not update the remote log files (like what happens with the `Files` node).

Sorry for the confusion!  Am closing with the by-design label.

v-dantch:
Hi Nathan, 

Thanks a lot for your update.
About edit, it is by design that the log files can be edited.

About saving,
   Does 'saved locally' means that the log files will be saved on PC? 
   Actually the remote log files can be updated not saved locally.  Please get the details from \\trinlab2\BAQAFiles\CloudToolsTest\v-dantch\Video\Save.mp4

nturinski:
Oh wow, that is definitely not what happens on my machine.  Thanks for the video-- I'll investigate this further!

PR #703, opened 2018-11-15
Add a delay check to each click to prevent spam opening

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/611

It seems there are some errors by repeatedly spam clicking a file to open multiple copies of it.  This will prevent users from opening the same file over and over again by repeatedly clicking it.

Also refactor `showFile` out of extensions while I was working on it.

PR #711, opened 2018-11-16
RegExp to check for LogFiles/ && LogFiles\

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/515

The cause of this was because the path for windows log Files is LogFiles\ and Linux is LogFiles/.  The RegEx checks for both slashes.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -286,9 +286,8 @@ export function activate(context: vscode.ExtensionContext): void {
     registerCommand('appService.DisableRemoteDebug', async (node?: SiteTreeItem) => disableRemoteDebug(node));
 
     registerCommand('appService.showFile', async (node: FileTreeItem) => {
-        const logFiles: string = 'LogFiles/';
         // we don't want to let users save log files, so rather than using the FileEditor, just open an untitled document
-        if (node.path.startsWith(logFiles)) {
+        if (node.path.toLowerCase().match(/logfiles(\/|\\)/g)) {
             const file: IFileResult = await getFile(node.root.client, node.path);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -286,9 +286,8 @@ export function activate(context: vscode.ExtensionContext): void {
     registerCommand('appService.DisableRemoteDebug', async (node?: SiteTreeItem) => disableRemoteDebug(node));
 
     registerCommand('appService.showFile', async (node: FileTreeItem) => {
-        const logFiles: string = 'LogFiles/';
         // we don't want to let users save log files, so rather than using the FileEditor, just open an untitled document
-        if (node.path.startsWith(logFiles)) {
+        if (node.path.toLowerCase().match(/logfiles(\/|\\)/g)) {
```
EricJizbaMSFT: - Use `^` to make sure this is at the start of the string (Right? `otherFolder\LogFiles\fdsafasd` shouldn't count as logs right?)

nits:
- Use the 'i' flag to indicate "case insensitive" instead of ".toLowerCase" stuff
- Use ".test" instead of ".match" which actually returns a boolean instead of match array
- Don't use 'g' flag (g indicates "global" which means "don't return after first match" which we don't care about here)

```
if (/^LogFiles(\/|\\)/i.test(node.path))
```

nturinski:
I merged the refactor from https://github.com/Microsoft/vscode-azureappservice/pull/703 because I am going to implement this in favor of Stephen's suggestion in registerCommand.

PR #716, opened 2018-11-19
Clean up activate method

EricJizbaMSFT:
!!Review with whitespace off!! lol

- Add basic perf telemetry (copied from functions)
- Return empty `AzureExtensionApiProvider` in case this supports an api in the future

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,10 +5,14 @@
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
Lol I thought about it - but this doesn't feel like the right time to do a refactor like that

PR #717, opened 2018-11-19
Add deployments to DeploymentSlot treeItems

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/700

Seems like Deployments works for Deployment Slots just fine.  I was able to connect, redeploy, view logs, and disconnect with no issues.  Command palette doesn't work, of course, but I don't think we were expecting palette support for slots.

image

### DIFF
--- a/src/explorer/DeploymentSlotTreeItem.ts
+++ b/src/explorer/DeploymentSlotTreeItem.ts
@@ -23,7 +23,6 @@ export class DeploymentSlotTreeItem extends SiteTreeItem {
         this.folderNode = new FolderTreeItem(this, 'Files', "/site/wwwroot");
         this.logFolderNode = new FolderTreeItem(this, 'Logs', '/LogFiles', 'logFolder');
         this.appSettingsNode = new AppSettingsTreeItem(this);
-
     }
 
     public get iconPath(): { light: string, dark: string } {
--- a/src/explorer/DeploymentSlotTreeItem.ts
+++ b/src/explorer/DeploymentSlotTreeItem.ts
@@ -11,13 +11,6 @@ import { FolderTreeItem } from './FolderTreeItem';
 import { SiteTreeItem } from './SiteTreeItem';
 
 export class DeploymentSlotTreeItem extends SiteTreeItem {
-
-    public get iconPath(): { light: string, dark: string } {
-        return {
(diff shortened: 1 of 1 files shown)


src/explorer/DeploymentSlotTreeItem.ts
```diff
@@ -3,15 +3,24 @@
...
 import * as path from 'path';
-import { AppSettingsTreeItem, ISiteTreeRoot, SiteClient } from 'vscode-azureappservice';
+import { AppSettingsTreeItem, DeploymentsTreeItem, ISiteTreeRoot, SiteClient } from 'vscode-azureappservice';
 import { AzureParentTreeItem, AzureTreeItem } from 'vscode-azureextensionui';
 import { FolderTreeItem } from './FolderTreeItem';
 import { SiteTreeItem } from './SiteTreeItem';
 
 export class DeploymentSlotTreeItem extends SiteTreeItem {
+
+    public get iconPath(): { light: string, dark: string } {
```
EricJizbaMSFT: nit: Why'd you move iconPath to the top of the class? I like it better where it was

src/explorer/DeploymentSlotTreeItem.ts
```diff
@@ -3,15 +3,24 @@
...
+        return {
+            light: path.join(__filename, '..', '..', '..', '..', 'resources', 'light', 'DeploymentSlot_color.svg'),
+            dark: path.join(__filename, '..', '..', '..', '..', 'resources', 'dark', 'DeploymentSlot_color.svg')
+        };
+    }
     public static contextValue: string = 'deploymentSlot';
     public readonly contextValue: string = DeploymentSlotTreeItem.contextValue;
+    public deploymentsNode: DeploymentsTreeItem | undefined;
```
EricJizbaMSFT: nit: readonly

src/explorer/DeploymentSlotTreeItem.ts
nturinski: I think it got automatically moved on save.

src/explorer/DeploymentSlotTreeItem.ts
nturinski: Can't, I'm assigning it in `loadMoreChildrenImpl`.

PR #718, opened 2018-11-19
Improve detectDocDBConnections to detect actual connection string

EricJizbaMSFT:
The more common pattern is to have three application settings "_ENDOINT", "_KEY", and "_DATABASE_ID", but it's easy enough for us to also detect an actual sql connection string and display that.


nturinski:
What does a SQL connection string look like??

EricJizbaMSFT:
`AccountEndpoint=https://emjsql.documents.azure.com:443/;AccountKey=...`

Issue #631, opened 2018-10-05
Improve sorting and alignment under the Files node


kraigb:
At present, when expanding the Files node for an App Service, the icon used for files is not aligned with the name of folders. Specifically, folders are a few pixels left of the file icons:

![image](https://user-images.githubusercontent.com/1339874/46509703-20732000-c7f9-11e8-8468-d34b5bc3283c.png)

Furthermore, all files and folders are sorted alphabetically together, rather than having folders appear first.

The upshot of this is that files that occur after a folder in the list _appear_ to be children of that folder, when in fact they're not.

Showing folders first (alphabetically), then files, and aligning the icons with the folder text, provide the right visual indicators.

PR #719, opened 2018-11-19
Fix sort and icons for folders

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/631

Puts folders above files. Also adds icons to subFolders. I know @nturinski didn't use icons for subFolders to match VS Code's file explorer, but I just don't think it's worth it given the alignment issues.

Icon before:
screen shot 2018-11-19 at 5 39 30 pm

Icon after:
screen shot 2018-11-19 at 5 40 44 pm

"settings.json" should definitely be indented _more_ than ".gitignore"

### DIFF
--- a/src/explorer/FolderTreeItem.ts
+++ b/src/explorer/FolderTreeItem.ts
@@ -71,10 +71,10 @@ export class FolderTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
 }
 
 // tslint:disable-next-line:no-any
-function instanceOfCompare<T>(ti1: AzureTreeItem, ti2: AzureTreeItem, type1: new (...args: any[]) => T): number | undefined {
-    if (!(ti1 instanceof type1) && ti2 instanceof type1) {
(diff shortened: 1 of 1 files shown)


src/explorer/FolderTreeItem.ts
```diff
@@ -60,13 +60,24 @@ export class FolderTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
         }
-        return ti1.label.localeCompare(ti2.label);
+
+        return result === undefined ? ti1.label.localeCompare(ti2.label) : result;
+    }
+}
+
+// tslint:disable-next-line:no-any
+function instanceOfCompare<T>(ti1: AzureTreeItem, ti2: AzureTreeItem, type1: new (...args: any[]) => T): number | undefined {
+    if (!(ti1 instanceof type1) && ti2 instanceof type1) {
```
nturinski: Why call it type1?  Are there going to be more types to compare?  Variable name seems unclear to me.

EricJizbaMSFT: Oh uh `type` is a reserved keyword so I got a tslint error break or something. Changed it a little

PR #737, opened 2018-11-28
Group deploy commands together

EricJizbaMSFT:
See here for discussion: https://github.com/Microsoft/vscode-azurestorage/pull/183#discussion_r198285619

These have been "ungrouped" for a few months now and with several releases coming up it feels like a good time to fix this:
screen shot 2018-11-28 at 9 41 34 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -697,12 +697,12 @@
                 {
                     "command": "appService.Deploy",
                     "when": "explorerResourceIsFolder == true",
-                    "group": "zzz_@appServiceDeployCommands"
+                    "group": "zzz_deploy@1"
                 },
                 {
                     "command": "appService.Deploy",
                     "when": "explorerResourceIsFolder == false && resourceExtname =~ /.[j|w]ar$/",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -697,12 +697,12 @@
...
                     "command": "appService.Deploy",
                     "when": "explorerResourceIsFolder == true",
-                    "group": "zzz_@appServiceDeployCommands"
+                    "group": "zzz_deploy@1"
                 },
                 {
                     "command": "appService.Deploy",
                     "when": "explorerResourceIsFolder == false && resourceExtname =~ /.[j|w]ar$/",
-                    "group": "zzz_@appServiceDeployCommands"
+                    "group": "zzz_deploy@1"
```
StephenWeatherford: Can we change the name to something that makes it obvious it's cross-extension?  Otherwise someone may break them again unknowingly.

E.g. zzz_azuretools_deploy?

EricJizbaMSFT: Sure

PR #763, opened 2018-12-03
Pass in 'Connect to GitHub' command id to DeploymentsTreeItem

nturinski:
Depends on Microsoft/vscode-azuretools#349

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -30,5 +30,6 @@
         "repos",
         "runtimes",
         "vsce"
-    ]
+    ],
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT:
FYI - remember you need to update the package-lock every time you bump a package from now on

nturinski:
... ruh roh.  Right, so that's just doing an `npm install` right?

EricJizbaMSFT:
Yep - but of course `npm install` won't work until after the shared package is published

nturinski:
Naturally 😎

PR #766, opened 2018-12-05
Nat/python deployment

### LINKED ISSUE #642: Smart defaults for Python
I'll work with the Python team to better understand the heuristics and defaults.

nturinski:
Fixes #642 in combination with https://github.com/Microsoft/vscode-azuretools/pull/352

For zipignore, I was wondering if I should just import [this](https://github.com/github/gitignore/blob/master/Python.gitignore) `.gitignore` file and parse it. I assume anything that isn't saved in a git repo shouldn't be deployed either?

The `.deployment` file will trigger a build, Kudu detects that it is Python, and deployment works with zipdeploy.  Local git will work as is and does not require a `.deployment` file.

The deployment will break without the scmDoBuildDeploy flag, so I just enable it rather than prompting the user in this case.


src/constants.ts
```diff
@@ -15,13 +15,16 @@ export enum runtimes {
...
+    javase = 'java|8-jre8',
+    python = 'python'
 }
 
 export function getIgnoredFoldersForDeployment(runtime: string): string[] {
     switch (runtime) {
         case runtimes.node:
             return ['node_modules{,/**}'];
+        case runtimes.python:
+            return ['.env', '.Python', 'env/{,/**}', 'build{,/**}', 'dist{,/**}'];
```
EricJizbaMSFT: Why do some folders have `{,/**}` and some don't?

src/commands/deploy.ts
```diff
@@ -115,10 +115,16 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
-                // tslint:disable-next-line:no-unsafe-any
-                await node.enableScmDoBuildDuringDeploy(fsPath, constants.runtimes[siteConfig.linuxFxVersion.substring(0, siteConfig.linuxFxVersion.indexOf('|'))], context.properties);
+            //check if node is being zipdeployed and that there is no .deployment file
+            if (siteConfig.linuxFxVersion && siteConfig.scmType === 'None' && !(await pathExists(path.join(fsPath, constants.deploymentFileName)))) {
+                if (siteConfig.linuxFxVersion.startsWith(constants.runtimes.node)) {
+                    // if it is node, prompt the user (as we can break them)
+                    await node.promptScmDoBuildDeploy(fsPath, constants.runtimes[constants.runtimes.node], context.properties);
+                } else if (siteConfig.linuxFxVersion.startsWith(constants.runtimes.python)) {
+                    // if it is a python app, enable it for them as it does not work without build during deploy
+                    await node.enableScmDoBuildDuringDeploy(fsPath, constants.runtimes.python);
```
EricJizbaMSFT: > it does not work without build during deploy

Can you point to docs for this? And will this always be the case? I don't like this for two reasons:
1. It happens on every deploy and users have no way to override the behavior if they don't want it
1. I think it's confusing when tooling changes a user's project without telling them

src/constants.ts
EricJizbaMSFT: How did you pick these virtual environment names? Couldn't users name it almost anything?

EricJizbaMSFT: `.env` is one of the most common names for python virtual environments. It's definitely a folder

nturinski: I thought `.env` was a file describing environment variables https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/com.ibm.aix.osdevice/env_file.htm where env is the folder for virtual environments

src/commands/deploy.ts
nturinski: Okay so, I tried creating the venv and then pip install the requirements and deployed that and it worked.  I'll handle it the same way we handle node.js.

src/constants.ts
```diff
@@ -15,13 +15,16 @@ export enum runtimes {
...
-    javase = 'java|8-jre8'
+    javase = 'java|8-jre8',
+    python = 'python'
 }
 
 export function getIgnoredFoldersForDeployment(runtime: string): string[] {
     switch (runtime) {
         case runtimes.node:
             return ['node_modules{,/**}'];
+        case runtimes.python:
```
EricJizbaMSFT: I still think users name a virtual environment whatever they want. Ideally we're smarter about this. For example, you can "detect" a venv by check the path for an activate script. Something similar to https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/commands/createNewProject/PythonProjectCreator.ts#L200

That being said, I'm fine if you hard-code the most common "envs" here and file an issue to be smarter later

src/constants.ts
```diff
@@ -15,13 +15,16 @@ export enum runtimes {
...
+            return ['.env', 'env/{,/**}', 'build{,/**}', 'dist{,/**}'];
```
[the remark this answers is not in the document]
nturinski: Yeah, that's a good point.  I can't get a definitive answer whether or not to ignore `.env` files on a deploy.  Furthermore, I think some developers use `dotenv` for production code, so their `.env` file might be required anyway.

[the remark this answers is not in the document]
nturinski: Also did some testing to confirm, but you do need the `/{,/**}` for it to ignore the directory.

src/constants.ts
nturinski: Yeah, fair enough.  I'll add the environments that the .gitignore lists for now and make a smarter solution in a separate PR.

src/constants.ts
```diff
@@ -15,13 +15,16 @@ export enum runtimes {
     dotnetcore = 'dotnetcore',
     ruby = 'ruby',
     tomcat = 'tomcat',
-    javase = 'java|8-jre8'
+    javase = 'java|8-jre8',
+    python = 'python'
 }
 
 export function getIgnoredFoldersForDeployment(runtime: string): string[] {
```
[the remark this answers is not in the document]
nturinski: 😬 Ugh, what was I _thinking?!_

nturinski:
I added a new check to see if the folder we are going to add to the `zipIgnorePattern` exists within the project.  A potential downfall could be if a user adds `node_modules` after the first deployment, we wouldn't catch it, but we also wouldn't add a bunch of unnecessary settings that could be confusing.

I'm thinking it might not be worth adding but wanted your opinion about it.

src/explorer/SiteTreeItem.ts
```diff
@@ -166,6 +157,28 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+        const zipIgnoreFolders: string[] = this.getIgnoredFoldersForDeployment(runtime);
+        let oldSettings: string[] | string | undefined = workspace.getConfiguration(constants.extensionPrefix, Uri.file(fsPath)).get(constants.configurationSettings.zipIgnorePattern);
+        if (!oldSettings) {
+            oldSettings = [];
+        } else if (typeof oldSettings === "string") {
+            oldSettings = [oldSettings];
+            // settings have to be an array to concat the proper zipIgnoreFolders
+        }
+        const newSettings: string[] = oldSettings;
+        for (const folder of zipIgnoreFolders) {
```
EricJizbaMSFT: Yeah I don't like the check for 'pathExists'. There's no harm in ignoring a folder that doesn't exist

src/explorer/SiteTreeItem.ts
```diff
@@ -182,4 +195,15 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
             telemetryProperties.promptToSaveDeployConfigs = 'Skip for now';
         }
     }
+
+    private getIgnoredFoldersForDeployment(runtime: string): string[] {
+        switch (runtime) {
+            case runtimes.node:
+                return ['node_modules{,/**}'];
+            case runtimes.python:
+                return ['env/{,/**}', 'venv{,/**}', 'lib{,/**}', 'lib64{,/**}'];
```
EricJizbaMSFT: You keep changing this list - how are you picking these??

nturinski: Just looking at a .gitignore file on GitHub and picking ones that seem relevant (like pip installs into a Lib folder, as far as I can tell) https://github.com/github/gitignore/blob/master/Python.gitignore

EricJizbaMSFT: I would prefer that all envs listed in the gitignore were included here before checking in (on the off chance that we don't implement the "auto-detect" before releasing)
aka all of these:
```
# Environments
.env
.venv
env/
venv/
ENV/
env.bak/
venv.bak/
```

src/explorer/SiteTreeItem.ts
```diff
@@ -182,4 +192,15 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
```
EricJizbaMSFT: 'env/{,/**}' has an extra slash

src/explorer/SiteTreeItem.ts
```diff
@@ -182,4 +192,15 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+            default:
```
EricJizbaMSFT: Please file an issue and/or send an email to the python team about the non-virtual environment values as well. You picked two values (lib and lib64) out of a pretty long list from the "# Distribution / packaging" section and it still seems arbitrary. I also think we might want to include the values from the "# Byte-compiled / optimized / DLL files" section

nturinski:
When you say the list of changed files will change before release, are you referring to the non-virtual environment values or the virtual environment detection logic or both?

EricJizbaMSFT:
Both. For virtual environment values I don't think auto-detection is required for release, but the list of hard-coded values should at least be longer than two values in that case

PR #771, opened 2018-12-06
Implement obfuscating app settings

### LINKED ISSUE #640: Obfuscate environment variable values
Default should be to not show the value in the tree nodes or tooltips and still allow editing. Add a config option to override this behavior.

nturinski:
Fixes #640 
Relies on this PR https://github.com/Microsoft/vscode-azuretools/pull/353


package.json
```diff
@@ -274,6 +274,11 @@
                 "command": "appService.RevealConnectionInAppSettings",
                 "title": "Reveal in Application Settings",
                 "category": "Azure App Service"
+            },
+            {
+                "command": "appService.toggleValueVisability",
```
EricJizbaMSFT: Visability -> Visibility

package.json
```diff
@@ -274,6 +274,11 @@
...
+                "title": "Reveal App Settings",
```
EricJizbaMSFT: Remove or change the title to reflect the actual command

src/commands/toggleValueVisability.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AppSettingTreeItem } from "vscode-azureappservice";
+
+export async function toggleValueVisability(node: AppSettingTreeItem): Promise<void> {
+    node.toggleValueVisability();
+    await node.refresh();
```
EricJizbaMSFT: Shouldn't `toggleValueVisability` handle the refresh?

EricJizbaMSFT: Also if this ends up being one line of code, I think I'm fine with it going directly in extension.ts

nturinski: Good point

package.json
nturinski: 😊

nturinski:
image
image

Issue #639, opened 2018-10-12
Import/upload app settings from .env file


fiveisprime:
Similar to how Functions uploads settings: right-click on app settings, choose "import from .env..."

I'm not familiar enough with the variations of the format, but the extension should support with and without `export` statements.

EricJizbaMSFT:
@fiveisprime duplicate of https://github.com/Microsoft/vscode-azureappservice/issues/386?

nnja:
@EricJizbaMSFT Close, but quite since #386 requests JSON as a format. A typical `.env` file looks like this for local use:

```shell
# Comment
EXPORT KEY="VALUE"
EXPORT KEY1="VALUE1"

# Quotes are optional if there are no special characters in the value.
EXPORT KEY2="VALUE2"

```
This allows you to run `source .env` from the command line to load all the env variables.

Or like 

```shell
# Comment
KEY=VALUE
KEY1=VALUE1
KEY2=VALUE2
```

for use with [Docker compose](https://docs.docker.com/compose/env-file/) to allow you to pass environment variables to the container.

EricJizbaMSFT:
Thanks for more detail @nnja. I think @nturinski just mentioned json as one possibility, but he cared more about the ability to upload/download app settings regardless of file type. I went ahead and closed that bug as 'duplicate' since '.env' sounds more useful

PR #779, opened 2018-12-11
Download App Settings as a .env and Upload local .env files as App Settings

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/639

I'll add `dotenv` to the thirdparty website after it is approved.

I need some help (probably from @fiveisprime) for the wording here.  These aren't exactly local settings and app settings.  While `.env` files are local, they are environment variables.  When the user uploads their local env vars to app settings, what should I refer to them as??  The titles on the commands are just placeholder for now.


fiveisprime:
What about import/export instead of upload/download? I'm not sold on that.. I kind of like upload/download because it makes it clear what the target is based on the language whereas import/export is generic enough to be confusing. :(

It should either be "App Settings" or "Application Settings." though for sure

nturinski:
In Functions, we use the language "Download Remote Settings" and "Upload Local Settings".  Should we just keep it consistent?

fiveisprime:
Yeah, just follow the Functions pattern for now.

package.json
```diff
@@ -613,6 +625,16 @@
                     "when": "view == azureAppService && viewItem == applicationSettings",
                     "group": "2_appSettingsGeneral@1"
                 },
+                {
+                    "command": "appService.appSettings.Download",
+                    "when": "view == azureAppService && viewItem == applicationSettings",
+                    "group": "1_appSettingEnv@1"
+                },
+                {
+                    "command": "appService.appSettings.Upload",
```
EricJizbaMSFT: This command should also show up when they right click on the file

src/commands/appSettings/confirmOverwriteSettings.ts
```diff
@@ -0,0 +1,80 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { DialogResponses } from "vscode-azureextensionui";
+import { ext } from "../../extensionVariables";
+
+export async function confirmOverwriteSettings(sourceSettings: { [key: string]: string }, destinationSettings: { [key: string]: string }, destinationName: string): Promise<void> {
```
EricJizbaMSFT: This function feels like too much code to just copy from functions. I think it should be in shared package

src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    const client: SiteClient = node.root.client;
+
+    const message: string = 'Select the destination file for your downloaded settings.';
+    const envVarPath: string = await workspaceUtil.selectWorkspaceFile(ext.ui, message, () => envFileName);
+    const envVarUri: vscode.Uri = vscode.Uri.file(envVarPath);
+
+    await node.runWithTemporaryDescription('Downloading...', async () => {
+        ext.outputChannel.appendLine(`Downloading settings from "${client.fullName}"...`);
```
EricJizbaMSFT: Do we really need to keep duplicating these messages in both the notifications and output channel? Seems like it just makes the code harder to read without providing much value

src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const doc: vscode.TextDocument = await vscode.workspace.openTextDocument(envVarUri);
+    await vscode.window.showTextDocument(doc);
+}
+
+export function convertAppSettingsToEnvVariables(appSettings: { [propertyName: string]: string } | undefined, appName: string): string {
+    let envData: string = `# Imported Application Settings from ${appName}${os.EOL}`;
+    if (appSettings) {
+        for (const property of Object.keys(appSettings)) {
+            envData += `${property}="${appSettings[property]}"`;
```
EricJizbaMSFT: Will this work if the value has quotes in it?

src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -0,0 +1,54 @@
...
+        await fse.writeFile(envVarPath, convertAppSettingsToEnvVariables(localEnvVariables, client.fullName));
+    });
+
+    const doc: vscode.TextDocument = await vscode.workspace.openTextDocument(envVarUri);
+    await vscode.window.showTextDocument(doc);
+}
+
+export function convertAppSettingsToEnvVariables(appSettings: { [propertyName: string]: string } | undefined, appName: string): string {
+    let envData: string = `# Imported Application Settings from ${appName}${os.EOL}`;
+    if (appSettings) {
```
EricJizbaMSFT: When can appsettings be undefined? I would expect an error

src/utils/workspace.ts
```diff
@@ -0,0 +1,51 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { IAzureQuickPickItem, IAzureUserInput } from 'vscode-azureextensionui';
+
+export async function selectWorkspaceFile(ui: IAzureUserInput, placeHolder: string, getSubPath?: (f: vscode.WorkspaceFolder) => string | undefined): Promise<string> {
```
EricJizbaMSFT: This repo already has some workspace functions here: https://github.com/Microsoft/vscode-azureappservice/blob/master/src/util.ts#L41

Can we merge these?

src/commands/appSettings/downloadAppSettings.ts
nturinski: Seems okay!  These are the tests I tried.

image
image

src/commands/appSettings/downloadAppSettings.ts
nturinski: Do you mean we should make a constant for `Downloading settings from "${client.fullName"`?  I'm not clear what you're trying to suggest here

src/commands/appSettings/downloadAppSettings.ts
nturinski: The `dotenv.parse` function returned `undefined` if it cannot parse any settings.  In this case, I return `{}` instead and alert the user that we didn't find any local settings.  But you're right in downloadAppSettings.  It will always at least be an empty object if there weren't any settings (an empty file for instance)

nturinski:
Relies on PR https://github.com/Microsoft/vscode-azuretools/pull/357

src/commands/appSettings/downloadAppSettings.ts
EricJizbaMSFT: I'm saying get rid of this line:
```
ext.outputChannel.appendLine(`Downloading settings from "${client.fullName}"...`);
```

It's a somewhat common pattern that you do, but it's starting to feel redundant and can honestly make the code look a little confusing.

package.json
```diff
@@ -176,6 +178,16 @@
...
                 "category": "Azure App Service"
             },
+            {
+                "command": "appService.appSettings.Download",
+                "title": "Download Remote Settings",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "appService.appSettings.Upload",
+                "title": "Upload Local Settings",
```
EricJizbaMSFT: Both of these commands should have an ellipses since they ask for more information

src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -0,0 +1,51 @@
...
+        if (remoteEnvVariables.properties) {
+            await node.confirmOverwriteSettings(remoteEnvVariables.properties, localEnvVariables, envFileName);
+        }
+
+        await fse.ensureFile(envVarPath);
+        await fse.writeFile(envVarPath, convertAppSettingsToEnvVariables(localEnvVariables, client.fullName));
+    });
+
+    const doc: vscode.TextDocument = await vscode.workspace.openTextDocument(envVarUri);
+    await vscode.window.showTextDocument(doc);
```
EricJizbaMSFT: I think we need to discuss this. For example, I like seeing this file right away, but based on your recent PR to obfuscate settings by default, we might not want to open a file that has the same information as app settings.

Maybe we can "focus" on the file in the file view without actually opening it? Not sure if that's possible. Also you could do a notification that says "Finished downloading" or whatever with a "View file" button. I think that more closely mirrors the "Click to view" thing

src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function convertAppSettingsToEnvVariables(appSettings: { [propertyName: string]: string }, appName: string): string {
+    let envData: string = `# Imported Application Settings from ${appName}${os.EOL}`;
```
EricJizbaMSFT: nit: "Downloaded" instead of "Imported" to stay consistent

src/commands/appSettings/uploadAppSettings.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function uploadAppSettings(target?: Uri | AppSettingsTreeItem | undefined): Promise<void> {
+    const message: string = 'Select the local .env file to upload.';
+    let node: AppSettingsTreeItem | undefined;
+    let envPath: string;
+    if (target instanceof Uri) {
+        envPath = target.fsPath;
+    } else {
+        node = target;
+        envPath = await workspaceUtil.selectWorkspaceFile(ext.ui, message, () => envFileName);
+
```
EricJizbaMSFT: nit: fix this weird blank line spacing

src/commands/appSettings/uploadAppSettings.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as dotenv from 'dotenv';
+import { Uri } from "vscode";
+import { AppSettingsTreeItem, SiteClient } from "vscode-azureappservice";
+import { envFileName } from "../../constants";
+import { ext } from "../../extensionVariables";
+import * as workspaceUtil from '../../utils/workspace';
+import { getLocalEnvironmentVariables } from "./getLocalEnvironmentVariables";
+
+export async function uploadAppSettings(target?: Uri | AppSettingsTreeItem | undefined): Promise<void> {
+    const message: string = 'Select the local .env file to upload.';
```
EricJizbaMSFT: nit: Move message down closer to where it's actually used (inside the "else" case)

src/commands/appSettings/uploadAppSettings.ts
```diff
@@ -0,0 +1,46 @@
...
+    } else {
+        node = target;
+        envPath = await workspaceUtil.selectWorkspaceFile(ext.ui, message, () => envFileName);
+
+    }
+    if (!node) {
+        node = <AppSettingsTreeItem>await ext.tree.showTreeItemPicker(AppSettingsTreeItem.contextValue);
+    }
+    const client: SiteClient = node.root.client;
+    await node.runWithTemporaryDescription('Uploading...', async () => {
```
EricJizbaMSFT: I still feel like I have no indication that this succeeded. I would prefer a notification saying "Finished uploading app settings"

src/utils/workspace.ts
```diff
@@ -0,0 +1,98 @@
...
+        });
+
+        folderPicks.push({ label: '$(file-directory) Browse...', description: '', data: undefined });
+        folder = await ui.showQuickPick(folderPicks, { placeHolder });
+    }
+
+    return folder && folder.data ? folder.data : (await ui.showOpenDialog(options))[0].fsPath;
+}
+
+export async function showWorkspaceFoldersQuickPick(placeHolderString: string, telemetryProperties: TelemetryProperties, subPathSetting: string | undefined): Promise<string> {
```
EricJizbaMSFT: So you copied this over, but you didn't actually merge the functions. This should call `selectWorkspaceItem`, similar to https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/utils/workspace.ts#L15

src/commands/appSettings/downloadAppSettings.ts
nturinski: image

src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -22,27 +24,31 @@ export async function downloadAppSettings(node?: AppSettingsTreeItem): Promise<v
...
         if (remoteEnvVariables.properties) {
-            await node.confirmOverwriteSettings(remoteEnvVariables.properties, localEnvVariables, envFileName);
+            await confirmOverwriteSettings(remoteEnvVariables.properties, localEnvVariables, envFileName);
         }
 
         await fse.ensureFile(envVarPath);
         await fse.writeFile(envVarPath, convertAppSettingsToEnvVariables(localEnvVariables, client.fullName));
     });
 
+    const input: string | undefined = await window.showInformationMessage(`Downloaded settings from "${client.fullName}".  View settings file?`, 'Open file');
```
EricJizbaMSFT: No strong preference between the two, but this should be consistent between "Open" and "View"

PR #791, opened 2018-12-28
Nat/automated test suites

nturinski:
I changed the deploy after create prompt to not await because it was screwing up the tests since showInformationWindow is not called from the extetensionVariable UI.  While we should solve this eventually, I don't think that this was a prompt we should await anyway.


src/extension.ts
```diff
@@ -193,32 +193,35 @@ export async function activate(context: vscode.ExtensionContext): Promise<AzureE
...
-            if (await vscode.window.showInformationMessage('Deploy to web app?', yesButton, noButton) === yesButton) {
-                this.properties[deployingToWebApp] = 'true';
-                await deploy(this, false, createdApp);
-            } else {
-                this.properties[deployingToWebApp] = 'false';
-            }
+            vscode.window.showInformationMessage('Deploy to web app?', yesButton, noButton).then(
+                async (input: vscode.MessageItem) => {
+                    if (input === yesButton) {
+                        this.properties[deployingToWebApp] = 'true';
```
EricJizbaMSFT: I'm fine with not awaiting this (that's why `showInformationMessage` isn't on the shared AzureUserInput since information messages should almost never be awaited), but be aware you probably wont' get this telemetry anymore

test/index.ts
```diff
@@ -15,6 +15,8 @@
 // to report the results back to the caller. When the tests are finished, return
 // a possible error to the callback or null if none.
 
+// tslint:disable-next-line:no-var-requires
```
EricJizbaMSFT: nit: I don't think you're doing a var-requires in the next line?

test/index.ts
```diff
@@ -36,6 +38,16 @@ const options: { [key: string]: string | boolean | number } = {
 //
 // See https://github.com/mochajs/mocha/wiki/Using-mocha-programmatically#set-options for all available options
 
+// For local testing, dotenv will load in the local .env file.  Change the path variable to the absolute path of the .env file.
+const localEnvVars: dotenv.DotenvConfigOutput | undefined = dotenv.config({ path: 'C:/Users/naturins/Desktop/Code/vscode-azureappservice/.env' });
```
EricJizbaMSFT: Soo we definitely don't need to have your 'naturins' path checked in - but I also don't understand why all of this ".env" stuff is necessary anyways. Can't you just set environment variables on your system?

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,104 @@
...
+                console.log(`Ignoring resource group "${resourceGroup}" because it does not exist.`);
+            }
+        }
+        ext.tree.dispose();
+    });
+
+    test('Create and Delete New Web App (Basic)', async () => {
+        const appName: string = getRandomHexString().toLowerCase();
+        await vscode.workspace.getConfiguration(constants.extensionPrefix).update('advancedCreation', false, vscode.ConfigurationTarget.Global);
+        const testInputs: string[] = [appName, 'Linux', 'Node.js 10.10 (LTS - Recommended for new apps)'];
```
EricJizbaMSFT: So we have to update this test every time the LTS changes? Can we refactor so that that's not necessary?

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,104 @@
...
+        ext.ui = new TestUserInput(testInputs);
+
+        await vscode.commands.executeCommand('appService.CreateWebApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const defaultRgName: string = 'appsvc_rg_linux_centralus';
```
EricJizbaMSFT: I don't see you adding this to `resourceGroupsToDelete`

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+function getWebsiteManagementClient(testAccount: TestAzureAccount): WebSiteManagementClient {
+    return new WebSiteManagementClient(testAccount.getSubscriptionCredentials(), testAccount.getSubscriptionId());
+}
+
+function getResourceManagementClient(testAccount: TestAzureAccount): ResourceManagementClient {
+    return new ResourceManagementClient(testAccount.getSubscriptionCredentials(), testAccount.getSubscriptionId());
+}
+
+function getRandomHexString(length: number = 10): string {
```
EricJizbaMSFT: You've also got this function here: https://github.com/Microsoft/vscode-azureappservice/blob/master/src/commands/deploy.ts#L188

Time to put it in a shared location

src/extension.ts
nturinski: I'm not sure if this very useful telemetry actually-- I think I'm okay with losing it.

test/index.ts
nturinski: Yes, we could, but it's harder to explain and to set up.  Plus I don't know if setting environment variables are the same on every platform, but `dotenv` should work for everything.  For other extensions that don't include `dotenv` already, it can just be a devDependency.

test/createAzureResource.test.ts
nturinski: Because 'appsrv_rg_linux_centralus" is the default name for our basic creation and we wouldn't want to delete that entire rg if users had created apps in there

EricJizbaMSFT: That's a valid point, but I still don't think you can risk tests constantly creating resources without deleting them. It's also not a problem in CI since we have a separate subscription for tests.

I think we have to refactor somehow. The first idea that comes to mind is an environment variable to turn on/off the delete behavior

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,103 @@
...
+import * as vscode from 'vscode';
+import { deleteDefaultResourceGroup, longRunningTestsEnabled } from './global.test';
+import { AzureTreeDataProvider, constants, DialogResponses, ext, getRandomHexString, ILinuxRuntimeStack, SiteRuntimeStep, TestAzureAccount, TestUserInput, WebAppProvider } from '../extension.bundle';
+
+suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Promise<void> {
+    this.timeout(1200 * 1000);
+    const resourceGroupsToDelete: string[] = [];
+    const testAccount: TestAzureAccount = new TestAzureAccount();
+    // tslint:disable-next-line:no-backbone-get-set-outside-model
+    const oldAdvancedCreationSetting: boolean | undefined = <boolean>vscode.workspace.getConfiguration(constants.extensionPrefix).get('advancedCreation');
```
EricJizbaMSFT: Almost all of this logic should be in suiteSetup. You don't need to execute it unless `longRunningTestsEnabled` is true.

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,103 @@
...
+import { IHookCallbackContext, ISuiteCallbackContext } from 'mocha';
+import * as vscode from 'vscode';
+import { deleteDefaultResourceGroup, longRunningTestsEnabled } from './global.test';
+import { AzureTreeDataProvider, constants, DialogResponses, ext, getRandomHexString, ILinuxRuntimeStack, SiteRuntimeStep, TestAzureAccount, TestUserInput, WebAppProvider } from '../extension.bundle';
+
+suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Promise<void> {
+    this.timeout(1200 * 1000);
+    const resourceGroupsToDelete: string[] = [];
+    const testAccount: TestAzureAccount = new TestAzureAccount();
+    // tslint:disable-next-line:no-backbone-get-set-outside-model
```
EricJizbaMSFT: nit: Just disable `no-backbone-get-set-outside-model` for this repo. It's for a framework that we don't use

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    const oldAdvancedCreationSetting: boolean | undefined = <boolean>vscode.workspace.getConfiguration(constants.extensionPrefix).get('advancedCreation');
+    const runtimeLTS: ILinuxRuntimeStack | undefined = new SiteRuntimeStep().getLinuxRuntimeStack().find((runtime: ILinuxRuntimeStack) => {
+        return runtime.displayName.includes('LTS - Recommended for new apps');
```
EricJizbaMSFT: I suppose this fixes this specific case, but I would prefer to see a more generic solution that helps other test authors as well.

For example, change the constructor of `TestUserInput` from `constructor(inputs: (string | undefined)[]) {` to `constructor(inputs: (string | RegExp | undefined)[]) {` so that you can just pass in a RegExp with 'LTS' in it

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        const resourceName: string = getRandomHexString().toLowerCase();
+        await vscode.workspace.getConfiguration(constants.extensionPrefix).update('advancedCreation', true, vscode.ConfigurationTarget.Global);
+        const testInputs: string[] = [resourceName, '$(plus) Create new resource group', resourceName, 'Linux', runtimeLTS.displayName, '$(plus) Create new App Service plan', resourceName, 'B1', 'West US'];
+        ext.ui = new TestUserInput(testInputs);
+
+        await vscode.commands.executeCommand('appService.CreateWebApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName, resourceName);
+        assert.ok(createdApp);
+        resourceGroupsToDelete.push(resourceName);
```
EricJizbaMSFT: You need to push the name to resourceGroupsToDelete before you try creating any resource. It could've failed to create the web app, but it might've already created the resource group or plan. These are running nightly so I want to be very careful to make sure we're cleaning up azure resources.

test/global.test.ts
```diff
@@ -5,12 +5,17 @@
 
 import { IHookCallbackContext } from 'mocha';
 import * as vscode from 'vscode';
+import { TestOutputChannel } from 'vscode-azureextensiondev';
+
+import { ext } from '../extension.bundle';
 
 // tslint:disable-next-line:strict-boolean-expressions export-name
 export let longRunningTestsEnabled: boolean = !/^(false|0)?$/i.test(process.env.ENABLE_LONG_RUNNING_TESTS || '');
+export let deleteDefaultResourceGroup: boolean = !/^(false|0)?$/i.test(process.env.DELETE_DEFAULT_RESOURCE_GROUP || '');
```
EricJizbaMSFT: I think you need to flip the name/logic of this flag so that it's on by default. I really want to make sure these tests aren't leaving around azure resources.

nturinski:
Relies on this PR https://github.com/Microsoft/vscode-azuretools/pull/389

src/extension.ts
EricJizbaMSFT: If you're okay with losing it, I would prefer to remove the telemetry rather than get inconsistent telemetry

test/global.test.ts
```diff
@@ -5,12 +5,18 @@
...
 import * as vscode from 'vscode';
+import { TestOutputChannel } from 'vscode-azureextensiondev';
 
-// tslint:disable-next-line:strict-boolean-expressions export-name
+import { ext } from '../extension.bundle';
+
+// tslint:disable-next-line:strict-boolean-expressions
 export let longRunningTestsEnabled: boolean = !/^(false|0)?$/i.test(process.env.ENABLE_LONG_RUNNING_TESTS || '');
+// tslint:disable-next-line:strict-boolean-expressions
+export let protectDefaultResources: boolean = !/^(false|0)?$/i.test(process.env.PROTECT_DEFAULT_RESOURCES || '');
```
EricJizbaMSFT: Lol "PROTECT" sounds so intense

StephenWeatherford: So if you don't opt in, then it will delete default resource groups?  That doesn't sound like a good idea...?

EricJizbaMSFT: I'm much more worried about nightly tests leaving extra resources around. I don't think people run these test locally very often and they shouldn't be running Azure tests (which you have to opt in to since they're long running) on a subscription they care about

StephenWeatherford: I disagree. I tried to run it yesterday to test out webpack. I certainly wasn't expecting it to delete anything in my subscription.

This should be opt in and set in the build parameters.  (Although really it shouldn't be creating anything in default resource groups in the first place - don't know the reason for that.)

extension.bundle.ts
```diff
@@ -18,3 +18,8 @@ export { activateInternal, deactivateInternal } from './src/extension';
 // The tests should import '../extension.bundle'. At design-time they live in tests/ and so will pick up this file (extension.bundle.ts).
 // At runtime the tests live in dist/tests and will therefore pick up the main webpack bundle at dist/extension.bundle.js.
 export { ext } from './src/extensionVariables';
+import * as constants from './src/constants';
+export { constants };
+export { AzureTreeDataProvider, DialogResponses, TestAzureAccount, TestUserInput } from 'vscode-azureextensionui';
```
EricJizbaMSFT: nit: I prefer `export * from 'vscode-azureextensionui';` for all of these.  Just makes it easier so you don't have to go back and individually add something every time. Plus this file isn't actually used in production so it has no effect there

Issue #644, opened 2018-10-12
Investigate "Cannot get /" immediately after deploy


fiveisprime:
It can take a few seconds (there's actually a message that claims it's up to 10 seconds) for the container to come up. The extension should either be smart about when the app comes up and show the notification after the app is verified available or just wait ~10 seconds before showing the notification.

fiveisprime:
I've dug into this and it appears that waiting a little longer to launch the site does result in a working site. I'll keep digging to determine a good debounce time.

nturinski:
New issue for the notification has been added #825

PR #798, opened 2019-01-04
Add a 10 second delay to notify user that a deployment completed

nturinski:
In reference to this issue: https://github.com/Microsoft/vscode-azureappservice/issues/644

When clicking the `View Website` button, if the user does it immediately after it is displayed, it will unavailable.  Waiting 10 seconds is to try to help mitigate that but is a less than ideal solution.  That being said, my example express app took way longer than 10 seconds to finish building.

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -17,6 +17,7 @@ import * as constants from '../constants';
 import { SiteTreeItem } from '../explorer/SiteTreeItem';
 import { WebAppTreeItem } from '../explorer/WebAppTreeItem';
 import { ext } from '../extensionVariables';
+import { delay } from '../util';
 import * as javaUtil from '../utils/javaUtils';
 import { isPathEqual, isSubpath } from '../utils/pathUtils';
 import * as workspaceUtil from '../utils/workspace';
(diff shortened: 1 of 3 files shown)


src/commands/deploy.ts
```diff
@@ -163,6 +164,9 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
     const browseWebsite: MessageItem = { title: 'Browse Website' };
     const streamLogs: MessageItem = { title: 'Stream Logs' };
 
+    // referring to this issue https://github.com/Microsoft/vscode-azureappservice/issues/644
+    // waiting 10 seconds is a temporary stop gap and should be changed
+    await delay(10000);
```
EricJizbaMSFT: If we're going to do a delay, I think it should at least happen while the "Deploying..." notification is still showing

nturinski:
As according to conversations, we will be implementing this delay for ONLY Basic SKU asps and include a message in the deployment complete notification that alerts the user the web app will require 10 seconds to restart.

nturinski:
Does adding a comment regarding the 10 second delay to the notification still make sense?  Or should I just close this in favor this PR https://github.com/Microsoft/vscode-azuretools/pull/362

PR #812, opened 2019-01-10
Add setting for preDeployTask

EricJizbaMSFT:
Based on this PR: https://github.com/Microsoft/vscode-azuretools/pull/369

For now it's just a setting. In the future would could consider prompting the user to make it more discoverable. (i.e. if they have a csproj file maybe we prompt them to say "Do you want to run `dotnet publish` before every deploy?") But I don't want to mess with that right now.

Fixes https://github.com/Microsoft/vscode-azureappservice/issues/389

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "icon": "resources/windows-azure-web-site.png",
     "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
     "engines": {
-        "vscode": "^1.23.0"
+        "vscode": "^1.24.0"
     },
     "repository": {
         "type": "git",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -851,6 +851,11 @@
                             }
                         }
                     }
+                },
+                "appService.preDeployTask": {
```
StephenWeatherford: I don't see this used anywhere

nturinski: This call: https://github.com/Microsoft/vscode-azureappservice/pull/812/files#diff-2916a9eae992830600ca3418c2c75814R160 is actually a function in the shared package that leverages `preDeployTask` (linked below)

https://github.com/Microsoft/vscode-azuretools/pull/369/files#diff-4cd3198960c945ebbf0dea96b508d228R13

PR #828, opened 2019-01-18
Enable webpack

StephenWeatherford:
Depends on Microsoft/vscode-azuretools#375


package.json
```diff
@@ -65,7 +65,7 @@
         "onCommand:appService.RevealConnectionInAppSettings",
         "onView:azureAppService"
     ],
-    "main": "./out/src/extension",
+    "main": "./entrypoint.js",
```
EricJizbaMSFT: I think we should discuss the entrypoint/main/extension naming. Here's my suggestions:
1. entrypoint.js -> main.js (to match the package.json name here)
1. Don't change the name of extension.ts (so we don't have to learn a new name)
1. Make webpack generate a `webpackedExtension.js` file instead of `extension.js` file

If this is the current file structure:
- dist
    - extension.js
- src
    - main.ts
- entrypoint.js
- extension.ts

I think my suggestion would result in a file structure like this:
- dist
    - webpackedExtension.js
- src
    - extension.ts
- main.js
- webpackedExtension.ts

EricJizbaMSFT:
Also @nturinski please bump to the next alpha before this gets merged as I don't want the webpack telemetry in with our current release

webpack.config.js
```diff
@@ -0,0 +1,69 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+//@ts-check
+
+// Full webpack documentation: [https://webpack.js.org/configuration/]().
+// Using webpack helps reduce the install- and startup-time of large extensions because instead of hundreds of files, a single file is produced.
+
+// How to fix "dynamic require", "Module not found", "the request of a dependency is an expression" etc. webpack errors:
```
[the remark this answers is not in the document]
StephenWeatherford: Yep, was planning that thx.

package.json
StephenWeatherford: How about this:

- dist
    - main.bundle.js
- src
    - extension.ts
- main.js
- main.bundle.ts

EricJizbaMSFT: I like the `.bundle` pattern, but shouldn't it be `extension.bundle` instead of `main.bundle`?
- dist
    - extension.bundle.js
- src
    - extension.ts
- main.js
- extension.bundle.ts

Webpack doesn't bundle `main.js` - it bundles `extension.ts`, right? Also it just feels weird to me that `main.js` would import `./dist/main.bundle.js` (Feels like that implies it's importing a bundle of itself)

StephenWeatherford: Works for me.

.vscodeignore
```diff
@@ -1,6 +1,7 @@
+# These files are excluded during npm pack
```
EricJizbaMSFT: npm pack?

gulpfile.ts
```diff
@@ -3,39 +3,41 @@
...
 import * as decompress from 'gulp-decompress';
 import * as download from 'gulp-download';
 import * as os from 'os';
 import * as path from 'path';
 
-function test() {
-    const env = process.env;
+const env = process.env;
+
+function webpack(mode: string): cp.ChildProcess {
```
EricJizbaMSFT: Reminding in case you forgot to leverage shared gulp tasks here

webpack.config.js
```diff
@@ -0,0 +1,38 @@
...
+// See https://github.com/Microsoft/vscode-azuretools/wiki/webpack for guidance
+
+const process = require('process');
+const webpack = require('webpack');
+const StringReplacePlugin = require("string-replace-webpack-plugin");
+const dev = require("vscode-azureextensiondev");
+
+let DEBUG_WEBPACK = !!process.env.DEBUG_WEBPACK;
+
+let config = dev.getDefaultWebpackConfig({
```
EricJizbaMSFT: The `vscode-azureappservice` npm package has icons just like the `vscode-azureextensionui` package does. Did a quick spot check and saw:
screen shot 2019-01-23 at 8 44 58 am

Don't care if you fix in this PR or file a bug to fix later

gulpfile.ts
StephenWeatherford: thx.

webpack.config.js
StephenWeatherford: https://github.com/Microsoft/vscode-azureappservice/issues/833, thanks

PR #834, opened 2019-01-23
Last PR fixes that got missed in merge

StephenWeatherford:


gulpfile.ts
```diff
@@ -3,72 +3,19 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// tslint:disable:no-console promise-function-async
-
 import * as cp from 'child_process';
```
EricJizbaMSFT: Should also be able to remove gulp-decompress, gulp-download, and glob from devDependencies in package.json

PR #854, opened 2019-02-05
Add `Toggle Slot Setting` context action menu

nturinski:
I'm not sure what's going on with this `package-lock.json`.   I tried doing this: https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285 and it totally changed my `package-lock` again, but maybe it'd be consistent if we all do it?


EricJizbaMSFT:
What exactly did you do? You have to completely clear your npm cache on your machine and remove the node_modules from this folder and then do an npm install

There shouldn't be that many 'http' ones

PR #875, opened 2019-02-25
Update app service and ui packages

EricJizbaMSFT:
Breaking change in app service was just a change to the signature of "deploy".

Also switch away from deprecated tslint extension. See here for more info: https://github.com/Microsoft/vscode-typescript-tslint-plugin#differences-with-the-vscode-tslint-extension

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,6 +1,6 @@
 {
     "recommendations": [
-        "eg2.tslint",
+        "ms-vscode.vscode-typescript-tslint-plugin",
         "ms-vscode.azure-account"
     ]
 }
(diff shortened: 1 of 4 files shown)


.vscode/extensions.json
```diff
@@ -1,6 +1,6 @@
 {
     "recommendations": [
-        "eg2.tslint",
+        "ms-vscode.vscode-typescript-tslint-plugin",
```
PrashanthCorp: Is this related to the other changes in this PR?

EricJizbaMSFT: ¯\\\_(ツ)\_/¯ It's all related in the sense that it's tedious process-related changes

PR #876, opened 2019-02-26
Remove remote debugging experiment flag + update version check

### LINKED ISSUE #802: Remove feature flag for remote debugging
End of January, the updated node start scripts will be udated for all 10.x.x images. Once that lands, the remote debugging option should always be available for apps that are Linux and Node >= 10.x.x

//cc @mrcrane

### LINKED ISSUE #806: Problems with node remote debugging telemetry
The image version was added to remote debugging telemetry in #710. After gathering some telemetry we are seeing 2 issues that should be addressed:

- In some cases the telemetry is not including any image version. We should investigate why and ensure that we can always determine the version being used.
- If the image version is a docker image, the name will include customer-specific container names. This information should not be included.

mrcrane:
Fixes #802 and fixes #806 

Remove the remote debugging experiment flag so that the menu items show for all users. At the same time, improve the check for remote debugging support to only allow remote debugging for Node >= 8.11

Error message is:
'Azure Remote Debugging is currently only supported for Node.js version >= 8.11 on Linux.'

### DIFF
--- a/package.json
+++ b/package.json
@@ -367,14 +367,6 @@
                     "when": "never"
                 },
                 {
-                    "command": "appService.StartRemoteDebug",
-                    "when": "never"
-                },
-                {
-                    "command": "appService.DisableRemoteDebug",
-                    "when": "never"
-                },
-                {
                     "command": "appService.ScaleUp",
                     "when": "never"
                 },
--- a/package.json
+++ b/package.json
@@ -368,11 +368,11 @@
                 },
                 {
                     "command": "appService.StartRemoteDebug",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -368,11 +368,11 @@
                 },
                 {
                     "command": "appService.StartRemoteDebug",
-                    "when": "config.appService.enableRemoteDebugging == true"
+                    "when": "never"
```
EricJizbaMSFT: "never" means these won't show up in the command palette, but you already work with the command palette. Just remove these 8 lines

mrcrane: Got it, yeah that makes sense.

PR #888, opened 2019-03-06
Use the javaUtils exported in azuretools

jdneo:
Signed-off-by: Sheng Chen <sheche@microsoft.com>

Using the javaUtils exported in the `azuretools`: https://github.com/Microsoft/vscode-azuretools/pull/434

Will fix the issue that if the user clicks `Deploy to Web App...` from a Java Wildfly Web App, the extension should ask the user to select a war file instead of a folder.

@EricJizbaMSFT Not sure if the library version should be updated or not...


jdneo:
Thank you @nturinski

The version has been updated.

Issue #726, opened 2018-11-21
Add "View commit in Github" to deployment item


EricJizbaMSFT:
[Feature request] If I right click on a deployment:
<img width="307" alt="screen shot 2018-11-21 at 1 25 35 pm" src="https://user-images.githubusercontent.com/11282622/48864053-0681a280-ed91-11e8-8585-f72b415fd1cc.png">
I want a command that lets me view that commit in GitHub.

nturinski:
Command is implemented in azuretools, but need to register command in extension.

PR #907, opened 2019-03-13
View commit in GitHub

nturinski:
Updated the context values so that only Github deployments can use `View in Github`.  I thought it was super confusing that you could use the command and it just wouldn't work.

Fixes https://github.com/Microsoft/vscode-azureappservice/issues/726


src/explorer/SiteTreeItem.ts
```diff
@@ -104,6 +105,9 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
             case WebJobsTreeItem.contextValue:
                 return this._webJobsNode;
             default:
+                if (typeof expectedContextValue === 'string' && DeploymentTreeItem.contextValue.test(expectedContextValue)) {
```
EricJizbaMSFT: Can you explain what you're doing here? I'm a little confused by this

nturinski: For the TreeItemPicker, `SiteTreeItem` wouldn't recognize that `/deployments\//` should return `this.deploymentsNode` since it is only looking for deep equals to the expectedContextValue.

The static `DeploymentTreeItem.contextValue` is a RegExp, but the treeItems in the tree have string context values (ie `deployments/github`).  So in order to recognize that `contextValue` as a deploymentNode, it checks if the `expectedContextValue` matches with the `DeploymentTreeItem` regexp.

nturinski: So `expectedContext` is whatever I pass in to `showTreeItemPicker`.  In this case, it's 'deployment/github' because **only** want deployments that have a github repo associated to it.  

image

Therefore, without that match statement, it looks like this:

image

EricJizbaMSFT: Okay makes sense now

PR #920, opened 2019-04-02
Don't use webpack for local debug

EricJizbaMSFT:
Same as https://github.com/Microsoft/vscode-azurefunctions/pull/1151

Also note for both repos, icons only work if you launch with webpack haha:
Screen Shot 2019-04-02 at 11 23 26 AM

Do we care about that? Doesn't seem like a blocker for this PR - but might be worth fixing eventually.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,7 +12,26 @@
             ],
             "stopOnEntry": false,
             "sourceMaps": true,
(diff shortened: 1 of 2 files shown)


StephenWeatherford:
> Do we care about that? Doesn't seem like a blocker for this PR - but might be worth fixing eventually.

I think it would be good to fix, but agreed not a blocker.

PR #923, opened 2019-04-03
Remote SSH for Linux Containers

### LINKED ISSUE #493: Suggestion: Add ability to ssh into app service console
It would be cool to be able to open up a console for the App Service, directly from within VS Code, much like you can open a Cloud Console (see https://www.michaelcrump.net/azure-tips-and-tricks49/).  Not sure if it is possible, but it would be helpful to see logs, run npm install, builds, etc.

nturinski:
Relies on this PR: https://github.com/Microsoft/vscode-azuretools/pull/466
Fixes #493 

Looks like this on the context menu:

image

We should probably start moving some of these context actions to other menus, such as Deploy/Configure Deployment Source.


src/commands/remoteDebug/startRemoteDebug.ts
```diff
@@ -56,7 +56,7 @@ async function startRemoteDebugInternal(actionContext: IActionContext, node?: Si
         await callWithTelemetryAndErrorHandling('appService.remoteDebugStartProxy', async function (this: IActionContext): Promise<void> {
             this.suppressErrorDisplay = true;
             this.rethrowError = true;
-            await tunnelProxy.startProxy();
+            await tunnelProxy.startProxy(49494);
```
EricJizbaMSFT: Why 49494?

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+
+async function startRemoteSshInternal(node: SiteTreeItem): Promise<void> {
+    const siteClient: SiteClient = node.root.client;
+    const siteConfig: SiteConfigResource = await siteClient.getSiteConfig();
+    const oldSetting: boolean = <boolean>siteConfig.remoteDebuggingEnabled;
+    // should always be an unbound port
+    const portNumber: number = await portfinder.getPortPromise();
+
+    await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification }, async (progress: vscode.Progress<{}>): Promise<void> => {
+        if (!siteConfig.linuxFxVersion) {
```
EricJizbaMSFT: use `SiteClient.isLinux`

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+            throw new Error('Azure Remote SSH is only supported for Linux web apps.');
+        }
+
+        remoteDebug.reportMessage('Checking app settings...', progress);
+
+        // remote debugging has to be disabled in order to tunnel to the 2222 port
+        await remoteDebug.setRemoteDebug(false, undefined /*skips confirmation*/, undefined, siteClient, siteConfig);
+
+        remoteDebug.reportMessage('Starting tunnel proxy...', progress);
```
EricJizbaMSFT: `Starting tunnel proxy...` is probably meaningless for most users. Can we get a better message?

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+
+        const publishCredential: User = await siteClient.getWebAppPublishCredential();
+        const tunnelProxy: TunnelProxy = new TunnelProxy(portNumber, siteClient, publishCredential);
+        await callWithTelemetryAndErrorHandling('appService.remoteSshStartProxy', async function (this: IActionContext): Promise<void> {
```
EricJizbaMSFT: Why is this a separate telemetry event?

package.json
```diff
@@ -220,6 +222,16 @@
                 "title": "Disable Remote Debugging",
                 "category": "Azure App Service"
             },
+            {
+                "command": "appService.StartRemoteSsh",
+                "title": "Start Remote SSH",
```
EricJizbaMSFT: I feel like "SSH" already implies "remote"?

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        remoteDebug.reportMessage('Checking app settings...', progress);
+
+        // remote debugging has to be disabled in order to tunnel to the 2222 port
+        await remoteDebug.setRemoteDebug(false, undefined /*skips confirmation*/, undefined, siteClient, siteConfig);
+
+        remoteDebug.reportMessage('Starting tunnel proxy...', progress);
+
+        const publishCredential: User = await siteClient.getWebAppPublishCredential();
```
EricJizbaMSFT: Should this code use `callWithMaskHandling`?

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+        terminal.sendText(sshCommand, true);
+        await delay(3000);
+        terminal.sendText('Docker!', true);
+        terminal.show();
+        ext.context.subscriptions.push(terminal);
+
+        vscode.window.onDidCloseTerminal(async (e: vscode.Terminal) => {
+            if (e.processId === terminal.processId) {
+                // clean up if the SSH task ends
+                if (tunnelProxy !== undefined) {
```
EricJizbaMSFT: How could this be undefined?

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+                    tunnelProxy.dispose();
+                }
+                remoteSsh.set(node.root.client.fullName, false);
+                ext.outputChannel.appendLine(`Azure Remote SSH for "${node.root.client.fullName}" has disconnected.`);
+                await remoteDebug.setRemoteDebug(oldSetting, undefined/*skips confirmation*/, undefined, siteClient, siteConfig);
```
EricJizbaMSFT: I don't like that we're disabling/enabling remote debugging without telling the user. I'm also skeptical that we'll always have time to set this back to the old value before VS Code closes. Instead, I would prefer to prompt the user with a message saying something like "You must disable remote debugging in order to ssh" and let them click a "Disable" button to continue. Then don't even try to set it back to the old value when this closes.

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+export async function stopRemoteSsh(node?: SiteTreeItem): Promise<void> {
+    if (!node) {
+        node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+    }
+
+    if (!remoteSsh.get(node.root.client.fullName)) {
+        throw new Error(`Azure Remote SSH is not currently running for "${node.root.client.fullName}".`);
+    }
+    const sshTerminalName: string = `${node.root.client.fullName} - Remote SSH`;
+    for (const terminal of vscode.window.terminals) {
```
EricJizbaMSFT: Your `remoteSsh` map above should somehow contain the terminal so that you don't have to search for it by name

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+                }
+                remoteSsh.set(node.root.client.fullName, false);
+                ext.outputChannel.appendLine(`Azure Remote SSH for "${node.root.client.fullName}" has disconnected.`);
+                await remoteDebug.setRemoteDebug(oldSetting, undefined/*skips confirmation*/, undefined, siteClient, siteConfig);
+            }
+        });
+    }
+}
+
+export async function stopRemoteSsh(node?: SiteTreeItem): Promise<void> {
```
EricJizbaMSFT: Can we clean up the naming for this file? 
1. It's in the remoteDebug folder, but it's a separate command
1. It's named `startRemoteSsh`, but it also has `stopRemoteSsh `

src/commands/remoteDebug/startRemoteSsh.ts
```diff
@@ -0,0 +1,110 @@
...
+            await tunnelProxy.startProxy(2222);
+            await connectToTunnelProxy(tunnelProxy);
+        });
+    });
+
+    async function connectToTunnelProxy(tunnelProxy: TunnelProxy): Promise<void> {
+        const sshTerminalName: string = `${node.root.client.fullName} - Remote SSH`;
+        // -o StrictHostKeyChecking=no doesn't prompt for adding to hosts
+        // -o "UserKnownHostsFile /dev/null" doesn't add host to known_user file
+        // -o "LogLevel ERROR" doesn't display Warning: Permanently added 'hostname,ip' (RSA) to the list of known hosts.
```
EricJizbaMSFT: Do we need `-o "LogLevel ERROR"` if you're not adding these to known hosts file?

nturinski: Yeah, due to the warning that says we added a RSA key to the known_hosts file (even though we actually didn't)

nturinski:
https://github.com/Microsoft/vscode-azureappservice/issues/926

src/commands/remoteDebug/startRemoteDebug.ts
nturinski: It's the port that's used for remote debugging (as far as I can tell) This seems dependent on kudu (we never set this port for remote debugging, it just happened to be the port that was opened)

src/commands/remoteDebug/startRemoteSsh.ts
nturinski: Sure. Any suggestions?

Something simple like `Starting SSH...`?

package.json
nturinski: Just Start SSH doesn't feel descriptive enough to me.  Start SSH into Web App?

src/commands/remoteDebug/startRemoteSsh.ts
nturinski: I dunno, might be able to find a way to close the proxy in the shell terminal.  But Mr. Crane had it that way so I figured I'd follow his pattern.

src/commands/remoteDebug/startRemoteSsh.ts
nturinski: Should I move `stopRemoteSsh` or change the name of the file?  Make it a class instead?

package.json
EricJizbaMSFT: I mean... I'm already right clicking on the web app. We don't say "Start Web App" or "Delete Web App"

src/commands/remoteDebug/startRemoteSsh.ts
EricJizbaMSFT: Yeah that's better

src/commands/remoteDebug/startRemoteSsh.ts
EricJizbaMSFT: Any of those sound fine

src/commands/remoteDebug/startRemoteSsh.ts
EricJizbaMSFT: Could this hide other important warnings?

nturinski: I'm not sure.  I would assume once you run the SSH command, it wouldn't affect anything else since you're technically in the other container at that point.

src/commands/remoteDebug/startRemoteSsh.ts
nturinski: I think we're okay here. We're just using the publishCredential to make a request to Azure.

src/commands/remoteDebug/remoteDebugCommon.ts
```diff
@@ -57,19 +57,25 @@ function isNodeVersionSupported(nodeVersion: string): boolean {
...
                 await siteClient.updateConfiguration(siteConfig);
             });
-            reportMessage('Updating site configuration done...', progress);
+
+            if (progress) {
+                reportMessage('Updating site configuration done...', progress);
+            }
+
         } else if (result === DialogResponses.learnMore) {
             await openUrl('https://aka.ms/appsvc-remotedebug');
```
EricJizbaMSFT: This learn more link won't make sense in your ssh scenario. I'm fine with creating your own learn more link or just removing the learn more button in this case.

It might make your life easier to use the shared `learnMoreLink` support I added a little while ago. (Where you just pass in learnMoreLink in the options to `ext.ui.showWarningMessage` and it will handle adding the learn more button if that link is defined). Like [here](https://github.com/Microsoft/vscode-azurefunctions/blob/4810e0ebaec0b1ad78d820ad0e519d3227019dd2/src/commands/createNewProject/verifyIsProject.ts#L60)

src/commands/ssh/startSsh.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+async function startSshInternal(node: SiteTreeItem): Promise<void> {
+    const siteClient: SiteClient = node.root.client;
+    const siteConfig: SiteConfigResource = await siteClient.getSiteConfig();
+
+    // should always be an unbound port
+    const localHostPortNumber: number = await portfinder.getPortPromise();
+    const sshPortNumber: number = 2222;
+    const confirmDisableMessage: string = 'The app configuration will be updated to disable remote debugging and restarted. Would you like to continue?';
```
EricJizbaMSFT: How about `You must disable remote debugging in order to ssh. This will restart the app.` with a `Disable` button

I just worry the existing message makes it seem like I accidentally clicked "Disable remote debugging" instead of "Start SSH"

src/commands/ssh/startSsh.ts
```diff
@@ -0,0 +1,99 @@
...
+        await connectToTunnelProxy(tunnelProxy);
+    });
+
+    async function connectToTunnelProxy(tunnelProxy: TunnelProxy): Promise<void> {
+        const sshTerminalName: string = `${node.root.client.fullName} - SSH`;
+        // -o StrictHostKeyChecking=no doesn't prompt for adding to hosts
+        // -o "UserKnownHostsFile /dev/null" doesn't add host to known_user file
+        // -o "LogLevel ERROR" doesn't display Warning: Permanently added 'hostname,ip' (RSA) to the list of known hosts.
+        const sshCommand: string = `ssh -c aes256-cbc -o StrictHostKeyChecking=no -o "UserKnownHostsFile /dev/null" -o "LogLevel ERROR" root@127.0.0.1 -p ${localHostPortNumber}`;
+        const terminal: vscode.Terminal = vscode.window.createTerminal(sshTerminalName);
```
EricJizbaMSFT: Have you tried the following?:
1. Start SSH
1. Stop SSH
1. Start SSH

I'm just curious if `vscode.window.createTerminal` works if the terminal with that name already exists

src/commands/ssh/startSsh.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        // because the container needs time to respond, there needs to be a delay between connecting and entering password
+        terminal.sendText(sshCommand, true);
+        await delay(3000);
+        terminal.sendText('Docker!', true);
```
EricJizbaMSFT: Can you add a comment describing what this password is for? Specifically that it's the password to the container _after_ you ssh rather than for ssh-ing itself

src/commands/ssh/startSsh.ts
```diff
@@ -0,0 +1,99 @@
...
+        terminal.show();
+        ext.context.subscriptions.push(terminal);
+        sshSessionsMap.set(node.root.client.fullName, { running: true, terminal: terminal });
+
+        vscode.window.onDidCloseTerminal(async (e: vscode.Terminal) => {
```
EricJizbaMSFT: I believe this returns a disposable. You should handle that by disposing the first time `if (e.processId === terminal.processId) {` is true. (Otherwise this event will keep being fired for every terminal that's closed even after you already cleaned up the ssh terminal)

src/commands/ssh/startSsh.ts
```diff
@@ -0,0 +1,99 @@
...
+export const sshSessionsMap: Map<string, sshTerminal> = new Map();
+
+export async function startSsh(node?: SiteTreeItem): Promise<void> {
+    if (!node) {
+        node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+    }
+
+    const currentSshTerminal: sshTerminal | undefined = sshSessionsMap.get(node.root.client.fullName);
+    if (currentSshTerminal && currentSshTerminal.running) {
+        throw new Error(`Azure SSH is currently starting or already started for "${node.root.client.fullName}".`);
```
EricJizbaMSFT: I'm not sure your map works in all cases. For example, if I send `ctrl+c` to the ssh terminal - wouldn't that exit me out of the ssh? But I believe your map would say I'm still connected and wouldn't let me connect again.

package.json
```diff
@@ -220,6 +222,16 @@
...
                 "category": "Azure App Service"
             },
+            {
+                "command": "appService.StartSsh",
+                "title": "Start SSH",
+                "category": "Azure App Service"
+            },
+            {
+                "command": "appService.StopSsh",
+                "title": "Stop SSH",
```
EricJizbaMSFT: I'm not sure how I feel about 'Stop SSH'. It almost feels more intuitive to delete the terminal in VS Code or to just send `ctrl+c` to the terminal. Thoughts? It would help reduce the number of context menu items

src/extension.ts
```diff
@@ -300,6 +302,8 @@ export async function activateInternal(
 
         registerCommand('appService.StartRemoteDebug', async function (this: IActionContext, node?: SiteTreeItem): Promise<void> { await startRemoteDebug(this, node); });
         registerCommand('appService.DisableRemoteDebug', async function (this: IActionContext, node?: SiteTreeItem): Promise<void> { await disableRemoteDebug(this, node); });
+        registerCommand('appService.StartSsh', async (node?: SiteTreeItem): Promise<void> => { await startSsh(node); });
```
EricJizbaMSFT: Just do `registerCommand('appService.StartSsh', startSsh);`

src/commands/ssh/startSsh.ts
nturinski: Yeah, I have.  Stop SSH will dispose of the terminal.  If you try to Stop SSH before it's actually connected, it's ignored rather than canceling the SSHing.

src/commands/ssh/startSsh.ts
nturinski: Yeah, this is true.  The proxy is still open in this case so you could reconnect manually, but I'll try to figure out how to handle that scenario.

EricJizbaMSFT: If it gets complicated, feel free to file an issue and handle after this PR

package.json
nturinski: Yeah, the problem is the only way to end SSH is to actually trash the terminal (unless I use Tasks).  I don't feel like that garbage icon is super intuitive to stop SSH.  If you Ctrl+C or type `exit`, only the SSH connection is killed, but as the terminal remains, there is no way for us to detect it.

src/commands/remoteDebug/startRemoteDebug.ts
```diff
@@ -52,7 +52,8 @@ async function startRemoteDebugInternal(actionContext: IActionContext, node?: Si
         remoteDebug.reportMessage('Starting tunnel proxy...', progress);
 
         const publishCredential: User = await siteClient.getWebAppPublishCredential();
-        const tunnelProxy: TunnelProxy = new TunnelProxy(portNumber, siteClient, publishCredential);
+        const remoteDebuggingPort: number = 49494;
```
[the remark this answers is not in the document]
nturinski: I'm waiting for final confirmation, but yeah, apparently remote debug mode is just if the port is different from 2222.  I'll update the appservice package

src/commands/startSsh.ts
```diff
@@ -0,0 +1,114 @@
...
+            await connectToTunnelProxy(node, currentSshTerminal.tunnel, currentSshTerminal.localPort);
+            return;
+        }
+    }
+
+    try {
+        sshSessionsMap.set(node.root.client.fullName, { starting: true, terminal: undefined, tunnel: undefined, localPort: undefined });
+        await startSshInternal(node);
+    } catch (error) {
+        sshSessionsMap.set(node.root.client.fullName, { starting: false, terminal: undefined, tunnel: undefined, localPort: undefined });
```
EricJizbaMSFT: I think this would work:
```
templatesMap.delete(node.root.client.fullName)
```

Same thing below.

src/commands/startSsh.ts
```diff
@@ -0,0 +1,114 @@
...
+    // -o "LogLevel ERROR" doesn't display Warning: Permanently added 'hostname,ip' (RSA) to the list of known hosts.
+    const sshCommand: string = `ssh -c aes256-cbc -o StrictHostKeyChecking=no -o "UserKnownHostsFile /dev/null" -o "LogLevel ERROR" root@127.0.0.1 -p ${port}`;
+
+    // if this terminal already exists, just reuse it otherwise create a new terminal.
+    // tslint:disable-next-line:strict-boolean-expressions
+    const terminal: vscode.Terminal = vscode.window.terminals.find((activeTerminal: vscode.Terminal) => { return activeTerminal.name === sshTerminalName; }) || vscode.window.createTerminal(sshTerminalName);
+
+    // because the container needs time to respond, there needs to be a delay between connecting and entering password
+    terminal.sendText(sshCommand, true);
+    await delay(3000);
```
EricJizbaMSFT: So this is a workaround, right? That'll be fixed when we switch to tasks? If so - please file an issue and add a comment in the code linking to that issue

PR #928, opened 2019-04-08
ERASE ALL PICTURES OF OPN

nturinski:


src/explorer/SiteTreeItem.ts
```diff
@@ -74,9 +74,9 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         return this.root.client.id;
     }
 
-    public browse(): void {
+    public async browse(): Promise<void> {
         // tslint:disable-next-line:no-unsafe-any
```
EricJizbaMSFT: nit: remove the tslint disable line in all these cases since it no longer applies

Issue #891, opened 2019-03-06
Expanding WebJobs node shows an error 'Failed with code:404'


v-xuanzh:
**OS:** All
**Build Version**: [20190306.3](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=3743)
**Regression:** Not a regression

**Steps to reproduce:**
1. Create a new web app with any OS.
2. Try to expand the created web app -> 'WebJobs' node.
3. Check the result.

**Expect:**
There is no error on 'WebJobs' node.

**Actual:**
There shows an error 'Failed with code:404' on 'WebJobs' node.
![image](https://user-images.githubusercontent.com/34729022/53872979-44a77880-3fb4-11e9-89a9-bc1b62998ee0.png)

nturinski:
Will look at using the ARM api for using WebJobs

PR #934, opened 2019-04-11
Add check for isLinux in Webjobs

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/891

Turns out the error is coming from the fact that Linux doesn't support WebJobs.  So the url returns a 404 error.

I didn't think it was worth to create a WebJobsNATreeItem (just yet) as WebJobs doesn't really have any features at this point and WebJobs not being available to Linux isn't an actionable item.  If we do decide to expand on WebJobs and would have to disable more commands, I'd gladly make a separate treeItem.

image


src/explorer/WebJobsTreeItem.ts
```diff
@@ -19,19 +19,27 @@ export class WebJobsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
     }
 
     public hasMoreChildrenImpl(): boolean {
         return false;
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzureTreeItem<ISiteTreeRoot>[]> {
-        const jobList: webJob[] = <webJob[]>await this.root.client.kudu.jobs.listAllJobs();
+        if (this.root.client.isLinux) {
+            return [new GenericTreeItem<ISiteTreeRoot>(this, { label: 'WebJobs are not available for Linux Apps.', contextValue: 'webJobNA' })];
```
EricJizbaMSFT: Can you add a link to docs for this? Just want to make sure we're not blocking something that might work a week or two from now.

Actually it might be better to do a try/catch and add this logic in the catch block so that we will work with webJobs if/when that ever happens

nturinski: Very smort

src/explorer/WebJobsTreeItem.ts
```diff
@@ -19,18 +19,33 @@ export class WebJobsTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+        let jobList: webJob[];
+        try {
+            jobList = <webJob[]>await this.root.client.listWebJobs();
+        } catch (err) {
+            if (this.root.client.isLinux) {
+                // Can't find actual documentation on this, but the portal claims it and this feedback suggests it's not planned https://aka.ms/AA4q5gi
+                return [new GenericTreeItem<ISiteTreeRoot>(this, { label: 'WebJobs are not available for Linux Apps.', contextValue: 'webJobNA' })];
+            }
+
+            throw parseError(err);
```
EricJizbaMSFT: nit: just `throw err;`

PR #939, opened 2019-04-16
Add intellinav upload to yml file

nturinski:


.azure-pipelines/windows/intellinav.yml
```diff
@@ -0,0 +1,8 @@
+steps:
```
EricJizbaMSFT: How often should this be run? And what exactly is the use case?

For example - will it work from forks? If not - take a look at this to disable for forks:
https://github.com/Microsoft/vscode-azureappservice/blob/38105bbcf6c6ca4f466c8451678450072274ff8c/.azure-pipelines/common/lint.yml#L10

Also - do we need this running in nightly or master builds? I thought it was only for PRs correct? Maybe we can save build time by filtering stuff out here.

nturinski: Well, in order for it to work, it has to be updated whenever the code changes.  I guess we can do it only in master, but we wouldn't be able to guarantee that the function definitions will be accurate.  But since every change is compared against master anyway, that'd probably be good enough?

arjun27: Just to add some context, in case it helps: by default, the build task only runs when the build is associated to a PR. It gets skipped automatically for builds running on master, or in any other context. Once the task runs, you will be able to use rich code navigation on PRs (editor-level hover, references, definitions).

EricJizbaMSFT: @arjun27 yes that's helpful - thanks!

Issue #946, opened 2019-04-17
With no access to the plan, you get a very non-descriptive error of why create app failed


nturinski:
Action: appService.CreateWebAppError 
type: 403Error Message: 
Failed with code 403.
Version: 0.13.1-alphaOS: win32Product: 
Visual Studio Code - Insiders
Product Version: 1.34.0-insiderLanguage:

Issue #769, opened 2018-12-06
Allow web app deploy even if user doesn't have access to plan


vibhort90:
&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: appService.Deploy
Error type: AuthorizationFailed
Error Message: The client 'Vibhor.Tyagi@markeldev.net' with object id '90bd40bf-2883-4656-9cf7-5cb13a7c8ef6' does not have authorization to perform action 'Microsoft.Web/serverfarms/read' over scope '/subscriptions/c9a62e1b-3391-4f41-a640-8d3601f57221/resourceGroups/MarkelDev/providers/Microsoft.Web/serverfarms/markeldevappserviceplan'.

Version: 0.10.0
OS: win32

nturinski:
Hi @vibhort90,

It looks like you do not have permission to deploy to the web app.  Is this web app under your subscription?  If not, you may need to ask the administration to give you write privileges.

kongsted:
Hi,
I'm having the same problem as described by @vibhort90 .
I'm trying to grant minimal access to the users who are responsible for the web app (means that they shouldn't have permissions on the subscription or the resource group ( if not needed for specific reason).
In my test-case, my normal user-account "only" having contribute access for the WebApp, and then I can't list the content of the WebApp. If I grant read-access for the subscription or the resource group, I can view the content of the WebApp in VS Code.
//Kongsted

EricJizbaMSFT:
@kongsted thanks for the info. If you want to restrict even further, I believe you only need to give access to the web app and the app service plan (aka not the subscription or resource group). We'll look into a fix in our extension to see if we can get that down to just the web app

nturinski:
@vibhort90 @kongsted

This should be fixed with the newest version that should be released at the beginning of May.  

I'll reach out when we release, but if you'd like to pick up these changes now, you can find a build of our vsix here: https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=5145

![image](https://user-images.githubusercontent.com/5290572/56396895-f2ce5100-61f5-11e9-83ed-dfd50e7378f0.png)

PR #947, opened 2019-04-17
Ignore plan if no access for creating treeItem

nturinski:
Related https://github.com/Microsoft/vscode-azureappservice/issues/769

Of course, this leads to more issues such as these: https://github.com/Microsoft/vscode-azureappservice/issues/946

### DIFF
--- a/src/explorer/WebAppTreeItem.ts
+++ b/src/explorer/WebAppTreeItem.ts
@@ -30,8 +30,16 @@ export class WebAppTreeItem extends SiteTreeItem {
     }
 
     public async loadMoreChildrenImpl(clearCache: boolean): Promise<AzureTreeItem<ISiteTreeRoot>[]> {
-        const asp: AppServicePlan | undefined = await this.root.client.getAppServicePlan();
-        const tier: string | undefined = asp && asp.sku && asp.sku.tier;
+        let tier: string | undefined
+        let asp: AppServicePlan | undefined;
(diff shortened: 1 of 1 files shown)


src/explorer/WebAppTreeItem.ts
```diff
@@ -30,8 +30,16 @@ export class WebAppTreeItem extends SiteTreeItem {
...
-        const asp: AppServicePlan | undefined = await this.root.client.getAppServicePlan();
-        const tier: string | undefined = asp && asp.sku && asp.sku.tier;
+        let tier: string | undefined
+        let asp: AppServicePlan | undefined;
+        try {
+            asp = await this.root.client.getAppServicePlan();
+            tier = asp && asp.sku && asp.sku.tier;
+        } catch (err) {
+            // ignore this error, we don't want to block users for deployment slots
+            tier = 'ignored';
```
EricJizbaMSFT: I would call this 'unknown'

nturinski:
Nah, this only fixes this specific error.

image

The other changes have to be made in the `deploy` command in the `appservice` package.

nturinski:
Are we okay with this behavior?  This command is used very much and I think the error makes sense, but wanted your opinion @EricJizbaMSFT 

deployscript_failure

EricJizbaMSFT:
Yeah that error is perfectly reasonable IMO

PR #953, opened 2019-04-23
Nat/appservice

nturinski:


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azureappservice",
-    "version": "0.13.2-alpha",
+    "version": "0.14.0",
```
EricJizbaMSFT: nit: Can we leave this as an alpha version for now? We're not releasing for over a week, so I don't want to be collecting that all that testing telemetry in `0.14.0`

Issue #780, opened 2018-12-13
Move smart defaults for Web App from azuretools to here


nturinski:
"I'm beginning to think setWizardContextDefaults should be directly in the appservice repo rather than in the shared package. (And then you'd pass that stuff in as a part of IAppCreateOptions). Thoughts? Doesn't have to be done in this PR."

PR #957, opened 2019-04-25
Refactor CreateWebApp from appservice and do it all in extension

nturinski:
Relies on https://github.com/Microsoft/vscode-azuretools/pull/489
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/780
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/979


src/explorer/WebAppProvider.ts
```diff
@@ -59,15 +57,17 @@ export class WebAppProvider extends SubscriptionTreeItem {
...
+            if (newSite === undefined) {
+                throw new UserCancelledError();
+            } else {
+                const siteClient: SiteClient = new SiteClient(newSite, this.root);
+                return new WebAppTreeItem(this, siteClient);
+            }
         } else {
-            const siteClient: SiteClient = new SiteClient(newSite, this.root);
-            return new WebAppTreeItem(this, siteClient);
+            throw new Error('Missing object "actionContext"');
```
nturinski: I wasn't really sure what to do in this scenario (but don't expect to run into it either)

EricJizbaMSFT: This is actually pretty common: https://github.com/Microsoft/vscode-azuretools/issues/120

Just create a fake one: https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/tree/FunctionAppProvider.ts#L73

src/explorer/WebAppProvider.ts
```diff
@@ -3,15 +3,14 @@
...
 
+import { Location } from 'azure-arm-resource/lib/subscription/models';
 import { WebSiteManagementClient } from 'azure-arm-website';
 import { Site, WebAppCollection } from 'azure-arm-website/lib/models';
-import { workspace, WorkspaceConfiguration } from 'vscode';
-import { createWebApp, SiteClient } from 'vscode-azureappservice';
-import { AzureTreeItem, createAzureClient, createTreeItemsWithErrorHandling, IActionContext, parseError, SubscriptionTreeItem, UserCancelledError } from 'vscode-azureextensionui';
-import { configurationSettings, extensionPrefix } from '../constants';
+import * as appservice from 'vscode-azureappservice';
+import * as extensionui from 'vscode-azureextensionui';
```
EricJizbaMSFT: Why'd you have to do these `appservice ` and `extensionui ` things? I try to avoid it because it makes the code harder to read and the imports are generally handled automatically anyways

src/commands/setAppCreateOptionsDefault.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import * as fse from 'fs-extra';
+import * as path from 'path';
+import { Uri, workspace, WorkspaceConfiguration } from 'vscode';
+import { IAppCreateOptions, javaUtils, LinuxRuntimes, WebsiteOS } from 'vscode-azureappservice';
+import { IActionContext } from 'vscode-azureextensionui';
+import { configurationSettings, extensionPrefix } from '../constants';
+
+export async function setAppCreateOptionsDefault(actionContext: IActionContext): Promise<IAppCreateOptions> {
+    const isJavaProject: boolean = await javaUtils.isJavaProject();
```
EricJizbaMSFT: As mentioned in the shared tools PR, I think you should delete `IAppCreateOptions`. Since you moved everything over here, you can just use the wizardContext directly.

src/commands/createWebApp.ts
```diff
@@ -0,0 +1,42 @@
...
+const noButton: MessageItem = { title: 'No', isCloseAffordance: true };
+
+export async function createWebApp(actionContext: IActionContext, node?: AzureParentTreeItem | undefined): Promise<void> {
+    if (!node) {
+        node = <AzureParentTreeItem>await ext.tree.showTreeItemPicker(SubscriptionTreeItem.contextValue);
+    }
+
+    const createOptions: IAppCreateOptions = await setAppCreateOptionsDefault(actionContext);
+
+    const createdApp = <WebAppTreeItem>await node.createChild(createOptions);
```
EricJizbaMSFT: createChild has known problems passing in createOptions (mostly with the node picker). It seems like you could do `setAppCreateOptionsDefault` directly within `createChildImpl`, which would be ideal

src/extension.ts
```diff
@@ -182,26 +182,7 @@ export async function activateInternal(
             await node.deleteTreeItem();
         });
         registerCommand('appService.CreateWebApp', async function (this: IActionContext, node?: AzureParentTreeItem): Promise<void> {
-            if (!node) {
```
EricJizbaMSFT: Removing lines from this file makes me happy 😊

src/explorer/WebAppProvider.ts
nturinski: I think there were two reasons:

1. appservice because I can't have a shadowed variable name `editScmType`.  I'm fine with calling it something else like `configureDeploymentSource` since that's what the UI says anyway

1. extensionui because there were so many things imported that it looked terrible and felt like I might as well just import the whole thing.

src/commands/createWebApp.ts
```diff
@@ -0,0 +1,64 @@
+
```
EricJizbaMSFT: nit: remove this blank line

src/commands/createWebApp.ts
```diff
@@ -0,0 +1,64 @@
...
+        throw error;
+    }
+
+    if (createdApp === undefined) {
+        throw new UserCancelledError();
+    }
+
+    createdApp.root.client.getSiteConfig().then(
+        (createdAppConfig: SiteConfigResource) => {
+            actionContext.properties.linuxFxVersion = createdAppConfig.linuxFxVersion ? createdAppConfig.linuxFxVersion : 'undefined';
```
EricJizbaMSFT: If you want to track `linuxFxVersion`, it should be a part of `createChildImpl` so that it's also tracked if they create a web app in the deploy scenario

src/commands/createWebApp.ts
```diff
@@ -0,0 +1,64 @@
...
+            actionContext.properties.createdFromDeploy = 'false';
```
EricJizbaMSFT: This property is unnecessary. We already know it's not created from deploy since the command id is `appService.CreateWebApp`

src/commands/createWebApp.ts
```diff
@@ -0,0 +1,64 @@
...
+            ext.ui.showWarningMessage(message, btn).then(async result => {
+                if (result === btn) {
+                    await workspaceConfig.update('advancedCreation', true, ConfigurationTarget.Global);
+                }
+            });
+        }
+        throw error;
+    }
+
+    if (createdApp === undefined) {
```
EricJizbaMSFT: I don't think this can ever happen

src/commands/setAppWizardContextDefault.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import * as fse from 'fs-extra';
+import * as path from 'path';
+import { Uri, workspace, WorkspaceConfiguration } from 'vscode';
+import { IAppServiceWizardContext, javaUtils, LinuxRuntimes, WebsiteOS } from 'vscode-azureappservice';
+import { LocationListStep } from 'vscode-azureextensionui';
+import { configurationSettings, extensionPrefix } from '../constants';
+
+export async function setAppWizardContextDefault(wizardContext: IAppServiceWizardContext): Promise<void> {
+    const isJavaProject: boolean = await javaUtils.isJavaProject();
```
EricJizbaMSFT: `javaUtils` should also be moved from the `vscode-azureappservice` package to this repo

src/commands/setAppWizardContextDefault.ts
```diff
@@ -0,0 +1,66 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
+import * as path from 'path';
+import { Uri, workspace, WorkspaceConfiguration } from 'vscode';
+import { IAppServiceWizardContext, javaUtils, LinuxRuntimes, WebsiteOS } from 'vscode-azureappservice';
+import { LocationListStep } from 'vscode-azureextensionui';
+import { configurationSettings, extensionPrefix } from '../constants';
+
+export async function setAppWizardContextDefault(wizardContext: IAppServiceWizardContext): Promise<void> {
```
EricJizbaMSFT: This isn't really a command so I'm not sure it fits in this folder. Can we move this file closer to where it's actually used?

src/explorer/WebAppProvider.ts
```diff
@@ -60,31 +62,72 @@ export class WebAppProvider extends SubscriptionTreeItem {
...
+        if (wizardContext.newSiteOS !== undefined) {
+            SiteOSStep.setLocationsTask(wizardContext);
         }
+
+        const title: string = 'Create new web app';
+        const wizard: AzureWizard<IAppServiceWizardContext> = new AzureWizard(wizardContext, { promptSteps, executeSteps, title });
+
+        await wizard.prompt(actionContext);
+        // tslint:disable-next-line strict-boolean-expressions
+        if (showCreatingTreeItem) {
```
EricJizbaMSFT: No need to check `if (showCreatingTreeItem)`. It'll always be defined.

src/explorer/WebAppProvider.ts
```diff
@@ -60,31 +62,72 @@ export class WebAppProvider extends SubscriptionTreeItem {
...
+            // tslint:disable-next-line no-non-null-assertion
+            showCreatingTreeItem(wizardContext.newSiteName!);
```
EricJizbaMSFT: Doesn't have to be this PR, but eventually this repo should start using `nonNullProp`, etc. like other repos

src/commands/createWebApp.ts
nturinski: We already track runtime of a created app so this is probably unnecessary now.

src/commands/setAppWizardContextDefault.ts
nturinski: Okay to do this is another PR?

https://github.com/Microsoft/vscode-azureappservice/issues/979

EricJizbaMSFT: Honestly I think I would block on this one. `javaUtils` was never supposed to be in `vscode-azureappservice` and we only allowed it in temporarily because we had a release coming up. This is definitely on-topic for the PR, it's should be an easy copy/paste of the file, and that way you can include it in the existing breaking-change release of the shared package instead of doing another breaking change release right after.

src/explorer/editors/setAppWizardContextDefault.ts
```diff
@@ -0,0 +1,66 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
```
EricJizbaMSFT: Did you mean to put this in "editors"?

nturinski: No... "explorer". That seems like the appropriate spot, right?

src/explorer/WebAppProvider.ts
```diff
@@ -60,31 +61,70 @@ export class WebAppProvider extends SubscriptionTreeItem {
...
+            SiteOSStep.setLocationsTask(wizardContext);
         }
+
+        const title: string = 'Create new web app';
+        const wizard: AzureWizard<IAppServiceWizardContext> = new AzureWizard(wizardContext, { promptSteps, executeSteps, title });
+
+        await wizard.prompt(actionContext);
+
+        // tslint:disable-next-line no-non-null-assertion
+        showCreatingTreeItem(wizardContext.newSiteName!);
```
EricJizbaMSFT: Should be able to use nonNullProp in this file now

Issue #727, opened 2018-11-21
"Configure Deployment Source" should be on "Deployments" tree item


EricJizbaMSFT:
Currently "Configure Deployment Source" lives in the web app tree item:
<img width="342" alt="screen shot 2018-11-21 at 1 31 32 pm" src="https://user-images.githubusercontent.com/11282622/48864330-d4247500-ed91-11e8-86e8-16ca27d4672c.png">

However it feels more natural on the Deployments tree item:
<img width="352" alt="screen shot 2018-11-21 at 1 31 37 pm" src="https://user-images.githubusercontent.com/11282622/48864346-e0103700-ed91-11e8-92e4-35740f026e53.png">

I'm not sure we want to remove it from the web app tree item since some people are used to it there, but I figure we could at least add it to Deployments.

fiveisprime:
Agreed. This seems especially necessary once #753 is implemented.

PR #966, opened 2019-04-30
Add configureDeploymentSource to Deployments Node

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/727

Adds telemetry so we know what's the most common user entry point.

### DIFF
--- a/src/commands/deployments/editScmType.ts
+++ b/src/commands/deployments/editScmType.ts
@@ -13,11 +13,7 @@ export async function editScmType(actionContext: IActionContext, node?: SiteTree
     if (!node) {
         node = <DeploymentsTreeItem>await ext.tree.showTreeItemPicker([DeploymentsTreeItem.contextValueConnected, DeploymentsTreeItem.contextValueUnconnected]);
     } else if (node instanceof SiteTreeItem) {
-        actionContext.properties.contextValue = node.contextValue;
         node = <DeploymentsTreeItem>node.deploymentsNode;
-    } else {
-        // entered through deployments node context menu
-        actionContext.properties.contextValue = node.contextValue;
     }
 
     await appservice.editScmType(node.root.client, node.parent, actionContext);
--- a/src/commands/deployments/editScmType.ts
+++ b/src/commands/deployments/editScmType.ts
@@ -1,23 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DeploymentsTreeItem } from "vscode-azureappservice";
 import * as appservice from "vscode-azureappservice";
-import { AzureTreeItem, IActionContext } from "vscode-azureextensionui";
+import { IActionContext } from "vscode-azureextensionui";
 import { SiteTreeItem } from "../../explorer/SiteTreeItem";
-import { WebAppTreeItem } from "../../explorer/WebAppTreeItem";
 import { ext } from "../../extensionVariables";
 
-export async function editScmType(actionContext: IActionContext, node?: SiteTreeItem | appservice.DeploymentsTreeItem): Promise<void> {
+export async function editScmType(actionContext: IActionContext, node?: SiteTreeItem | DeploymentsTreeItem): Promise<void> {
     if (!node) {
-        node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -206,11 +207,9 @@ export async function activateInternal(
         registerCommand('appService.Deploy', async function (this: IActionContext, target?: vscode.Uri | WebAppTreeItem | undefined): Promise<void> {
             await deploy(this, true, target);
         });
-        registerCommand('appService.ConfigureDeploymentSource', async function (this: IActionContext, node?: SiteTreeItem): Promise<void> {
-            if (!node) {
-                node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
-            }
-            await editScmType(node.root.client, node, this);
+        registerCommand('appService.ConfigureDeploymentSource', async function (this: IActionContext): Promise<void> {
+            await editScmType(this);
```
EricJizbaMSFT: Shouldn't there be a `node` to pass along?

src/commands/deployments/editScmType.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function editScmType(actionContext: IActionContext, node?: SiteTreeItem | appservice.DeploymentsTreeItem): Promise<void> {
+    if (!node) {
+        node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+    } else if (node instanceof appservice.DeploymentsTreeItem) {
+        node = <SiteTreeItem>node.parent;
+    }
+
+    await appservice.editScmType(node.root.client, node, actionContext);
+    const children: AzureTreeItem[] = await node.getCachedChildren();
+    const deploymentsTreeItem: AzureTreeItem | undefined = children.find(ti => {
```
EricJizbaMSFT: You can just use this to get the deployments node:
https://github.com/Microsoft/vscode-azureappservice/blob/master/src/explorer/SiteTreeItem.ts#L27

src/commands/deployments/editScmType.ts
```diff
@@ -0,0 +1,23 @@
+import * as appservice from "vscode-azureappservice";
```
EricJizbaMSFT: nit: copyright header

src/commands/deployments/editScmType.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AzureTreeItem, IActionContext } from "vscode-azureextensionui";
+import { SiteTreeItem } from "../../explorer/SiteTreeItem";
+import { WebAppTreeItem } from "../../explorer/WebAppTreeItem";
+import { ext } from "../../extensionVariables";
+
+export async function editScmType(actionContext: IActionContext, node?: SiteTreeItem | appservice.DeploymentsTreeItem): Promise<void> {
+    if (!node) {
+        node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue);
+    } else if (node instanceof appservice.DeploymentsTreeItem) {
```
EricJizbaMSFT: It looks like you removed the context menu from SiteTreeItem. Do you need to check this anymore?

nturinski: Yeah because the node picker will still return a SiteTreeItem

EricJizbaMSFT: Then change the node picker to get the deployments tree item?

nturinski:
@fiveisprime Are we fine to remove this from the web app context menu and have it only on the Deployments context menu?

src/commands/deployments/editScmType.ts
```diff
@@ -0,0 +1,25 @@
...
+import * as appservice from "vscode-azureappservice";
+import { IActionContext } from "vscode-azureextensionui";
+import { SiteTreeItem } from "../../explorer/SiteTreeItem";
+import { ext } from "../../extensionVariables";
+
+export async function editScmType(actionContext: IActionContext, node?: SiteTreeItem | DeploymentsTreeItem): Promise<void> {
+    if (!node) {
+        node = <DeploymentsTreeItem>await ext.tree.showTreeItemPicker([DeploymentsTreeItem.contextValueConnected, DeploymentsTreeItem.contextValueUnconnected]);
+    } else if (node instanceof SiteTreeItem) {
+        actionContext.properties.contextValue = node.contextValue;
```
EricJizbaMSFT: No need to do this. It's handled here: https://github.com/Microsoft/vscode-azuretools/blob/de0a8e1a1c17ff9f5b3b4fecc53b163cb64e3625/ui/src/AzureActionHandler.ts#L30

nturinski: Your knowledge of the codebase is impressive as always 😅

Issue #976, opened 2019-05-06
Pop up an error "Cannot read property 'root' of undefined" when selecting "Configure Deployment Source..." with a collapsed web app


v-dantch:
**OS:** Mac
**Build Version:** [20190505.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=5719)
**Regression:** Previous Release (Not repro on 0.14.0)

**Repro Steps:**
1. Refresh the App Service Explorer and make sure all web apps are collapsed.
2. Right click one web app and select "Configure Deployment Source...".
3. Check the result.

**Expect:**
Succeed to execute this action without error.

**Actual:**
Pop up an error "Cannot read property 'root' of undefined".
![image](https://user-images.githubusercontent.com/34729092/57218214-9e72f180-6fa9-11e9-9e6f-e592d724d1a1.png)

**More Info:**
1. This issue doesn't reproduce on the previous release 0.14.0.
2. This issue doesn't reproduce on Windows.
3. This issue doesn't reproduce when expanding the web app.

PR #978, opened 2019-05-06
Get children if they arenlt loaded yet

nturinski:
Fixes https://github.com/Microsoft/vscode-azureappservice/issues/976

### DIFF
--- a/src/commands/deployments/editScmType.ts
+++ b/src/commands/deployments/editScmType.ts
@@ -7,18 +7,19 @@ import { DeploymentsTreeItem } from "vscode-azureappservice";
 import * as appservice from "vscode-azureappservice";
 import { IActionContext } from "vscode-azureextensionui";
 import { SiteTreeItem } from "../../explorer/SiteTreeItem";
+import { WebAppTreeItem } from "../../explorer/WebAppTreeItem";
 import { ext } from "../../extensionVariables";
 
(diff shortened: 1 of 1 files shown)


src/commands/deployments/editScmType.ts
```diff
@@ -13,6 +13,9 @@ export async function editScmType(actionContext: IActionContext, node?: SiteTree
     if (!node) {
         node = <DeploymentsTreeItem>await ext.tree.showTreeItemPicker([DeploymentsTreeItem.contextValueConnected, DeploymentsTreeItem.contextValueUnconnected]);
     } else if (node instanceof SiteTreeItem) {
+        if (!node.deploymentsNode) {
```
EricJizbaMSFT: I think a better fix would be to define `deploymentsNode` in the constructor of `SiteTreeItem` (you don't even have to check the tier, you could just default it to something). Otherwise it feels like the type of `deploymentsNode` should be changed to `| undefined` so that strictNullChecks will find any other similar bugs

nturinski: With changing the constructor, but unless you refresh the web app, you won't have an accurate ScmType on the Deployments node.  I could make `loadMoreChildren` update `this.deploymentsNode` immediately but that feels kind of pointless.

The type actually is defined as `DeploymentsTreeItem | undefined` but I cast it as `<DeploymentsTreeItem>` which is why I missed this error 😓

PR #990, opened 2019-05-16
NA tree type to easily filter NA tree items

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/955

It should be easier to check for this type and filter it out of the tree picker and for sorting children.


src/explorer/WebAppTreeItem.ts
```diff
@@ -46,9 +47,9 @@ export class WebAppTreeItem extends SiteTreeItem {
     }
 
     public compareChildrenImpl(ti1: AzureTreeItem<ISiteTreeRoot>, ti2: AzureTreeItem<ISiteTreeRoot>): number {
```
EricJizbaMSFT: You should probably move `compareChildrenImpl` to `SiteTreeItem` since `WebJobsNATreeItem` could also show up under a slot

PR #1004, opened 2019-05-22
Move project selection before node picker

### LINKED ISSUE #899: Ask for folder to deploy before asking for web app
For two reasons:
1. Creating a web app can take a while, whereas there's no long-running action after selecting the folder. Better to get all prompts out of the way first
1. The folder should be used to select smart defaults when creating the web app (Not a huge deal because this only matters if you have a multi-root workspace)

nturinski:
Fixes #899 

So I wasn't sure how to address 2. of the issue.  I'm guessing that I would have to make the `AzExtParentTreeItem` in the `ui` accept optional parameters and then check if fsPath exists when I'm creating the app in the extension??


src/commands/deploy.ts
```diff
@@ -66,6 +66,22 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
+        // we can only get the siteConfig if the entry point was a treeItem
+        siteConfig = node ? await node.root.client.getSiteConfig() : undefined;
+
+        if (siteConfig && javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
+            const fileExtension: string = javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion);
+            fsPath = await javaUtils.showQuickPickByFileExtension(context, `Select the ${fileExtension} file to deploy...`, fileExtension);
+            // if there is a siteConfig, then there is a node
+            await javaUtils.configureJavaSEAppSettings(nonNullValue(node));
+        } else {
+            fsPath = await workspaceUtil.showWorkspaceFoldersQuickPick("Select the folder to deploy", context, constants.configurationSettings.deploySubpath);
```
EricJizbaMSFT: This won't work for Java projects if they use the command palette or blue up arrow. One idea - you could combine `showQuickPickByFileExtension` and `showWorkspaceFoldersQuickPick` to work in both cases. In other words, show both the workspace folders _and_ any *.war/*.jar files

src/commands/deploy.ts
```diff
@@ -66,6 +66,22 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
+
+    if (!fsPath) {
+        // we can only get the siteConfig if the entry point was a treeItem
+        siteConfig = node ? await node.root.client.getSiteConfig() : undefined;
+
+        if (siteConfig && javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
+            const fileExtension: string = javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion);
+            fsPath = await javaUtils.showQuickPickByFileExtension(context, `Select the ${fileExtension} file to deploy...`, fileExtension);
+            // if there is a siteConfig, then there is a node
+            await javaUtils.configureJavaSEAppSettings(nonNullValue(node));
```
EricJizbaMSFT: I think this needs to happen in all entry points

EricJizbaMSFT:
> So I wasn't sure how to address 2. of the issue. I'm guessing that I would have to make the AzExtParentTreeItem in the ui accept optional parameters and then check if fsPath exists when I'm creating the app in the extension??

This was the main point of https://github.com/microsoft/vscode-azuretools/pull/503. You can add the fsPath to the context object that gets passed to `showTreeItemPicker`. It's the same context object that you end up getting [here](https://github.com/microsoft/vscode-azureappservice/blob/master/src/explorer/SubscriptionTreeItem.ts#L64)

src/explorer/setAppWizardContextDefault.ts
```diff
@@ -9,32 +9,37 @@ import { Uri, workspace, WorkspaceConfiguration } from 'vscode';
...
+            wizardContext.recommendedSiteRuntime = [
+                LinuxRuntimes.java,
+                LinuxRuntimes.tomcat,
+                LinuxRuntimes.wildfly
+            ];
+
+            // considering high resource requirement for Java applications, a higher plan sku is set here
+            wizardContext.newPlanSku = { name: 'P1v2', tier: 'PremiumV2', size: 'P1v2', family: 'P', capacity: 1 };
+            // to avoid 'Requested features are not supported in region' error
+            await LocationListStep.setLocation(wizardContext, 'weseteurope');
```
nturinski: For L62-65 where we recommend Windows based on a `csproj` file, there wasn't a clean way to check the current fsPath since I am using `workspace.findFiles`.  I can use fse and recursively search their directory but wasn't sure if it was worth it

src/commands/deploy.ts
```diff
@@ -102,16 +118,10 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
-        if (javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
-            const fileExtension: string = javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion);
-            fsPath = await javaUtils.showQuickPickByFileExtension(context, `Select the ${fileExtension} file to deploy...`, fileExtension);
-            await javaUtils.configureJavaSEAppSettings(node);
-        } else {
-            fsPath = await workspaceUtil.showWorkspaceFoldersQuickPick("Select the folder to deploy", context, constants.configurationSettings.deploySubpath);
-        }
+    siteConfig = siteConfig ? siteConfig : await node.root.client.getSiteConfig();
+    if (javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
+        await javaUtils.configureJavaSEAppSettings(node);
```
nturinski: This will cause another prompt to come up after creation.  I was thinking I could get the port value from the user before creation and then just retroactively run the command after it's done (since this relies on the app being created before it can update the app settings).

EricJizbaMSFT: I'm fine with prompting for port after creation in the Java case.

src/commands/deploy.ts
```diff
@@ -66,6 +67,21 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
 
+    let siteConfig: WebSiteModels.SiteConfigResource | undefined;
+
+    if (!fsPath) {
+        // we can only get the siteConfig if the entry point was a treeItem
+        siteConfig = node ? await node.root.client.getSiteConfig() : undefined;
+
+        if (siteConfig && javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
+            const fileExtension: string = javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion);
+            fsPath = await javaUtils.showQuickPickByFileExtension(context, `Select the ${fileExtension} file to deploy...`, fileExtension);
```
EricJizbaMSFT: Can't you get rid of `javaUtils.showQuickPickByFileExtension` and just use `showWorkspaceFoldersAndFiles` in all cases? I don't like having different code-paths/prompts for command palette vs. right click on a node.

src/IDeployWizardContext.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from "vscode-azureextensionui";
+
+export interface IDeployWizardContext extends IActionContext {
```
EricJizbaMSFT: Can you move this interface into `setAppWizardContextDefault.ts`? Seems excessive to have a file at the root of `src` just for this

src/explorer/setAppWizardContextDefault.ts
EricJizbaMSFT: findFiles lets you pass in a `RelativePattern` to restrict the glob to just a single workspace

src/explorer/setAppWizardContextDefault.ts
```diff
@@ -9,32 +9,37 @@ import { Uri, workspace, WorkspaceConfiguration } from 'vscode';
...
 import { configurationSettings, extensionPrefix } from '../constants';
+import { IDeployWizardContext } from '../IDeployWizardContext';
 import { javaUtils } from '../utils/javaUtils';
 
-export async function setAppWizardContextDefault(wizardContext: IAppServiceWizardContext): Promise<void> {
-    const isJavaProject: boolean = await javaUtils.isJavaProject();
+export async function setAppWizardContextDefault(wizardContext: IAppServiceWizardContext & IDeployWizardContext): Promise<void> {
+    // if the user entered through "Deploy", we'll have a project to base our recommendations on
+    // otherwise, look at their current workspace and only suggest if one workspace is opened
+    const workspaceForRecommendation: string | undefined = wizardContext.fsPath ?
```
EricJizbaMSFT: Feels safer to get the containing workspace for fsPath and do your checks against that. Like [here](https://github.com/microsoft/vscode-azurefunctions/blob/master/src/utils/workspace.ts#L74)

src/utils/workspace.ts
```diff
@@ -51,22 +51,30 @@ export async function selectWorkspaceItem(placeHolder: string, options: vscode.O
     return folder && folder.data ? folder.data : (await ext.ui.showOpenDialog(options))[0].fsPath;
 }
 
-export async function showWorkspaceFoldersQuickPick(placeHolderString: string, context: IActionContext, subPathSetting: string | undefined): Promise<string> {
-    context.telemetry.properties.cancelStep = 'showWorkspaceFolders';
+export async function showWorkspaceFoldersAndFiles(placeHolderString: string, context: IActionContext, subPathSetting: string | undefined, fileExtensions?: string | string[]): Promise<string> {
+    context.telemetry.properties.cancelStep = 'showWorkspaceFoldersAndExtensions';
+
+    if (!fileExtensions) {
```
EricJizbaMSFT: I don't see you passing in `fileExtensions` ever? So we don't need this check?

src/utils/workspace.ts
```diff
@@ -51,22 +51,30 @@ export async function selectWorkspaceItem(placeHolder: string, options: vscode.O
...
+        fileExtensions = [fileExtensions];
+    }
+
     return await selectWorkspaceItem(
         placeHolderString,
         {
-            canSelectFiles: false,
-            canSelectFolders: true,
+            canSelectFiles: true,
+            canSelectFolders: false,
```
EricJizbaMSFT: The point of doing `showWorkspaceFoldersAndFiles` was to support both java and non-java in the same code path. If you can't select both files and folders I think that defeats the point

nturinski: So this is true and this was a mistake, buuuuut...

```
        /**
	 * Options to configure the behaviour of a file open dialog.
	 *
	 * * Note 1: A dialog can select files, folders, or both. This is not true for Windows
	 * which enforces to open either files or folder, but *not both*.
	 * * Note 2: Explicitly setting `canSelectFiles` and `canSelectFolders` to `false` is futile
	 * and the editor then silently adjusts the options to select files.
	 */```

😬

File: src/explorer/setAppWizardContextDefault.ts (hunk 7 above)
Comment (nturinski) [reply src/explorer/setAppWizardContextDefault.ts]: But if the project that they're deploying isn't opened in their workspace, this wouldn't work, right?  I thought we wanted to use the deployed project to set defaults regardless of whether or not it's in the current workspace.

File: src/explorer/setAppWizardContextDefault.ts (hunk 3 above)
Comment (nturinski) [reply src/explorer/setAppWizardContextDefault.ts]: The `RelativePattern` only works if the workspace is opened so if the user deployed a not opened workspace, findFiles won't recognize that fsPath. :/

Comment (nturinski) [reply src/explorer/setAppWizardContextDefault.ts]: Never mind, I'm wrong about this.  I didn't understand how to use `RelativePattern`s

File: src/explorer/setAppWizardContextDefault.ts (hunk 7 above)
Comment (EricJizbaMSFT) [reply src/explorer/setAppWizardContextDefault.ts]: If you want to make it work in that scenario I'm totally fine with that. I was only thinking about the multi-root workspace when I originally filed the issue

File: src/utils/workspace.ts (hunk 9 above)
Comment (EricJizbaMSFT) [reply src/utils/workspace.ts]: Summarizing offline discussion:
That's unfortunate. At the end of the day this PR is about improving the experience for the majority of languages by prompting for folder before the web app. We still have to handle edge-cases (aka Java), but it doesn't have to be an ideal user experience. I don't have an exact answer on what the experience should be and I'm hoping you can come up with something 🙂 but let me or Matt know if you need help brainstorming.

**Comment by nturinski** [comment]
Okay so the new experience has a regression for Java users, but I tried to mitigate it as much as I could.

Since the current experience is to pick the app before deploying, we could detect if it was a Java project and prompt the user for a `.jar/.war` file.

However, the new experience is as follows:

On Windows, you can only pick a File OR a Folder.  If both are set to true, then it will default to Folder.  On OSX, this limitation doesn't exist, so you will be able to pick either.  I'm wondering if I should just filter out for `.zip/.jar/.war` files by default since I think that's all we support but wasn't sure if we'd want to have arbitrary restrictions like that.

Once you select the workspace, you select the app.  If you are creating a new app, then that workspace is what's used to for giving users the recommendations.  This only applies to workspaces that are currently opened.

If you select a non-Java app, everything proceeds as normal.  If you selected a Java app, then it will check if there is an artifact file within that folder.  If yes, then it'll prompt for the artifact which is an attempt to imitate the old behavior.  If there isn't a matching artifact file, then it'll proceed to deploy that folder.

javadeploy

File: src/utils/workspace.ts [hunk 10]
```diff
@@ -31,32 +32,36 @@ export async function selectWorkspaceFile(placeHolder: string, getSubPath?: (f:
...
-export async function showWorkspaceFoldersQuickPick(placeHolderString: string, context: IActionContext, subPathSetting: string | undefined): Promise<string> {
-    context.telemetry.properties.cancelStep = 'showWorkspaceFolders';
+export async function showWorkspaceFolders(placeHolderString: string, context: IActionContext, subPathSetting: string | undefined, fileExtension?: string): Promise<string> {
+    context.telemetry.properties.cancelStep = 'showWorkspaceFoldersAndExtensions';
     return await selectWorkspaceItem(
         placeHolderString,
         {
-            canSelectFiles: false,
+            // on Windows, if both files and folder are set to true, then it defaults to folders
+            canSelectFiles: true,
```
EricJizbaMSFT: > I'm wondering if I should just filter out for .zip/.jar/.war files by default since I think that's all we support but wasn't sure if we'd want to have arbitrary restrictions like that.

Yes I think you should only show those file types

src/utils/workspace.ts
```diff
@@ -66,7 +71,28 @@ export async function showWorkspaceFoldersQuickPick(placeHolderString: string, c
...
+export function getContainingWorkspace(fsPath: string): vscode.WorkspaceFolder | undefined {
+    // tslint:disable-next-line:strict-boolean-expressions
+    const openFolders: vscode.WorkspaceFolder[] = vscode.workspace.workspaceFolders || [];
+    return openFolders.find((f: vscode.WorkspaceFolder): boolean => {
+        return isPathEqual(f.uri.fsPath, fsPath) || isSubpath(f.uri.fsPath, fsPath);
+    });
+}
+
+export async function findFilesByFileExtension(deployPath: string | undefined, fileExtension: string): Promise<IAzureQuickPickItem<string | undefined>[]> {
+    // if there is a deployPath, then only check the deployed folder for the file extension, otherwise use all currently opened workspaces
```
EricJizbaMSFT: This file should have generic workspace utils, so you shouldn't reference a specific command like "deploy". Just do "folderPath" or "fsPath".

Also this returns quick picks? That seems weird to me. I would expect it to return the files.

nturinski:
Made enough changes that I wouldn't mind a final feedback loop 😅

src/utils/workspace.ts
```diff
@@ -31,42 +32,71 @@ export async function selectWorkspaceFile(placeHolder: string, getSubPath?: (f:
...
-    context.telemetry.properties.cancelStep = 'showWorkspaceFolders';
+export async function showWorkspaceFolders(placeHolderString: string, context: IActionContext, subPathSetting: string | undefined, fileExtension?: string): Promise<string> {
+    context.telemetry.properties.cancelStep = 'showWorkspaceFoldersAndExtensions';
     return await selectWorkspaceItem(
         placeHolderString,
         {
-            canSelectFiles: false,
+            // on Windows, if both files and folder are set to true, then it defaults to folders
+            canSelectFiles: true,
             canSelectFolders: true,
```
EricJizbaMSFT: This should probably be `canSelectFolders: !fileExtension`

src/utils/workspace.ts
```diff
@@ -31,42 +32,71 @@ export async function selectWorkspaceFile(placeHolder: string, getSubPath?: (f:
...
             canSelectMany: false,
-            defaultUri: vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0].uri : undefined
+            defaultUri: vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0].uri : undefined,
+            filters: { Artifacts: ['jar', 'war', 'zip'] }
```
EricJizbaMSFT: I think `fileExtension` should take priority here if it's defined.

Issue #982, opened 2019-05-07
Remote Debugging doesn't work for Node LTS version


nturinski:
The way that the extension checks for compatibility doesn't account for the LTS version used by the portal.

Issue #889, opened 2019-03-06
[Suggestion]Change 'Start Remote Debugging' to 'Enable Remote Debugging'


v-dantch:
**OS**: All
**Build Version:** [20190306.3](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=3743)
**Regression:** Not a regression

**Repro Steps:**
Right click one web app-->Check the context menu.

**Actual:**
Display 'Start Remote Debugging'.
![image](https://user-images.githubusercontent.com/34729092/53872029-2e98b880-3fb2-11e9-9984-8cc02d80d724.png)

**Suggestion:**
Change 'Start Remote Debugging' to 'Enable Remote Debugging'.

nturinski:
Definitely need to change the 'Yes' confirmation buttons to Enable/Disable.  Also change "Disable Remote Debugging" context action to "Stop".

PR #1005, opened 2019-05-22
For Remote debugging check for lts verison and change disable to stop

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/982
Fixes https://github.com/microsoft/vscode-azureappservice/issues/889

I think I changed it from Yes to Enable/Disable back when I did the SSH work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -218,7 +218,7 @@
             },
             {
                 "command": "appService.DisableRemoteDebug",
-                "title": "Disable Remote Debugging",
+                "title": "Stop Remote Debugging",
                 "category": "Azure App Service"
             },
             {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -218,7 +218,7 @@
             },
             {
                 "command": "appService.DisableRemoteDebug",
```
EricJizbaMSFT: I'd prefer if you made the command name consistent with 'Stop' as well

nturinski: What about the telemetry event?

EricJizbaMSFT: What about it?

Issue #816, opened 2019-01-14
Display 'Deploy to Web App...' for slots


v-dantch:
**OS:** All
**Node.js:** V10.14.0
**Build Version:** [20190112.3](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=2047)

**Repro Steps:**
1. Create a new web app ->deployment slot.
2. Right click the new created slot and check the context menu.

**Actual:**
Display 'Deploy to Web App...'.

**Suggestion:**
Display 'Deploy to Deployment Slot...' or other similar.
![image](https://user-images.githubusercontent.com/34729092/51105705-b4a34900-1824-11e9-91cd-a134569cb75d.png)

PR #1009, opened 2019-05-25
Add deploy to slot command

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/816

If deployment slots haven't been loaded yet, I didn't think it was worth loading everything before calling the command.  If the user doesn't, they will see this instead:

image

It will not appear on the quick pick list after every subsequent call through the palette.  That being said, I need to leverage the back button so that it'll be a less frustrating experience.

### DIFF
--- a/src/explorer/WebAppTreeItem.ts
+++ b/src/explorer/WebAppTreeItem.ts
@@ -18,9 +18,6 @@ import { DeploymentSlotTreeItem } from './DeploymentSlotTreeItem';
 import { SiteTreeItem } from './SiteTreeItem';
 
 export class WebAppTreeItem extends SiteTreeItem {
-    public static contextValue: string = extensionPrefix;
-    public readonly contextValue: string = WebAppTreeItem.contextValue;
-    public deploymentSlotsNode: DeploymentSlotsTreeItem | DeploymentSlotsNATreeItem;
 
     public get label(): string {
         return this.root.client.siteName;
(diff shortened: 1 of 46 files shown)


EricJizbaMSFT:
Summarizing offline discussion:

1. Current implementation won't quite work because clicking "Scale up to ..." in your screenshot won't actually open the browser to the scale up page. Try implementing [`pickTreeItemImpl`](https://github.com/microsoft/vscode-azuretools/blob/master/ui/index.d.ts#L257) to do your own custom quick pick or warning
1. Not sure how I feel about filtering out web apps that don't support slots. I would prefer one of two options:
    1. Copy the behavior when filtering storage accounts, which says something like "Some accounts were filtered. Learn more..." at the bottom of the quick pick
    1. Just include all web apps and throw an error or show a warning if they don't support slots. This seems easier and in a way makes the "Scale Up..." option more discoverable. Plus I'm not sure how many users half a bunch of web apps with different skus

src/explorer/WebAppTreeItem.ts
```diff
@@ -29,6 +26,9 @@ export class WebAppTreeItem extends SiteTreeItem {
     public get iconPath(): IThemedIconPath {
         return getThemedIconPath('WebApp_color');
     }
+    public static contextValue: string = extensionPrefix;
```
EricJizbaMSFT: Did you mean to move these?

nturinski: No, I wonder if it was some auto-formatting shenanigans going on...

Issue #795, opened 2019-01-03
Alert when creating in an app in a (B1) plan that already has three apps


fiveisprime:
Perf starts to degrade when adding the fourth app to a plan. We can help in the following ways

1. Show the number of apps in a plan in the plan selection prompt
1. Warn when creating in a plan that already has three apps (for basic mode in B1 plans)

"The selected plan currently has _n_ apps. Deploying additional apps may degrade the performance on the apps in the plan. Are you sure you want to deploy to _plan_?"

nturinski:
@fiveisprime 

Discussed this with @StephenWeatherford offline. What if we just created a new RG for each app in Basic creation?  So if I'm create naturins-app-01, it just creates a rg and asp with the same name.  I think when you're trying to go back and delete resources, it can be confusing to have a bunch of apps in one rg.

StephenWeatherford:
Does basic creation use consumption or service plans?  If think you want to reduce usage of service plans, 'cause each is tied to a VM instance, right?  That doesn't preclude the fact that app services could be created in new RGs I don't think.

PR #1022, opened 2019-06-04
Scale App Service Plans if in basic creation, free/basic tier have more than 3 apps

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/795
Relies on https://github.com/microsoft/vscode-azuretools/pull/517

I moved the try/catch in CreateWebApp since users can't get that prompt if an app were to error out when created from the node picker.

Would it be worth it to show this prompt even in advanced creation?


src/explorer/SubscriptionTreeItem.ts
```diff
@@ -105,9 +106,41 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             const location: Location = nonNullProp(wizardContext, 'location');
             wizardContext.newResourceGroupName = `appsvc_rg_${wizardContext.newSiteOS}_${location.name}`;
             wizardContext.newPlanName = `appsvc_asp_${wizardContext.newSiteOS}_${location.name}`;
+
+            const asp: AppServicePlan | null = await getAppServicePlan(wizardContext, wizardContext.newResourceGroupName, wizardContext.newPlanName);
+            if (asp && checkPlanForPerformanceDrop(asp)) {
+                // Subscriptions can only have 1 free tier Linux plan so show the warning if there are too many apps on the plan
+
+                if (isPlanLinux(asp)) {
+                    const showPerfWarning: boolean | undefined = workspaceConfig.get(configurationSettings.showPlanPerformanceWarning);
```
EricJizbaMSFT: I would prefer if you moved this logic inside of `showPerformancePrompt` (You'd probably have to rename the function at that point)

Otherwise people might look at `showPerformancePrompt` and think you forget to check the "Dont warn again" setting. Also then you don't have to define `showPlanPerformanceWarning` in `constants.ts` since it'll only be used in one function.

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -105,9 +106,41 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             const location: Location = nonNullProp(wizardContext, 'location');
             wizardContext.newResourceGroupName = `appsvc_rg_${wizardContext.newSiteOS}_${location.name}`;
             wizardContext.newPlanName = `appsvc_asp_${wizardContext.newSiteOS}_${location.name}`;
+
+            const asp: AppServicePlan | null = await getAppServicePlan(wizardContext, wizardContext.newResourceGroupName, wizardContext.newPlanName);
+            if (asp && checkPlanForPerformanceDrop(asp)) {
```
EricJizbaMSFT: createChildImpl is already pretty big - can we move this out somewhere?

src/validatePlanPerformance.ts
```diff
@@ -0,0 +1,73 @@
+import WebSiteManagementClient from "azure-arm-website";
+import { AppServicePlan } from "azure-arm-website/lib/models";
+import { ConfigurationTarget, MessageItem, workspace, WorkspaceConfiguration } from "vscode";
+import { AppServicePlanNameStep, IAppServiceWizardContext } from "vscode-azureappservice";
+import { createAzureClient, DialogResponses, IActionContext, UserCancelledError } from "vscode-azureextensionui";
+import { AppServiceDialogResponses, extensionPrefix, maxNumberOfSites } from "./constants";
+import { ext } from "./extensionVariables";
+import { nonNullProp } from "./utils/nonNull";
+
+export async function validatePlanPerformance(wizardContext: IAppServiceWizardContext, rgName: string, newPlanName: string): Promise<void> {
```
EricJizbaMSFT: Based on my other comments, I'm beginning to think we should rename this to `setDefaultRgAndPlanName` so that this code can be included in the function:

```
// this should always be set when in the basic creation scenario
const location: Location = nonNullProp(wizardContext, 'location');	            const location: Location = nonNullProp(wizardContext, 'location');
wizardContext.newResourceGroupName = `appsvc_rg_${wizardContext.newSiteOS}_${location.name}`;	            wizardContext.newResourceGroupName = `appsvc_rg_${wizardContext.newSiteOS}_${location.name}`;
wizardContext.newPlanName = `appsvc_asp_${wizardContext.newSiteOS}_${location.name}`;	            wizardContext.newPlanName = `appsvc_asp_${wizardContext.newSiteOS}_${location.name}`;
```

It just feels like a cleaner delineation than `validatePlanPerformance` because it's not obvious _why_ you're validating the plan performance (to pick the default names).

src/validatePlanPerformance.ts
```diff
@@ -0,0 +1,73 @@
...
+    const asp: AppServicePlan | null = await getAppServicePlan(wizardContext, rgName, newPlanName);
+    if (asp) {
+        if (checkPlanForPerformanceDrop(asp)) {
+            // Subscriptions can only have 1 free tier Linux plan so show the warning if there are too many apps on the plan
+            if (isPlanLinux(asp)) {
```
EricJizbaMSFT: Just check `wizardContext.newSiteOS`

src/validatePlanPerformance.ts
```diff
@@ -0,0 +1,73 @@
...
+                await promptPerformanceWarning(wizardContext, asp);
+            } else {
+                // Subscriptions can have 10 free tier Windows plans so just create a new one with a suffixed name
+                // If there are 10 plans, it'll throw an error that directs them to advancedCreation
+                wizardContext.newPlanName = await new AppServicePlanNameStep().getRelatedName(wizardContext, newPlanName);
```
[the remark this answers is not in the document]
nturinski: I don't think resource groups have the same performance decrease as the plans.  At least, the original issue (https://github.com/microsoft/vscode-azureappservice/issues/795) only mentions that the asp directly relates to performance issues.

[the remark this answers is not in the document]
EricJizbaMSFT: Yeah I'm not concerned about perf in this case - it just felt cleaner IMO to have a separate resource group for each plan and it's web apps. But that could just be a personal preference, so I won't push on this if you prefer to have them all in the same group.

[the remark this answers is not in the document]
nturinski: The `isRelatedNameAvailable` is different from what I want to validate.  The name property assumes that I'm using the same name for the `resource group` and the `app service plan`, but that isn't usually going to be the case.  I guess we can increment the resource group name as well, but I wanted to throw all of the basic creation apps into one plan.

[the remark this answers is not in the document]
EricJizbaMSFT: Resource group naming rules are usually the loosest of all things in Azure. I don't think it'll affect the related name generated for the app service plan

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -103,19 +105,36 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         if (!advancedCreation) {
             // this should always be set when in the basic creation scenario
             const location: Location = nonNullProp(wizardContext, 'location');
-            wizardContext.newResourceGroupName = `appsvc_rg_${wizardContext.newSiteOS}_${location.name}`;
-            wizardContext.newPlanName = `appsvc_asp_${wizardContext.newSiteOS}_${location.name}`;
+            wizardContext.newResourceGroupName = `appsvc_${wizardContext.newSiteOS}_${location.name}`;
+            wizardContext.newPlanName = wizardContext.newResourceGroupName;
```
EricJizbaMSFT: Move these few lines inside `setDefaultRgAndPlanName` (the lines that are setting the first defaults)

nturinski: 😂

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -18,22 +18,42 @@ const maxNumberOfSites: number = 3;
...
-                // Subscriptions can have 10 free tier Windows plans so just create a new one with a suffixed name
-                // If there are 10 plans, it'll throw an error that directs them to advancedCreation
-                wizardContext.newResourceGroupName = await siteNameStep.getRelatedName(wizardContext, wizardContext.newPlanName);
-                wizardContext.newPlanName = wizardContext.newResourceGroupName;
+    const asp: AppServicePlan | null = await getAppServicePlan(wizardContext, defaultName, defaultName);
+    if (asp && checkPlanForPerformanceDrop(asp)) {
+        // Subscriptions can only have 1 free tier Linux plan so show a warning if there are too many apps on the plan
+        if (wizardContext.newSiteOS === WebsiteOS.linux) {
+            await promptPerformanceWarning(wizardContext, asp);
+            wizardContext.newResourceGroupName = defaultName;
```
EricJizbaMSFT: Would prefer if you did this:
```
wizardContext.newResourceGroupName = defaultName;
wizardContext.newPlanName = defaultName;
```

That way someone can't accidentally cause a bug by switching the order.

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -18,22 +18,42 @@ const maxNumberOfSites: number = 3;
...
+            const client: WebSiteManagementClient = createAzureClient(wizardContext, WebSiteManagementClient);
+            const allAppServicePlans: AppServicePlan[] = await client.appServicePlans.list();
+            const defaultPlans: AppServicePlan[] = allAppServicePlans.filter(plan => { return plan.name && plan.name.includes(defaultName); });
+
+            // try and find a default plan that has less than 3 plans first
+            for (const plan of defaultPlans) {
+                if (plan.name) {
+                    const fullPlanData: AppServicePlan | null = await getAppServicePlan(wizardContext, plan.name, plan.name);
+                    if (fullPlanData && !checkPlanForPerformanceDrop(fullPlanData)) {
+                        wizardContext.newResourceGroupName = plan.name;
```
EricJizbaMSFT: Same as other comment
```
wizardContext.newResourceGroupName = plan.name;
wizardContext.newPlanName = plan.name;
```

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -18,22 +18,42 @@ const maxNumberOfSites: number = 3;
...
+                    }
+                }
             }
+
+            // otherwise create a new rg and asp
+            wizardContext.newResourceGroupName = wizardContext.newResourceGroupName || await siteNameStep.getRelatedName(wizardContext, defaultName);
+            wizardContext.newPlanName = wizardContext.newResourceGroupName;
         }
+    } else {
+        wizardContext.newResourceGroupName = defaultName;
```
EricJizbaMSFT: same

Issue #1017, opened 2019-05-30
Stifle web app created notification when deploying to a new web app


nturinski:
Too many prompts

PR #1023, opened 2019-06-05
Checks to see if there was a deploy path to stifle create notification

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1017

I wanted to leverage the `confirmDeployment` boolean that we alter when we detect the web app was created from the deployer, but it's unfortunately too late by that point 😓


nturinski:
The accidental fix doesn't include the create message in the output channel.

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -61,7 +61,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         );
     }
 
-    public async createChildImpl(context: ICreateChildImplContext): Promise<AzureTreeItem> {
+    public async createChildImpl(context: ICreateChildImplContext & IDeployWizardContext): Promise<AzureTreeItem> {
```
EricJizbaMSFT: You should either change this to `Partial<IDeployWizardContext>` or use a separate `ICreateWebAppContext`.

I prefer a separate `ICreateWebAppContext` that has a an optional property `suppressCreateNotification`? I would trust that more than checking `context.fsPath` and that isolates `create` from `deploy` more cleanly.

Issue #748, opened 2018-11-30
Prompting to select one web app to deploy when there is no opened folder even though 'appService.defaultWebAppToDeploy' has been set


v-dantch:
**OS**: All
**Node.js:** V10.13.0
**Build Definition**: [20181129.5](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=1106)

**Repro Steps:**
1. Set 'appService.defaultWebAppToDeploy' with one valid full Azure web app id like below:
Azure id:
/subscriptions/92be7e56-8a60-427c-a3e3-0ebd54a827a1/resourceGroups/megantest/providers/Microsoft.Web/sites/window-1129
2. F1 ->Type and select 'Azure App Service: Deploy to Web App...'.
3. Check the result.

**Expect:**
Deploy to the default web app without asking to select.

**Actual:**
Prompt to select to one web app.
![image](https://user-images.githubusercontent.com/34729092/49278212-1733c600-f4bf-11e8-92c1-c7654c254b98.png)

**More Info:**
This issue doesn't reproduce when there is a opened folder.

EricJizbaMSFT:
Not sure what @nturinski means - "defaultWebAppToDeploy" should apply whether or not there's zero or more than 1 workspaces

nturinski:
@EricJizbaMSFT the setting is workspace specific, not on the user level.  It was specifically designed to work only when one workspace is opened because each workspace will have its own web app to deploy to.  We can change this behavior, but this isn't a bug.

v-dantch:
Got it. Thanks for your update.

EricJizbaMSFT:
We would need to handle the "Reset Default" button when deploying

PR #1024, opened 2019-06-05
Respect defaultApp settting even if there is multiroot or no workspace

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/748

It also will work for a multiroot workspace and only respect/reset the setting in the appropriate workspaces.

Also I'll move the IDeployWizardContext to its own file, but other PR also changes it so will do it in a separate PR

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -7,10 +7,9 @@ import * as WebSiteModels from 'azure-arm-website/lib/models';
 import { SiteConfigResource } from 'azure-arm-website/lib/models';
 import { pathExists } from 'fs-extra';
 import * as path from 'path';
-import { join } from 'path';
-import { commands, Disposable, MessageItem, Uri, window, workspace, WorkspaceConfiguration, WorkspaceFolder } from 'vscode';
+import { commands, ConfigurationTarget, Disposable, MessageItem, Uri, window, workspace, WorkspaceConfiguration, WorkspaceFolder } from 'vscode';
 import * as appservice from 'vscode-azureappservice';
-import { AzureTreeItem, DialogResponses, IAzureQuickPickItem, parseError } from 'vscode-azureextensionui';
+import { DialogResponses, IAzureQuickPickItem, parseError } from 'vscode-azureextensionui';
 import * as constants from '../constants';
 import { IDeployWizardContext } from '../explorer/setAppWizardContextDefault';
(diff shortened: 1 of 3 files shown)


src/commands/deploy.ts
```diff
@@ -23,64 +22,40 @@ import { isPathEqual, isSubpath } from '../utils/pathUtils';
 import { getRandomHexString } from "../utils/randomUtils";
 import * as workspaceUtil from '../utils/workspace';
 import { cancelWebsiteValidation, validateWebSite } from '../validateWebSite';
+import { getDefaultWebAppToDeploy } from './getDefaultWebAppToDeploy';
 import { startStreamingLogs } from './startStreamingLogs';
 
 // tslint:disable-next-line:max-func-body-length cyclomatic-complexity
```
EricJizbaMSFT: I like moving `getDefaultWebAppToDeploy` to its own file - does that mean we enable either or both of these rules now? 😄

src/commands/deploy.ts
```diff
@@ -154,38 +131,46 @@ export async function deploy(context: IDeployWizardContext, confirmDeployment: b
...
         if (result === resetDefault) {
-            const localRootPath = nonNullValue(currentWorkspace).uri.fsPath;
-            const settingsPath = path.join(localRootPath, '.vscode', 'settings.json');
-            const doc = await workspace.openTextDocument(Uri.file(settingsPath));
-            window.showTextDocument(doc);
-            await workspaceConfig.update(constants.configurationSettings.defaultWebAppToDeploy, '');
+            if (context.configurationTarget === ConfigurationTarget.Global) {
+                await commands.executeCommand('workbench.action.openGlobalSettings');
+            } else {
+                const localRootPath = nonNullValue(currentWorkspace).uri.fsPath;
```
EricJizbaMSFT: Can you change this case to use a command as well? `'workbench.action.openSettings'` maybe?

src/commands/deploy.ts
nturinski: There's still quite a bit of work to do on that front. . .

image

😓

EricJizbaMSFT: Oof

src/commands/deploy.ts
nturinski: Using `workbench.action.openSettingsJson` seems to just open the User settings as well whether I have a single root or multiroot workspace 😕

PR #1027, opened 2019-06-06
Copy production app's deployment source on createSlot

nturinski:
After reviewing the experience of skipping the org and repo prompts on slot creation, @fiveisprime  thought it was kind of weird since there's no context and I agree with him.

Also removed the deployToPrompt (I know that'll make Eric happy) for slots.  Matt says slots are an advanced scenario and it's unlikely an advanced user would deploy immediately after creating a slot without setting up some CI or app settings first.


src/explorer/DeploymentSlotTreeItem.ts
```diff
@@ -30,17 +25,4 @@ export class DeploymentSlotTreeItem extends SiteTreeItem {
     public get iconPath(): IThemedIconPath {
         return getThemedIconPath('DeploymentSlot_color');
     }
-    public promptToDeploy(context: IActionContext): void {
-        const createdNewSlotMsg: string = `Created new slot "${this.root.client.fullName}": https://${this.root.client.defaultHostName}`;
```
EricJizbaMSFT: > I know that'll make Eric happy

I think we should keep this notification 😬😬😬 It's a long running operation and I like having notifications when those complete. (just _one_ notification though haha)

I have less strong feelings about the "Deploy" button. I don't like it in the "Create Web App" scenario because there's a clear alternative to create/deploy at the same time (the big blue up arrow). There is no up arrow to create/deploy a slot at the same time. But I also agree that I don't think many people will do that so idk

EricJizbaMSFT:
Don't forget to bump the appservice package version. Pretty sure that's why builds are failing

EricJizbaMSFT:
Also can you bump the [ui package](https://github.com/microsoft/vscode-azuretools/releases/tag/v0.25.1-ui) while you're at it. Had some things like tracking accountStatus in telemetry that would be nice to include

PR #1044, opened 2019-06-12
Add "Generate Azure CLI Script" test cases

v-wuzhai:
Add "Generate Azure CLI Script" test cases

### DIFF
--- a/test/createAzureResource.test.ts
+++ b/test/createAzureResource.test.ts
@@ -52,7 +52,7 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
         ext.azureAccountTreeItem.dispose();
     });
 
-    test('Create and Delete New Web App (Advanced)', async () => {
+    test('Create New Web App (Advanced)', async () => {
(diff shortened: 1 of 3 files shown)


test/createAzureResource.test.ts
```diff
@@ -112,6 +115,14 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
         assert.equal(createdApp.scmType, constants.ScmType.None, `Web App scmType's property value should be ${constants.ScmType.None} rather than ${createdApp.scmType}.`);
     });
 
+    test('Generate Azure CLI Script', async () => {
+        const scriptTemplate: string = await generateAzureCLIScript('linux-default.sh', 'windows-default.sh');
```
nturinski: Since these are basically relying on the same template, it's not actually testing whether or not the CLI script would work.  I'm not sure if this test is worth having unless we run the script in the CLI and verify that it works.

Maybe we can try running it after the delete command and see if we can recreate it?  Of course, a challenge would be installing the Azure CLI

EricJizbaMSFT:
Closing due to lack of activity

PR #1056, opened 2019-06-27
Expose optional parameters for programmatic app creation

nturinski:
An intern team asked to expose the command appService.createWebApp so that they can leverage it in their extension.  They only asked for site name and resource group name and runtime, but I went ahead and added the easy to add properties.

Example usage: `await vscode.commands.executeCommand('appService.CreateWebApp', undefined, node.root.subscriptionId, 'naturinsmadewithapi1', 'naturinsmadewithapi1', undefined, undefined, WebsiteOS.linux, "NODE|10.14");`


EricJizbaMSFT:
We don't want to expose commands as our official api. This should be done similar to Cosmos DB: 

https://github.com/microsoft/vscode-cosmosdb/blob/master/src/vscode-cosmosdb.api.d.ts

https://github.com/microsoft/vscode-cosmosdb/blob/master/src/extension.ts#L151

src/vscode-azureappservice.api.d.ts
```diff
@@ -0,0 +1,36 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+export interface AzureAppServiceExtensionApi {
+    apiVersion: string;
+    /**
+     * Creates a web app with the given options.  Empty properties will be their default value or the user will be prompted for it
+     *
+     * @param subscriptionId The subscription id of an Azure subscription
```
nturinski: Most of the params are self-explanatory, so I only included the ones that might be difficult to understand the format.  But should I just explain all the parameters?

src/extension.ts
```diff
@@ -318,7 +319,10 @@ export async function activateInternal(
         registerCommand(showOutputChannelCommandId, () => { ext.outputChannel.show(); });
     });
 
-    return createApiProvider([]);
+    return createApiProvider([<AzureExtensionApi>{
+        apiVersion: '1.0.0',
+        createWebApp: api.createWebApp
```
EricJizbaMSFT: Anything in an API should have a test to accompany it to prevent us from accidentally breaking backwards compatibility

src/vscode-azureappservice.api.d.ts
EricJizbaMSFT: Soo... did they ask for all of these? When creating an API, it's generally best to only provide as little as possible. Otherwise you're stuck maintaining a bunch of stuff for backwards compatibility that no one's ever using.

src/commands/api/createWebApp.ts
```diff
@@ -0,0 +1,23 @@
...
+    subscriptionId?: string,
+    siteName?: string,
+    rgName?: string,
+    planName?: string,
+    planSku?: SkuDescription,
+    websiteOS?: WebsiteOS,
+    runtime?: string
+}): Promise<void> {
+
+    await callWithTelemetryAndErrorHandling('api.appService.CreateWebApp', async (context: IAppServiceWizardContext) => {
```
EricJizbaMSFT: nit: remove `appService`. All telemetry events are already prefixed with the extension name

src/vscode-azureappservice.api.d.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface AzureAppServiceExtensionApi {
+    apiVersion: string;
+    /**
+     * Creates a web app with the given options.  Empty properties will be their default value or the user will be prompted for it
+     *
+     * @param subscriptionId The subscription id of an Azure subscription: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+     * @param runtime The runtime must be formatted in the following: "NODE|10.14"
+     */
+    createWebApp(createOptions: {
```
EricJizbaMSFT: Add an interface for the `createOptions` type so that other people don't have to copy the whole type every time

src/vscode-azureappservice.api.d.ts
```diff
@@ -0,0 +1,23 @@
...
+        subscriptionId?: string,
+        siteName?: string,
+        rgName?: string,
+        os?: WebsiteOS,
+        runtime?: string
```
EricJizbaMSFT: I still feel like this has to be more options than necessary. If they know all these values - they could pretty much create the site themselves. In that case, I would guide them to [SiteCreateStep](https://github.com/microsoft/vscode-azuretools/blob/master/appservice/src/createAppService/SiteCreateStep.ts) and they can basically just copy that single file instead of leveraging our extension.

If they _don't_ know all these values (which seems more likely), then we should remove them from the API so that we don't have to support it going forward.

nturinski: Well, they only want to specifically make Linux apps that are node|lts or python|3.7.  Should I just make the command more inflexible and add support as needed?  i.e. change it so that runtime is a type that accepts either node or python?

nturinski: Also the intern team wants support for deploy as well.

EricJizbaMSFT: Adding an API should be a very deliberative process and has a higher bar for merging vs. typical PRs because it deals with external teams and backwards compatibility. I don't think `inflexible` is the right word, but the API should definitely be as limited as you can make it.

Based on your comment, it seems like they only want to specify `os` and `runtime`, which means you should remove `siteName`, `rgName`, and `subscriptionId`. In addition - would our default runtime/OS recommendations work for them? Why do they need to specify specific versions?

nturinski: Well, they specifically requested to be able to make their own `siteName`, `rgName`, and to be able to pick the subscription that they want.  I think they want as little of user interactivity as possible.

I'll ask them if they plan on having the user or the dev team decide on the siteName and rgName, but subscription id does seem like the user would have to select.

As far as the runtimes, I think that for their specs, they are only supporting node/python projects, so our list of runtimes would confuse their users since we display everything even if the recommended runtimes are on top.

nturinski: https://us-prod.asyncgw.teams.microsoft.com/v1/objects/0-cus-d9-f4c1c20924fc926ed1a59e1c1fd58787/views/imgo

Since it's all wizard-based, our extension's quickpick steps wouldn't fit their needs.  It seems like it might be easier for the team to just use the azure-arm-resources API or copy our SiteCreateStep.ts from here https://github.com/microsoft/vscode-azuretools/blob/master/appservice/src/createAppService/SiteCreateStep.ts

EricJizbaMSFT: Can't see that image, but yeah it sounds like they shouldn't go through us if they don't want user interactive stuff

test/api.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        const appServiceExtension: vscode.Extension<AzureExtensionApiProvider | undefined> | undefined = vscode.extensions.getExtension('ms-azuretools.vscode-azureappservice');
+        if (!appServiceExtension || !appServiceExtension.exports) {
+            throw new Error('App Service Extension did not load properly.');
+        }
+
+        appServiceApi = appServiceExtension.exports.getApi<AzureAppServiceExtensionApi>('^1.0.0');
+    });
+
+    test('Deploy command loaded properly', () => {
+        assert.ok(appServiceApi.deploy);
```
nturinski: I figured I would wait until we have a deploy test case and just use the same test case for the api, but for now, I just want to make sure it loaded properly.

src/commands/deploy.ts
```diff
@@ -36,6 +34,9 @@ export async function deploy(context: IDeployWizardContext, confirmDeployment: b
     if (target instanceof Uri) {
         context.fsPath = target.fsPath;
         context.telemetry.properties.deploymentEntryPoint = 'fileExplorerContextMenu';
+    } else if (target && !(target instanceof WebAppTreeItem)) {
```
nturinski: Is there a better way to do this?  Since Site is an interface, its instance can't be determined at runtime.  I tried using BaseResource (it's base class) but it wasn't recognizing it as that type.

nturinski:
The full scenario should be as the following:

The team who needs this is creating an extension that scaffolds either a Node.js or Python project for the user.  They will then use a wizard-like form to create a web app for the user.  Then they will immediately deploy the scaffolded project to the web app to show the user their deployed project.

I think the intention is every subsequent deployment after that would be via the App Service Extension.

The parameters required should only be the Site id and the vscode.workspace of the directory.  We'll use the id to find the node in the App Service tree view and deploy the workspace.

EricJizbaMSFT:
As discussed offline, I'm not sold that we need to provide an api for deploy. It seems weird to automatically deploy something without ever seeing the source code open in a workspace.

Either way, I think they should leverage `appService.defaultWebAppToDeploy` and `appService.deploySubpath` so that the user can have a one-click deploy for all deployments. (If we end up providing an api - I think this would replace them passing site id and workspace through the api)

EricJizbaMSFT:
@nturinski I think we decided not to support an API for create - so can we close this? I'm not sure where we landed on the api for 'deploy', but feels like that might be better as a fresh new PR anyways.

nturinski:
Closing due to lack of need for an API.

Issue #945, opened 2019-04-17
Improve smart defaults for .NET Core apps


fiveisprime:
- Use Linux by default
- Create settings.json with...
  - `preDeploy` of `dotnet publish`
  - `deploySubpath` set to correct path `bin/Debug/<framework version>/publish`

Note that users with Omnisharp may already have the publish task created. This requires the user to confirm they want to add the config files.

bradygaster:
mentioning @timheuer and @spboyer here as i discussed this with them too and they may have addendum.

timheuer:
```
"appService.deploySubpath": "bin\\Debug\\<fwversion>\\publish",
"appService.preDeployTask": "publish"
```
Even with these defaults the ```deploySubpath``` setting changes based on user interaction so that needs to be hardened once set or confirming overwriting more than it does today.

EricJizbaMSFT:
We could potentially add support for a special variable (something like `${targetFramework}`) that automatically parses your csproj to get the value and replaces it when deploying. Should be easy enough in most cases, although not sure what we'd do if they have multiple csproj files or a weird folder structure

timheuer:
That would be helpful @EricJizbaMSFT but we'd have to do it on first deploy.  We don't want to assume every csproj is something that will be deployed to Azure, otherwise we could do all this in Omnisharp extension today on startup.

EricJizbaMSFT:
@timheuer I may have missed some context here, but I was already assuming all this logic would be in the appservice extension, not omnisharp

timheuer:
@EricJizbaMSFT no you are right...just meant at the time of parsing the csproj file can't annoy the non-Azure customer :-).  I am pretty sure we are saying the same things, just differently.  VS Code has a nice means to enable setting vars and inspecting things on startup of an extension in a workspace which would have made inspecting the csproj easier...but we wouldn't want to do that in this case IMO at that sequence.

fiveisprime:
Another test case: `$ dotnet new mvp` fails to deploy today.

spboyer:
@fiveisprime what is the framework? 
can you `dotnet --info`

fiveisprime:
This was also reported by a customer. My result is:

```
.NET Core SDK (reflecting any global.json):
 Version:   2.2.300
 Commit:    73efd5bd87

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18908
 OS Platform: Windows
 RID:         win10-x64
[...]

PR #1060, opened 2019-07-02
Improve smart defaults for .NET Core apps

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/945

Right now, I'm just throwing the bits into a folder called `publish`.  But I'm contemplating whether I should put the output in a temp folder or clean up the bits after deployment.

The default will be defined by the `tasks.json`.  If the `preDeployTask` is already defined, then we assume that the task exists.  If it doesn't, this error comes up:

image


nturinski:
Offline feedback:
- Publish in a temp folder-- clean post-deployment
- Default web app to Linux (in the original issue)
- If the user deploys the published bits themselves, don't set preDeployTask for them

src/commands/createWebApp/setAppWizardContextDefault.ts
```diff
@@ -35,6 +36,9 @@ export async function setAppWizardContextDefault(wizardContext: IAppServiceWizar
             // requirements.txt are used to pip install so a good way to determine it's a Python app
             wizardContext.recommendedSiteRuntime = [LinuxRuntimes.python];
 
+        } else if ((await findFilesByFileExtension(fsPath, 'csproj')).length > 0) {
+            // wizardContext.recommendedSiteRuntime = ['dotnet'];
```
nturinski: Will need to uncomment this when this PR is in https://github.com/microsoft/vscode-azuretools/pull/528

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,77 @@
...
+    const dotnetOutputPath: string = path.join('bin', 'Debug', 'publish');
+
+    // tslint:disable-next-line: strict-boolean-expressions
+    if (!currentWorkspace || !getWorkspaceSetting('configurePreDeployTasks', currentWorkspace.uri.fsPath)) {
+        // if the workspace being deployed is not opened, return and do nothing
+        return;
+    }
+
+    const csProj = await workspaceUtil.findFilesByFileExtension(currentWorkspace.uri.fsPath, 'csproj');
+    if (csProj.length > 0) {
```
nturinski: Would it be better to check the siteConfig.linuxFxVersion is .NET? Or both?

nturinski:
Offline feedback:
Don't overwrite the preDeployTask if they already have something there.

src/commands/deploy/setPreDeployTaskForDotnet.ts
EricJizbaMSFT: Good question - I'd have to think about it more. Can you explain why you chose what you did?

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (!getWorkspaceSetting('configurePreDeployTasks', currentWorkspace.uri.fsPath)) {
+        return;
+    }
+
+    const csProj = await workspaceUtil.findFilesByFileExtension(currentWorkspace.uri.fsPath, 'csproj');
+    if (csProj.length > 0) {
+        // if a publish task is already defined, then assume that we don't need this logic
+        if (getWorkspaceSetting(constants.configurationSettings.preDeployTask, currentWorkspace.uri.fsPath) !== 'publish') {
+            // if this doesn't have the publish preDeployTask, configure for .NET depoyment
+            const csProjDoc: TextDocument = await workspace.openTextDocument(csProj[0]);
```
EricJizbaMSFT: I usually prefer `fs-extra` for reading a file instead of VS Code apis

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,75 @@
...
+            const csProjJson: string = csProjDoc.getText();
+
+            const tfw: string = 'TargetFrameworks';
```
EricJizbaMSFT: I'm not sold on this logic. I think it would be confusing to prompt for a target framework out of the blue and this feels like an edge case.

I would rather handle the more standard case like this: https://github.com/Azure-Samples/dotnet-core-api/blob/master/TodoApi.csproj#L4 and wait for feedback/telemetry before we get into more complex cases.

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,75 @@
...
+    const fsPath: string = nonNullProp(context, 'fsPath');
+    const currentWorkspace: WorkspaceFolder | undefined = nonNullValue(workspaceUtil.getContainingWorkspace(fsPath));
+    // follow the publish output patterns, but leave out tfw
+    const dotnetOutputPath: string = path.join('bin', 'Debug', 'publish');
+
+    if (!getWorkspaceSetting('configurePreDeployTasks', currentWorkspace.uri.fsPath)) {
+        return;
+    }
+
+    const csProj = await workspaceUtil.findFilesByFileExtension(currentWorkspace.uri.fsPath, 'csproj');
```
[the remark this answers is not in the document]
nturinski: I believe that this only looks in the root folder.  I'm using the vscode.workspace.findFiles with a relative pattern.

src/commands/deploy/setPreDeployTaskForDotnet.ts
nturinski: I think the .csproj file is what's use to detect whether or not something is a C# project in the Omnisharp extension @fiveisprime at least recommended that's what we do.  

A publish will fail if we don't have the .csproj file, so I guess it makes sense to check for that.

src/commands/deploy/setPreDeployTaskForDotnet.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Well I think it's pretty common to put it in a sub folder so this is worth figuring out

src/commands/createWebApp/setAppWizardContextDefault.ts
EricJizbaMSFT: That was merged a while ago - anything you're waiting on to fix this comment?

src/commands/deploy/setPreDeployTaskForDotnet.ts
EricJizbaMSFT: Omnisharp is a completely different scenario IMO. They _add_ .NET functionality to your workspace, but don't block other languages. Here we have to choose one and _only_ one language used for deploying.

We already have logic to detect if a project is a node.js app (aka if it has a package.json) or java app. I think you need to somehow refactor so that if a workspace matches multiple languages (aka both a csproj and a package.json) you can bail out and not use the defaults for either language.

As an alternative, perhaps you _should_ check `siteConfig.linuxFxVersion` since that seems more deterministic. But that would only work for linux and might have other drawbacks

src/commands/deploy/deploy.ts
```diff
@@ -7,53 +7,40 @@ import * as WebSiteModels from 'azure-arm-website/lib/models';
...
-        siteConfig = node ? await node.root.client.getSiteConfig() : undefined;
-
-        if (siteConfig && javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
-            const fileExtension: string = javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion);
-            context.fsPath = await workspaceUtil.showWorkspaceFolders(`Select the ${fileExtension} file to deploy...`, context, constants.configurationSettings.deploySubpath, fileExtension);
-        } else {
-            context.fsPath = await workspaceUtil.showWorkspaceFolders("Select the folder to deploy", context, constants.configurationSettings.deploySubpath);
-        }
-    }
+    context.workspace = currentWorkspace;
```
EricJizbaMSFT: How were you able to get rid of the java logic?

src/commands/createWebApp/setAppWizardContextDefault.ts
```diff
@@ -5,28 +5,30 @@
...
 
 export interface IDeployWizardContext extends IActionContext {
-    fsPath?: string;
+    workspace: WorkspaceFolder;
     deployedWithConfigs?: boolean;
     configurationTarget?: ConfigurationTarget;
 }
 
-export async function setAppWizardContextDefault(wizardContext: IAppServiceWizardContext & IDeployWizardContext): Promise<void> {
+export async function setAppWizardContextDefault(wizardContext: IAppServiceWizardContext, deployedWorkspace?: WorkspaceFolder): Promise<void> {
```
EricJizbaMSFT: `wizardContext: IAppServiceWizardContext & Partial<IDeployWizardContext>`

and then you should be able to use `wizardContext.workspace`

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,78 @@
...
+import { openUrl } from '../../utils/openUrl';
+import * as workspaceUtil from '../../utils/workspace';
+import { getWorkspaceSetting, updateWorkspaceSetting } from '../../vsCodeConfig/settings';
+import * as tasks from '../../vsCodeConfig/tasks';
+import { IDeployWizardContext } from "../createWebApp/setAppWizardContextDefault";
+
+export async function setPreDeployTaskForDotnet(context: IDeployWizardContext): Promise<void> {
+    await validateDotnetInstalled(context);
+    // follow the publish output patterns, but leave out tfw
+    const dotnetOutputPath: string = path.join('bin', 'Debug', 'publish');
```
EricJizbaMSFT: How were you able to get rid of all tfw logic?

nturinski: If we just rely on `dotnet publish`, we don't need to specify the framework. The output path is something that we are making ourselves.

In a multiple target framework scenario, the publish will fail telling the user to specify their twf

nturinski:
Offline feedback:
Use the web app's siteConfig if we are deploying to determine the site's type.  It should be more reliable than parsing the project.

Look for the .csproj file, but if we can't find it, do the following:
- In Linux case: Request the csproj OR the publish folder
- In Windows case: Don't set up any defaults

Do NOT block users from deploying .NET projects if they don't have .NET core installed.  There's no guarantee that our method of finding .NET is 100% accurate.  Plus they might just have the published bits received from somebody else.

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    // if there was no .csproj in the root space, try to find it within subfolders
+    if (csProjInRoot.length === 0) {
+        // to have a recursive search of the opened workspace, pass in undefined rather than the fsPath
+        csProjInSubfolders = await findFilesByFileExtension(undefined, 'csproj');
+        try {
+            if (csProjInSubfolders.length === 1) {
+                csProjFsPath = csProjInSubfolders[0].fsPath;
+            } else if (csProjInSubfolders.length > 1) {
+                csProjFsPath = (await ext.ui.showQuickPick(mapFilesToQuickPickItems(csProjInSubfolders), { placeHolder: 'Select a .csproj file' })).data;
```
EricJizbaMSFT: It seems out-of-the-blue and not obvious why I'm selecting a csproj file. I would prefer not to prompt for now and just assume you didn't find one. Just add telemetry so that we know how often this happens and if we need to handle it.

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    // assume that the csProj is in the root
+    let csProjFsPath: string = context.workspace.uri.fsPath;
+
+    const csProjInRoot: Uri[] = await findFilesByFileExtension(context.workspace.uri.fsPath, 'csproj');
+    let csProjInSubfolders: Uri[] = [];
+
+    // if there was no .csproj in the root space, try to find it within subfolders
+    if (csProjInRoot.length === 0) {
+        // to have a recursive search of the opened workspace, pass in undefined rather than the fsPath
```
EricJizbaMSFT: This is completely recursive? That could really slow down deployment if you have a big workspace and I don't think we want to do that.

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -91,12 +92,12 @@ async function promptPerformanceWarning(context: IActionContext, asp: AppService
...
 
         if (input === AppServiceDialogResponses.turnOnAdvancedCreation) {
-            await workspaceConfig.update('advancedCreation', true, ConfigurationTarget.Global);
+            await updateGlobalSetting(configurationSettings.advancedCreation, true);
             context.telemetry.properties.cancelStep = AppServiceDialogResponses.turnOnAdvancedCreation.title;
             throw new UserCancelledError();
         } else if (input === DialogResponses.dontWarnAgain) {
             context.telemetry.properties.turnOffPerfWarning = 'true';
-            workspaceConfig.update(showPlanPerformanceWarningSetting, false, ConfigurationTarget.Global);
+            await updateGlobalSetting(showPlanPerformanceWarningSetting, false);
```
EricJizbaMSFT: Any chance you could move these changes to a separate PR? The addition of "../../vsCodeConfig/settings"; affects a lot of files so I find myself re-reviewing it over and over, but it's pretty simple stuff by itself and could've been merged already

src/commands/deploy/getDeployFsPath.ts
```diff
@@ -0,0 +1,73 @@
...
+import * as vscode from 'vscode';
+import { DialogResponses, IActionContext } from 'vscode-azureextensionui';
+import { configurationSettings, extensionPrefix } from '../../constants';
+import { SiteTreeItem } from '../../explorer/SiteTreeItem';
+import { ext } from '../../extensionVariables';
+import { isPathEqual, isSubpath } from '../../utils/pathUtils';
+import * as workspaceUtil from '../../utils/workspace';
+import { getWorkspaceSetting, updateGlobalSetting } from '../../vsCodeConfig/settings';
+
+export async function getDeployFsPath(context: IActionContext, target: vscode.Uri | string | SiteTreeItem | undefined, fileExtension?: string): Promise<string> {
```
EricJizbaMSFT: Also can this be a separate PR? It affects all languages, not just .NET Core - and it's mostly copied so it should be easy to get approved/merged.

EricJizbaMSFT:
Closing because this was split into smaller PRs

Issue #911, opened 2019-03-22
Use detectors to offer insights for failed deployments on Linux


nturinski:
There are also detectors for swapping slots, ip configuration, and configuring custom domains.

PR #1079, opened 2019-07-25
Nat/detector

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/911

So the idea is after a "completed deployment" if validateWebSite returns a 400 or 500 code, then we should check the detectors to see if there is any troubleshooting available to the user.

The types all describe how the data comes back, but since it's primarily supposed to be used as a table, the way I had to get the data is a little unconventional.

Moved to opn because vscode.openUrl does not work for the deeplink that will take them to the portal.

I can't currently test due to portal issues, but will add screenshots when I'm able.


src/validateWebSite.ts
```diff
@@ -88,6 +88,10 @@ export async function validateWebSite(deploymentCorrelationId: string, siteTreeI
             statusCodes.push({ code: currentStatusCode, elapsed: elapsedSeconds });
 
             if (Date.now() > start + maximumValidationMs) {
+
+                if (currentStatusCode !== undefined && currentStatusCode >= 300) {
```
EricJizbaMSFT: `throw new Error();` is too generic. You don't want to use detectors for _any_ error, just status code errors. Maybe just change `validateWebSite` to return the status code.

src/extension.ts
```diff
@@ -9,6 +9,7 @@ import * as vscode from 'vscode';
 import { AppSettingsTreeItem, AppSettingTreeItem, DeploymentsTreeItem, ISiteTreeRoot, registerAppServiceExtensionVariables, SiteClient, stopStreamingLogs } from 'vscode-azureappservice';
 import { AzExtTreeDataProvider, AzureTreeItem, AzureUserInput, callWithTelemetryAndErrorHandling, createApiProvider, createTelemetryReporter, IActionContext, IAzureUserInput, openInPortal, registerCommand, registerEvent, registerUIExtensionVariables } from 'vscode-azureextensionui';
 import { AzureExtensionApiProvider } from 'vscode-azureextensionui/api';
+import { checkLinuxWebAppDownDetector } from './checkLinuxWebAppDownDetector';
```
EricJizbaMSFT: Remove if not used here

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,215 @@
...
+            return;
+        }
+
+        const deployResultTime: Date = new Date(deployment.startTime);
+
+        const detectorOutput: string = `Diagnosing web app "${node.root.client.siteName}" for critical errors...`;
+        ext.outputChannel.appendLine(detectorOutput);
+
+        const detectorUri: string = `https://management.azure.com/subscriptions/${node.root.subscriptionId}/resourceGroups/${node.root.client.resourceGroup}/providers/Microsoft.Web/sites/${node.root.client.siteName}/detectors/${detectorId}`;
+        const requestOptions: WebResource & Partial<{ qs: queryString }> = new WebResource();
```
EricJizbaMSFT: This is the second or third time you've had to workaround the fact that the app service sdk doesn't support something. Can you create a separate utility function to help with these calls? Even if we don't expose it in the shared package, it can at least be a separate file that we can copy/paste between repos.

Also is there a bug to add this to the app service sdk?

For example, [here](https://github.com/microsoft/vscode-azuretools/blob/master/appservice/src/createAppService/SiteRuntimeStep.ts#L59) you correctly used `resourceManagerEndpointUrl`. In the above code, it will fail for other sovereigns.

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,215 @@
...
+        });
+    });
+}
+
+async function getLinuxDetectorError(requestOptions: WebResource, deployResultTime: Date, siteName: string): Promise<string | undefined> {
+    const requestPromise = <(options: RequestOptions | string | URL) => Promise<IncomingMessage>><Function>requestP;
+    const detectorResponse: string = <string><unknown>(await requestPromise(requestOptions));
+    const responseJson: detectorResponseJSON = <detectorResponseJSON>JSON.parse(detectorResponse);
+
+    const datasets: detectorDataset[] = responseJson.properties.dataset;
```
EricJizbaMSFT: Please move all of this parsing logic to a separate file and add unit tests. It's the kind of low-level logic that's easy to mess up and hard to review

src/commands/deploy.ts
```diff
@@ -189,10 +190,17 @@ export async function deploy(context: IDeployWizardContext, confirmDeployment: b
...
             // ignore
         },
-        () => {
-            // ignore
+        async () => {
+            if (!node || !(await node.root.client.getSiteConfig()).linuxFxVersion) {
+                // this currently only works for Linux apps, so don't delay and just exit if it's a Windows app
+                return;
+            }
+            // it can take over 10 minutes for the detectors to appear
```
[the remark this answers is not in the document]
nturinski: Well, I have a delay before I even start checking, and then check it for 10 more minutes (because who knows when the heck it'll up ready to view.)  I'm fine with polling for it more often, but it just seemed pointless to start polling because I know it's gonna be a while.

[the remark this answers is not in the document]
EricJizbaMSFT: Well that's my question - does it take _at least_ 10 minutes? That's different than "it can take over 10 minutes".  If so I'm fine with this delay but I would clarify the comment in the code.

src/utils/tableUtil.ts
```diff
@@ -0,0 +1,54 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ColumnName, detectorDataset, detectorTable } from "../checkLinuxWebAppDownDetector";
+
+// the dataset contains several tables all with the same tableName and columnNames so to find the proper table, look for a specific value
+export function findTableByRowValue(datasets: detectorDataset[], searchValue: string): detectorTable | undefined {
```
EricJizbaMSFT: Feels like it would be better to have all the detector logic in one folder, rather than having a file like this in `utils`. As it is now, I would have no idea this file is for detectors based on the naming

`validateWebsite` could probably go the same folder, too.

test/parseDetectorTables.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    test('Find table by row value', async () => {
+        const table: detectorTable | undefined = findTableByRowValue(detectorDataset.properties.dataset, vsCodeIntegration);
+        assert.equal(table, detectorDataset.properties.dataset[1].table);
+    });
+
+    test('Get values by column name', async () => {
+        const timestamp: string = "7/26/2019 8:40:00 PM UTC ";
+        const table: detectorTable | undefined = findTableByRowValue(detectorDataset.properties.dataset, vsCodeIntegration);
+        assert.ok(table);
+        const values: string[] = getValuesByColumnName(nonNullValue(table), ColumnName.value);
```
EricJizbaMSFT: How about a unit test for `getValuesByColumnName` for a column that returns multiple values

test/parseDetectorTables.test.ts
```diff
@@ -0,0 +1,26 @@
...
+const vsCodeIntegration: string = 'Latest time seen by detector. To be used in VSCode integration.';
+suite('Detector Dataset Parser', () => {
```
EricJizbaMSFT: I like to name the file and suite the same as the function it's testing. This is a good example: https://github.com/microsoft/vscode-azuretools/blob/master/appservice/test/convertStacksToPicks.test.ts#L472

FWIW I like `parseDetectorTables` better than `tableUtils`

test/parseDetectorTables.test.ts
```diff
@@ -0,0 +1,26 @@
...
...nce", dataType: "String", columnType: null }, { columnName: "Facility", dataType: "String", columnType: null }, { columnName: "FailureCount", dataType: "Int64", columnType: null }], rows: [["2019-07-26T20:40:00", "467983_Small_90", "naturins-linux", 1]] }, renderingProperties: { defaultValue: 0, graphType: 1, graphOptions: null, timestampColumnName: null, roleInstanceColumnName: null, seriesColumns: null, type: 2, title: "Number of app Start Failures", description: null } }, { table: { tableName: "", columns: [{ columnName: "Time", dataType: "DateTime", columnType: null }, { columnName: "Instance", dataType: "String", columnType: null }, { columnName: "Facility", dataType: "String", columnType: null }, { columnName: "Message", dataType: "String", columnType: null }, { columnName: "FailureCount", dataType: "Int64", columnType: null }], rows: [["2019-07-26T20:40:00", "467983_Small_90", "naturins-linux", "docker container could not be started: naturins-linux_1", 1]] }, renderingProperties: { displayColumnNames: null, groupByColumnName: null, type: 1, title: null, description: null } }], status: { message: null, statusId: 0 }, dataProvidersMetadata: null, suggestedUtterances: null } };
```
EricJizbaMSFT: Can you add some space and formatting to this? Also maybe move it below the suite (If you get complaints about "no-use-before-declare", just turn the rule off. It's actually [discouraged](https://palantir.github.io/tslint/rules/no-use-before-declare/))

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        const detectorUri: string = `${node.root.environment.resourceManagerEndpointUrl}${node.id}/detectors/${detectorId}`;
+        const requestOptions: WebResource & Partial<{ qs: queryString }> = new WebResource();
+
+        requestOptions.method = 'GET';
+        requestOptions.url = detectorUri;
+        requestOptions.qs = {
+            'api-version': "2015-08-01",
+            fId: "1",
+            btnId: "2",
```
EricJizbaMSFT: What are these random number values? Please add a comment in the code

src/checkLinuxWebAppDownDetector.ts
EricJizbaMSFT: Still waiting on this. As an aside, maybe try `ServiceClient.sendRequest`

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+                throw new UserCancelledError();
+            }
+
+            if (Date.now() > detectorTimeoutMs) {
+                const noIssuesFound: string = `No critical issues found for web app "${node.root.client.siteName}".`;
+                ext.outputChannel.appendLine(noIssuesFound);
+                return undefined;
+            }
+            // update the new end time to be now
+            requestOptions.qs.endTime = new Date().toISOString();
```
EricJizbaMSFT: You have to specify end time? You can't just leave it empty to get latest?

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+        // wait 10 minutes for a response from the detector
+        const detectorTimeoutMs: number = Date.now() + 20 * 60 * 1000;
+        do {
+            if (tokenSource.token.isCancellationRequested) {
+                // the user cancelled the check by deploying again
+                throw new UserCancelledError();
+            }
+
+            if (Date.now() > detectorTimeoutMs) {
+                const noIssuesFound: string = `No critical issues found for web app "${node.root.client.siteName}".`;
```
EricJizbaMSFT: This message feels misleading. I'd just say we timed out or something

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+            }
+            // update the new end time to be now
+            requestOptions.qs.endTime = new Date().toISOString();
+            detectorErrorMessage = await getLinuxDetectorError(requestOptions, deployResultTime, node.root.client.fullName);
+            // poll every minute
+            await delay(1000 * 60);
+        } while (!detectorErrorMessage);
+
+        await ext.ui.showWarningMessage(detectorErrorMessage, { title: 'Open in Portal' });
+        const portalDeeplink = `${node.root.environment.portalUrl}/?websitesextension_ext=asd.featurePath%3Ddetectors%2F${detectorId}#@microsoft.onmicrosoft.com/resource/${node.root.client.id}/troubleshoot`;
```
EricJizbaMSFT: Please use `openInPortal` from the UI package

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+        const errorMessages: string[] = getValuesByColumnName(criticalErrorTable, ColumnName.value);
+        const failureCount: string = getValuesByColumnName(failureCountTable, ColumnName.failureCount)[0];
+
+        return `Critical insights found for "${siteName}": ${failureCount} as of ${new Date(timestamp).toLocaleString()}. Critical insight preview: "${errorMessages[0]}". Click here to access "App Service Diagnostics" for recommendations.`;
+    }
+
+    return undefined;
+}
+
+export function validateTimestamp(detectorTime: Date, deployResultTime: Date): boolean {
```
EricJizbaMSFT: So this checks that the detector and deploy result time are within a minute of each other? Can you explain why it does that? Seems like it should pick the first detector that exists after the deploy, regardless of how close in time they are

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+    const datasets: detectorDataset[] = responseJson.properties.dataset;
+    const vsCodeIntegration: string = 'Latest time seen by detector. To be used in VSCode integration.';
+    const timestampTable: detectorTable | undefined = findTableByRowValue(datasets, vsCodeIntegration);
+    if (!timestampTable) {
+        // if we can't find the timestamp, exit and try again
+        return undefined;
+    }
+
+    const timestamp = new Date(getValuesByColumnName(timestampTable, ColumnName.value)[0]);
+    if (validateTimestamp(timestamp, deployResultTime)) {
```
EricJizbaMSFT: Add a unit test validating it doesn't return anything when the timestamp is not valid

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    const datasets: detectorDataset[] = responseJson.properties.dataset;
+    const vsCodeIntegration: string = 'Latest time seen by detector. To be used in VSCode integration.';
+    const timestampTable: detectorTable | undefined = findTableByRowValue(datasets, vsCodeIntegration);
+    if (!timestampTable) {
+        // if we can't find the timestamp, exit and try again
+        return undefined;
+    }
+
+    const timestamp = new Date(getValuesByColumnName(timestampTable, ColumnName.value)[0]);
```
EricJizbaMSFT: Add at least one unit test that validates essentially all of `getLinuxDetectorError` except the request part. Aka given `datasets`, it will return these errors and these failure counts

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+                return undefined;
+            }
+            // update the new end time to be now
+            requestOptions.qs.endTime = new Date().toISOString();
+            detectorErrorMessage = await getLinuxDetectorError(requestOptions, deployResultTime, node.root.client.fullName);
+            // poll every minute
+            await delay(1000 * 60);
+        } while (!detectorErrorMessage);
+
+        await ext.ui.showWarningMessage(detectorErrorMessage, { title: 'Open in Portal' });
```
EricJizbaMSFT: Per offline discussion, we're gonna either hide this message for now or add a feature flag. At least until more people on our team (especially Matt) can try it out and see how they feel about the experience

src/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,175 @@
...
+    solutions = "Solutions",
+    time = "Time",
+    instance = "Instance",
+    facility = "Facility",
+    failureCount = "FailureCount"
+}
+
+export const detectorCancelTokens: Map<string, CancellationTokenSource> = new Map();
+export async function checkLinuxWebAppDownDetector(node: SiteTreeItem, tokenSource: CancellationTokenSource): Promise<void> {
+    return await callWithTelemetryAndErrorHandling('appService.linuxWebAppDownDetector', async (context: IActionContext): Promise<void> => {
```
EricJizbaMSFT: Probably want `context.errorHandling.suppressDisplay` set to true

src/checkLinuxWebAppDownDetector.ts
nturinski: Yeah, I asked the team working on this and he responded:

> If you do not provide a timestamp, it takes last 1 day... I think it ignores the second's but considers timestamp up to a minute's granularity by default

It's also how the portal does it, though their end time is the current time and their start time is exactly 24 hours prior:
image

src/checkLinuxWebAppDownDetector.ts
nturinski: This is resolved by Eric's changes in the azuretools https://github.com/microsoft/vscode-azuretools/commit/1d23d4f2c8126b9ff1ab7d70f8f1ca7c0266c9a9

src/checkLinuxWebAppDownDetector.ts
nturinski: Unrelated, but in this PR, you have a comment regarding unconventional portal URLs and I'd like to pull that same card 😓

EricJizbaMSFT: This is what openInPortal does:
```
export async function openInPortal(root: types.ISubscriptionContext, id: string, options?: types.OpenInPortalOptions): Promise<void> {
    const queryPrefix: string = (options && options.queryPrefix) ? `?${options.queryPrefix}` : '';
    const url: string = `${root.environment.portalUrl}/${queryPrefix}#@${root.tenantId}/resource${id}`;

    await openUrl(url);
}
```

Can't you pass `websitesextension_ext=asd.featurePath%3Ddetectors%2F${detectorId}` as the query and `${node.root.client.id}/troubleshoot` as the id?

My url looks like this:
```
${treeItem.root.environment.portalUrl}/#blade/AppInsightsExtension/QuickPulseBladeV2/ComponentId/${componentId}/ResourceId/${resourceId}
```

It doesn't have `#@microsoft.onmicrosoft.com/resource` like yours does. You have a bug because you're not using tenantId (you always use "microsoft.onmicrosoft.com" as the tenant). That's why I wanted you to use the shared one

PR #1080, opened 2019-07-25
Add util code for workspace settings and deployFsPath

nturinski:
Splitting up this PR: https://github.com/microsoft/vscode-azureappservice/pull/1060 into smaller components.

### DIFF
--- a/src/commands/createWebApp/setAppWizardContextDefault.ts
+++ b/src/commands/createWebApp/setAppWizardContextDefault.ts
@@ -11,6 +11,7 @@ import { IActionContext, LocationListStep } from 'vscode-azureextensionui';
 import { configurationSettings, extensionPrefix } from '../../constants';
 import { javaUtils } from '../../utils/javaUtils';
 import { findFilesByFileExtension, getContainingWorkspace } from '../../utils/workspace';
+import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
 
 export interface IDeployWizardContext extends IActionContext {
     fsPath?: string;
@@ -24,9 +25,10 @@ export async function setAppWizardContextDefault(wizardContext: IAppServiceWizar
     const workspaceForRecommendation: WorkspaceFolder | undefined = wizardContext.fsPath ?
         getContainingWorkspace(wizardContext.fsPath) : workspace.workspaceFolders && workspace.workspaceFolders.length === 1 ?
             workspace.workspaceFolders[0] : undefined;
+    let fsPath: string | undefined;
 
     if (workspaceForRecommendation) {
-        const fsPath: string = workspaceForRecommendation.uri.fsPath;
+        fsPath = workspaceForRecommendation.uri.fsPath;
 
         if (await fse.pathExists(path.join(fsPath, 'package.json'))) {
             wizardContext.recommendedSiteRuntime = [LinuxRuntimes.node];
@@ -49,8 +51,7 @@ export async function setAppWizardContextDefault(wizardContext: IAppServiceWizar
         }
     }
 
-    const workspaceConfig: WorkspaceConfiguration = workspace.getConfiguration(extensionPrefix);
-    const advancedCreation: boolean | undefined = workspaceConfig.get(configurationSettings.advancedCreation);
+    const advancedCreation: boolean | undefined = getWorkspaceSetting(configurationSettings.advancedCreation, fsPath);
 
     if (!advancedCreation) {
         if (!wizardContext.location) {
--- a/src/commands/createWebApp/setAppWizardContextDefault.ts
+++ b/src/commands/createWebApp/setAppWizardContextDefault.ts
@@ -12,12 +12,7 @@ import { configurationSettings, extensionPrefix } from '../../constants';
 import { javaUtils } from '../../utils/javaUtils';
 import { findFilesByFileExtension, getContainingWorkspace } from '../../utils/workspace';
 import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
-
-export interface IDeployWizardContext extends IActionContext {
-    fsPath?: string;
-    deployedWithConfigs?: boolean;
-    configurationTarget?: ConfigurationTarget;
-}
+import { IDeployWizardContext } from '../deploy/deploy';
 
 export async function setAppWizardContextDefault(wizardContext: IAppServiceWizardContext & IDeployWizardContext): Promise<void> {
     // if the user entered through "Deploy", we'll have a project to base our recommendations on
--- a/src/commands/createWebApp/setAppWizardContextDefault.ts
+++ b/src/commands/createWebApp/setAppWizardContextDefault.ts
@@ -15,8 +15,8 @@ import { IDeployWizardContext } from '../deploy/deploy';
 export async function setAppWizardContextDefault(wizardContext: IAppServiceWizardContext & Partial<IDeployWizardContext> & Partial<ICreateChildImplContext>): Promise<void> {
     // if the user entered through "Deploy", we'll have a project to base our recommendations on
     // otherwise, look at their current workspace and only suggest if one workspace is opened
-    const workspaceForRecommendation: WorkspaceFolder | undefined = wizardContext.fsPath ?
-        getContainingWorkspace(wizardContext.fsPath) : workspace.workspaceFolders && workspace.workspaceFolders.length === 1 ?
+    const workspaceForRecommendation: WorkspaceFolder | undefined = wizardContext.deployFsPath ?
+        getContainingWorkspace(wizardContext.deployFsPath) : workspace.workspaceFolders && workspace.workspaceFolders.length === 1 ?
             workspace.workspaceFolders[0] : undefined;
     let fsPath: string | undefined;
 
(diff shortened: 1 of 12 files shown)


src/vsCodeConfig/settings.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export async function updateGlobalSetting<T = string>(section: string, value: T, prefix: string = extensionPrefix): Promise<void> {
+    const projectConfiguration: WorkspaceConfiguration = workspace.getConfiguration(prefix);
+    await projectConfiguration.update(section, value, ConfigurationTarget.Global);
+}
+
+/**
+ * Uses extensionPrefix 'appService' unless otherwise specified
+ */
+export async function updateWorkspaceSetting<T = string>(section: string, value: T, fsPath: string, prefix: string = extensionPrefix): Promise<void> {
```
EricJizbaMSFT: Would be nice if this PR also had changes like this one:
https://github.com/microsoft/vscode-azureappservice/pull/1060/files#diff-caaef5e8ec24b167d4d47bee5dec061bL31

It's a very simple change - it's related to adding this file - and it happens in quite a few files

nturinski: Will do after advancedCreation gets in since a lot of workspace setting stuff deals with that.

src/commands/deploy/deploy.ts
```diff
@@ -150,7 +147,7 @@ export async function deploy(context: IDeployWizardContext, confirmDeployment: b
                 window.showTextDocument(doc);
             }
 
-            await workspaceConfig.update(constants.configurationSettings.defaultWebAppToDeploy, '', context.configurationTarget);
+            await updateWorkspaceSetting(constants.configurationSettings.defaultWebAppToDeploy, '', context.fsPath);
```
EricJizbaMSFT: I think there's a bug with this one. Passing `context.configurationTarget` is def different than fspath

src/commands/deploy/getDeployFsPath.ts
```diff
@@ -0,0 +1,73 @@
...
+import { getWorkspaceSetting, updateGlobalSetting } from '../../vsCodeConfig/settings';
+
+export async function getDeployFsPath(context: IActionContext, target: vscode.Uri | string | SiteTreeItem | undefined, fileExtension?: string): Promise<string> {
+    context.telemetry.properties.deploymentEntryPoint = target ? 'webAppContextMenu' : 'deployButton';
+    if (target instanceof vscode.Uri) {
+        context.telemetry.properties.deploymentEntryPoint = 'fileExplorerContextMenu';
+        return await appendDeploySubpathSetting(target.fsPath);
+    } else if (typeof target === 'string') {
+        // not sure where this entry point would be
+        return await appendDeploySubpathSetting(target);
```
EricJizbaMSFT: lol

It's for the api
https://github.com/microsoft/vscode-azurefunctions/wiki/Azure-Functions-API-(Preview)#deploy

App service could probably remove the `string` case

src/commands/deploy/getDeployFsPath.ts
```diff
@@ -0,0 +1,73 @@
...
+        const deploySubPath: string | undefined = getWorkspaceSetting(configurationSettings.deploySubpath, targetPath);
+        if (deploySubPath) {
+            if (vscode.workspace.workspaceFolders.some(f => isPathEqual(f.uri.fsPath, targetPath))) {
+                return path.join(targetPath, deploySubPath);
+            } else {
+                const folder: vscode.WorkspaceFolder | undefined = vscode.workspace.workspaceFolders.find(f => isSubpath(f.uri.fsPath, targetPath));
+                if (folder) {
+                    const fsPathWithSetting: string = path.join(folder.uri.fsPath, deploySubPath);
+                    if (!isPathEqual(fsPathWithSetting, targetPath)) {
+                        const settingKey: string = 'showDeploySubpathWarning';
```
EricJizbaMSFT: Add this setting to package.json. Should be able to copy from functions

src/commands/getDefaultWebAppToDeploy.ts
```diff
@@ -27,7 +20,8 @@ export async function getDefaultWebAppToDeploy(context: IDeployWizardContext): P
             return <WebAppTreeItem>defaultWebApp;
         } else {
             // if defaultPath or defaultNode cannot be found or there was a mismatch, delete old settings and prompt to save next deployment
-            workspaceConfig.update(configurationSettings.defaultWebAppToDeploy, undefined, context.configurationTarget);
+            context.fsPath ? await updateWorkspaceSetting(configurationSettings.defaultWebAppToDeploy, undefined, context.fsPath) :
```
EricJizbaMSFT: Yeah double check `context.configurationTarget` for this file, too. I don't understand how you were refactoring, but I don't think it's correct

src/commands/deploy/deploy.ts
nturinski: The configurationTarget was to determine whether or not it was workspace or global specific.  With updateWorkspaceSetting, if undefined is passed for the fsPath, it will default to global effectively making the configurationTarget pointless.

nturinski:
Regarding the recursive search, it doesn't seem to take very long (less than a second).  I have a pretty big temp folder with some recursive structure.

image

src/vsCodeConfig/settings.ts
```diff
@@ -0,0 +1,62 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { ConfigurationTarget, Uri, workspace, WorkspaceConfiguration } from "vscode";
+import { extensionPrefix } from '../constants';
+
+/**
+ * Uses extensionPrefix 'appService' unless otherwise specified
+ */
+export async function updateGlobalSetting<T = string>(section: string, value: T, prefix: string = extensionPrefix): Promise<void> {
+    const projectConfiguration: WorkspaceConfiguration = workspace.getConfiguration(prefix);
```
EricJizbaMSFT: I'm starting to get pretty annoyed that this setting stuff is still in an open PR ...the logic is so simple, affects so many files, is unrelated to `getDeployFsPath`, and could've easily been merged weeks ago if it was submitted as a PR by itself

src/commands/deploy/deploy.ts
```diff
@@ -0,0 +1,186 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as WebSiteModels from 'azure-arm-website/lib/models';
```
EricJizbaMSFT: Would've preferred a separate PR to move the deploy file. Now I can't easily see the diff

src/utils/workspace.ts
```diff
@@ -86,8 +72,8 @@ export function getContainingWorkspace(fsPath: string): vscode.WorkspaceFolder |
 }
 
 export async function findFilesByFileExtension(fsPath: string | undefined, fileExtension: string): Promise<vscode.Uri[]> {
-    // if there is a fsPath, then only check the folder for the file extension, otherwise use all currently opened workspaces
-    const relativeDirectory: vscode.RelativePattern | string = fsPath ? new vscode.RelativePattern(fsPath, `*.${fileExtension}`) : path.join('**', `*.${fileExtension}`);
+    // if there is a fsPath, then only check the folder for the file extension, otherwise recursively check all currently opened workspaces
+    const relativeDirectory: vscode.RelativePattern | string = fsPath ? new vscode.RelativePattern(fsPath, `*.${fileExtension}`) : `**/*.{${fileExtension}}`;
```
EricJizbaMSFT: > Regarding the recursive search, it doesn't seem to take very long (less than a second). I have a pretty big temp folder with some recursive structure.

I will continue to stress this any time we're talking performance. You can test _relative_ performance (aka is one option slower than the other) with your default setup, but you can't test absolute performance (is this option too slow). Get an old machine with bad specs and slow internet and you _might_ be able to make an argument, but ultimately we just have to follow best practices (especially with any operation involving the file system or internet), and make decisions based on telemetry

nturinski:
I'm not trying to frustrate you, I'm just having a hard time distinguishing what to/not to include because it's all kind of intertwined.  

I thought you requested a PR for all the util files I copied over which includes settings/deployFsPath, but then with deployFsPath, it didn't make sense until I added the deploy logic, which touches other stuff.  And then it becomes too complicated again.

So I'll break the PRs up in this order:

- Settings files + update settings
- Move deploy file
- Change workspace settings files
- Add deployFsPath with new deploy logic

src/commands/createWebApp/setAppWizardContextDefault.ts
```diff
@@ -40,19 +36,11 @@ export async function setAppWizardContextDefault(wizardContext: IAppServiceWizar
                 LinuxRuntimes.tomcat,
                 LinuxRuntimes.wildfly
             ];
-
-            // considering high resource requirement for Java applications, a higher plan sku is set here
-            wizardContext.newPlanSku = { name: 'P1v2', tier: 'PremiumV2', size: 'P1v2', family: 'P', capacity: 1 };
```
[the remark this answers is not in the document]
nturinski: Agreed, but I had to move it because that logic is determining recommended runtimes based on projects.  But if you create a Java app with no project opened, you don't get the defaults we recommend.  

We can determine the new runtime by this point so we should leverage that information.

EricJizbaMSFT:
Closing because this was split into smaller PRs

PR #1081, opened 2019-07-26
Support new advancedCreation

nturinski:
Have an advanced create command rather than a setting.

Any prompt that alerted users about advanced creation has been replaced by directly calling advancedCreate for them.  I reset the context except for siteName and runtime so that the user does not have to re-enter their details.

The only hiccup with this is if they create it via deploy, but after the creation, it will include the deploy button when it has finished creating.

### DIFF
--- a/src/commands/createWebApp/setAppWizardContextDefault.ts
+++ b/src/commands/createWebApp/setAppWizardContextDefault.ts
@@ -5,10 +5,9 @@
 
 import * as fse from 'fs-extra';
 import * as path from 'path';
-import { ConfigurationTarget, workspace, WorkspaceConfiguration, WorkspaceFolder } from 'vscode';
+import { ConfigurationTarget, workspace, WorkspaceFolder } from 'vscode';
 import { IAppServiceWizardContext, LinuxRuntimes, WebsiteOS } from 'vscode-azureappservice';
-import { IActionContext, LocationListStep } from 'vscode-azureextensionui';
-import { configurationSettings, extensionPrefix } from '../../constants';
+import { IActionContext, ICreateChildImplContext, LocationListStep } from 'vscode-azureextensionui';
 import { javaUtils } from '../../utils/javaUtils';
 import { findFilesByFileExtension, getContainingWorkspace } from '../../utils/workspace';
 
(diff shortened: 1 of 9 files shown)


src/explorer/SubscriptionTreeItem.ts
```diff
@@ -101,22 +100,22 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
-                const message: string = `Modify the setting "${extensionPrefix}.${configurationSettings.advancedCreation}" if you want to change the default values when creating a Web App in Azure.`;
+            if (!parseError(error).isUserCancelledError && !context.advancedCreation) {
 
+                const message: string = `Use advanced create to change the default values when creating a Web App in Azure.`;
+                const btn: MessageItem = { title: 'Create Web App... (Advanced)' };
                 // tslint:disable-next-line: no-floating-promises
-                ext.ui.showWarningMessage(message, AppServiceDialogResponses.turnOnAdvancedCreation).then(async result => {
-                    if (result === AppServiceDialogResponses.turnOnAdvancedCreation) {
-                        await workspaceConfig.update('advancedCreation', true, ConfigurationTarget.Global);
+                ext.ui.showWarningMessage(message, btn).then(async result => {
```
EricJizbaMSFT: > I reset the context except for siteName and runtime so that the user does not have to re-enter their details.

Not really sold on this logic. I would expect a "fresh" create to start and I would re-enter the app name, etc. For example, I wonder if there's a situation where the wizard partially creates the site, it fails, the user clicks "Create Web App... (Advanced)" and then it fails again because you re-used the name of a partially failed site. Users can handle re-entering that info

src/commands/createWebApp/setAppWizardContextDefault.ts
```diff
@@ -72,3 +68,14 @@ export async function setAppWizardContextDefault(wizardContext: IAppServiceWizar
         }
     }
 }
+
+export function resetAppWizardContextDefaults(context: Partial<IAppServiceWizardContext>): void {
+    context.plan = undefined;
```
EricJizbaMSFT: I wonder if instead of individually setting these to undefined (which we would have to maintain going forward if we add more properties), you could just create a clone of `context` before calling `setAppWizardContextDefault` and then use that for advanced create

src/explorer/SubscriptionTreeItem.ts
nturinski: I'll create an issue and ask @fiveisprime to weigh in on it when he returns.

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -101,31 +98,14 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
-                });
-            }
-            throw error;
-        }
-
+        await wizard.execute();
         context.telemetry.properties.os = wizardContext.newSiteOS;
         context.telemetry.properties.runtime = wizardContext.newSiteRuntime;
-        context.telemetry.properties.advancedCreation = advancedCreation ? 'true' : 'false';
+        context.telemetry.properties.advancedCreation = context.advancedCreation ? 'true' : 'false';
```
EricJizbaMSFT: You can remove this line. It's already in the shared package: https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/treeDataProvider/AzExtParentTreeItem.ts#L67

Issue #1054, opened 2019-06-25
Deploy to azure is asking for .jar file after upgrade to 0.15.0


TeamEliteSports:
I was about to deploy to an Azure App Service but while searching for the project output/publish folder the navigation was disabled, but I could navigate writing the path directly in the navigation window, then it throw an error `ProjectName.jar` was not found, then I noticed that was asking for an Artifact.

installing past `0.14.0` version solved the problem

using `Manjaro` OS and `pcmanfm` as file explorer
also the same problem was found on `Ubuntu` OS and `pantheon` as file explorer

any help appreciated

epgeroy:
+1

luiarhs:
+2 the same thing

nturinski:
@TeamEliteSports

Thanks for reporting and sorry about this issue.  What kind of project are you trying to deploy?  There was a recent change where if the selected app is a Linux Java runtime, it will automatically prompt you for a `.jar` file.

Also, how are you trying to deploy?  Are you right-clicking the app or are you clicking the blue deploy button?  If you need to deploy a specific folder, you can use the File Explorer view and right-click a `.zip` file or a folder and click `Deploy to Web App.`

epgeroy:
is a .net core 2.2 app.
nothing to do with java

I'm trying to deploy hitting `Ctrl-P` then `Deploy to Web App` it guides me to select the correct AppService and the correct folder. Also the extension is updating itself so every time I close and reopen VSCode it happens again

thxs for the quick response

nturinski:
@epgeroy Thanks for reporting and providing details.  I'll investigate what's causing this.  In the meantime, as a workaround, you can try right-clicking the folder that you intend to deploy.

Sorry about this issue!  Will try to resolve with our next release.

PR #1087, opened 2019-07-31
Add workspace changes to enforce workspace for deploy

nturinski:
Might fix? https://github.com/microsoft/vscode-azureappservice/issues/1054, but need to test on a Linux machine.

Moved Java default logic because it is relying on a Java project being opened at creation, but we can detect what user the runtime selected AFTER prompt, but BEFORE create, so it makes sense to set it based on the runtime.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -9,7 +9,7 @@ import { pathExists } from 'fs-extra';
 import * as path from 'path';
 import { commands, Disposable, MessageItem, Uri, window, workspace, WorkspaceFolder } from 'vscode';
 import * as appservice from 'vscode-azureappservice';
-import { DialogResponses, IAzureQuickPickItem, parseError } from 'vscode-azureextensionui';
+import { DialogResponses, IActionContext, parseError } from 'vscode-azureextensionui';
 import * as constants from '../../constants';
 import { SiteTreeItem } from '../../explorer/SiteTreeItem';
 import { WebAppTreeItem } from '../../explorer/WebAppTreeItem';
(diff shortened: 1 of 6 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -93,34 +103,25 @@ export async function deploy(context: IDeployWizardContext, confirmDeployment: b
     const correlationId = getRandomHexString();
     context.telemetry.properties.correlationId = correlationId;
 
+    // if we already got siteConfig, don't waste time getting it again
     siteConfig = siteConfig ? siteConfig : await node.root.client.getSiteConfig();
 
     if (javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
-        const javaArtifactFiles: Uri[] = await workspaceUtil.findFilesByFileExtension(context.fsPath, javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion));
```
EricJizbaMSFT: Revert these java changes as well

src/commands/deploy/deploy.ts
```diff
@@ -27,42 +26,53 @@ import { startStreamingLogs } from '../startStreamingLogs';
...
-        context.fsPath = target.fsPath;
+        context.deployFsPath = target.fsPath;
         context.telemetry.properties.deploymentEntryPoint = 'fileExplorerContextMenu';
     } else {
         context.telemetry.properties.deploymentEntryPoint = target ? 'webAppContextMenu' : 'deployButton';
-        node = target;
+        if (target) {
+            node = target;
+            siteConfig = await node.root.client.getSiteConfig();
+            if (javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
```
EricJizbaMSFT: Revert these java changes as well

PR #1101, opened 2019-08-15
Remove logpoints code

### LINKED ISSUE #1100: Remove Logpoints related code
This is being handled by VS Code when the user has started remote debugging.

nturinski:
As I understand it, Logpoints was implemented to be used with remote debugging.  But it seems like VS Code is handling that natively so I don't think we need this code anymore?

I've verified that remote debugging and breakpoints still work without any of this code, though to be honest, I'm not entirely to verify that logpoints is working as intended.

Fixes #1100


fiveisprime:
> ...though to be honest, I'm not entirely to verify that logpoints is working as intended.

When you're remote debugging an app, right-click in the gutter next to some code and choose "Add Logpoint..."

image

Enter a message with some interpolation and hit enter. You'll get a red diamond rather than a circle for a breakpoint.

image

When the logpoint is hit, the message will show in the debug console rather than breaking.

image

nturinski:
Got it.  Can confirm logpoints still works after removing this code.

image

package.json
```diff
@@ -889,72 +819,6 @@
...
-                        "trace": true,
-                        "hostname": "",
-                        "instanceId": "",
-                        "sessionId": "",
-                        "debugId": ""
-                    }
-                ]
-            }
         ]
     },
```
EricJizbaMSFT: Please double check the `dependencies` and `devDependencies` to see if there are any that can be removed. I believe at least `vscode-debugadapter` and `vscode-debugprotocol` could be removed

src/extension.ts
```diff
@@ -84,26 +79,6 @@ export async function activateInternal(
...
-        const provider = new LoadedScriptsProvider(context);
-        context.subscriptions.push(vscode.window.registerTreeDataProvider('appservice.loadedScriptsExplorer.jsLogpoints', provider));
-
-        const documentProvider = new RemoteScriptDocumentProvider();
-        context.subscriptions.push(vscode.workspace.registerTextDocumentContentProvider(RemoteScriptSchema.schema, documentProvider));
-
-        const logPointsManager = new LogPointsManager();
-        context.subscriptions.push(logPointsManager);
-
-        const pathIcon = context.asAbsolutePath('resources/logpoint.svg');
```
EricJizbaMSFT: Remove this svg from `resources` folder

Issue #1077, opened 2019-07-24
"Don't show again" dialog on App deployment doesn't work


plmcgrn:
When deploying an App to Azure, the "Always deploy to workspace" dialog has a "don't show again" option.  Clicking this option doesn't actually suppress it on future deployments.  For folks running separate Azure target environments where we want to deploy to Staging, test, and then deploy again to Production, this dialog gets annoying.

![image](https://user-images.githubusercontent.com/889547/61826171-e4fd1d00-ae2f-11e9-84b6-abadda45053d.png)

nturinski:
Hi @plmcgrn,

Thanks for reporting.

Are you deploying the same project to different slots?  This setting is workspace specific, so it would only stifle the prompt for that specific project.

If this is the same project everytime, then let me know because that's definitely a bug!

plmcgrn:
> Hi @plmcgrn,
> 
> Thanks for reporting.
> 
> Are you deploying the same project to different slots? This setting is workspace specific, so it would only stifle the prompt for that specific project.
> 
> If this is the same project every time, then let me know because that's definitely a bug!

It's the same one, every time, for now.  I'm essentially doing repeated app deployments from the same project to the same Azure app, for now.  Once I'm comfy that my staging app on Azure is stable, I'll be deploying the same project to a second app, production, on Azure.

nturinski:
Alright, thanks for responding.  I'll look into this and hopefully have it resolved soon.

One other question, but do you have the workspace opened when you deploy?  Do you have a multi-root workspace or only your one project opened?

plmcgrn:
Hi,
I do have a workspace open, and for this, it's a single project within it.

PR #1103, opened 2019-08-17
Check for DefaultWebApp setting before prompting

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1077

The previous check didn't work because `deployWithConfigs` was set to false when there wasn't a default web app.  This didn't account for the `none` behavior.

I wanted to change deployWithConfigs to include a `none` type, but `getDefaultWebApp` doesn't get called if the user entered via the context menu (since we already have a node) so I have to check the workspace setting instead.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -158,7 +158,8 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         deployContext.telemetry.properties.cancelStep = '';
     }
 
-    if (!deployContext.deployedWithConfigs) {
+    const defaultWebAppToDeploySetting: string | undefined = getWorkspaceSetting(constants.configurationSettings.defaultWebAppToDeploy, deployFsPath);
+    if (!defaultWebAppToDeploySetting) {
         // tslint:disable-next-line:no-floating-promises
         node.promptToSaveDeployDefaults(deployContext.workspace.uri.fsPath, deployContext.deployFsPath, deployContext);
     }
--- a/src/commands/getDefaultWebAppToDeploy.ts
+++ b/src/commands/getDefaultWebAppToDeploy.ts
@@ -11,8 +11,9 @@ import { IDeployWizardContext } from "./deploy/IDeployWizardContext";
 
 export async function getDefaultWebAppToDeploy(context: IDeployWizardContext): Promise<WebAppTreeItem | undefined> {
     const defaultWebAppId: string | undefined = getWorkspaceSetting(configurationSettings.defaultWebAppToDeploy, context.workspace.uri.fsPath);
-
-    if (defaultWebAppId && defaultWebAppId !== none) {
+    if (defaultWebAppId === none) {
+        context.telemetry.properties.deployedWithConfigs = 'none';
+    } else if (defaultWebAppId) {
         const defaultWebApp: WebAppTreeItem | undefined = await ext.tree.findTreeItem(defaultWebAppId, context); // resolves to undefined if app can't be found
         if (defaultWebApp) {
             context.deployedWithConfigs = true;


src/commands/deploy/deploy.ts
```diff
@@ -158,7 +158,8 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         deployContext.telemetry.properties.cancelStep = '';
     }
 
-    if (!deployContext.deployedWithConfigs) {
+    const defaultWebAppToDeploySetting: string | undefined = getWorkspaceSetting(constants.configurationSettings.defaultWebAppToDeploy, deployFsPath);
```
EricJizbaMSFT: > I wanted to change deployWithConfigs to include a none type, but getDefaultWebApp doesn't get called if the user entered via the context menu (since we already have a node) so I have to check the workspace setting instead.

We usually have a separate setting like `appService.showDefaultWebAppPrompt` that is checked regardless of the entry point. Instead, the `none` value essentially mashes the `defaultWebAppToDeploy` setting with `showDefaultWebAppPrompt` which is why you ran into that problem. That being said I wouldn't want to change this without discussing backwards compatibility.

src/commands/getDefaultWebAppToDeploy.ts
```diff
@@ -11,8 +11,9 @@ import { IDeployWizardContext } from "./deploy/IDeployWizardContext";
 
 export async function getDefaultWebAppToDeploy(context: IDeployWizardContext): Promise<WebAppTreeItem | undefined> {
```
EricJizbaMSFT: This file should be in the `deploy` folder

src/commands/getDefaultWebAppToDeploy.ts
```diff
@@ -11,8 +11,9 @@ import { IDeployWizardContext } from "./deploy/IDeployWizardContext";
...
     const defaultWebAppId: string | undefined = getWorkspaceSetting(configurationSettings.defaultWebAppToDeploy, context.workspace.uri.fsPath);
-
-    if (defaultWebAppId && defaultWebAppId !== none) {
+    if (defaultWebAppId === none) {
+        context.telemetry.properties.deployedWithConfigs = 'none';
+    } else if (defaultWebAppId) {
         const defaultWebApp: WebAppTreeItem | undefined = await ext.tree.findTreeItem(defaultWebAppId, context); // resolves to undefined if app can't be found
         if (defaultWebApp) {
             context.deployedWithConfigs = true;
```
EricJizbaMSFT: The name `deployedWithConfigs` doesn't make sense to me. Which configs is it talking about? I would guess defaultWebAppToDeploy _and_ deploySubpath based on the name, but as far as I can tell only `defaultWebAppToDeploy` matters for this. As a possible alternative, we could refactor to have a `webAppSource` property that allows the values `setting`, `rightClick` or `nodePicker`

src/commands/deploy/deploy.ts
```diff
@@ -158,7 +158,8 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         deployContext.telemetry.properties.cancelStep = '';
```
EricJizbaMSFT: I think another root cause is disabling `max-func-body-length cyclomatic-complexity` in `deploy` because it makes it hard to follow all of the possible code-paths. Would like to see that fixed sometime soon, but not in this PR

src/commands/deploy/deploy.ts
nturinski: I could write migration code that checks for the `none` value in showDefaultWebApp prompt and then sets `appService.showDefaultWebAppPrompt`.  That seems simple enough, but is an ugly solution that would have to remain in the codebase _FOREVER_ 😱

src/commands/getDefaultWebAppToDeploy.ts
nturinski: Yeah, agreed. I think back when I made that property, we never talked about using `deploySubpath` settings.  I think I'd prefer a `webAppSource` property as you describe as the subDeployPath stuff should be handled by the `appservice` package now.

src/commands/deploy/deploy.ts
nturinski: I'll assign this to @wwlorey 😁 

jk, but yes, I agree.  I'll start refactoring this after the release.  This code is getting to be a mess: https://github.com/microsoft/vscode-azureappservice/issues/1112

src/commands/deploy/deploy.ts
```diff
@@ -21,23 +21,26 @@ import { getRandomHexString } from "../../utils/randomUtils";
...
 import { getWorkspaceSetting, updateWorkspaceSetting } from '../../vsCodeConfig/settings';
-import { getDefaultWebAppToDeploy } from '../getDefaultWebAppToDeploy';
 import { startStreamingLogs } from '../startStreamingLogs';
-import { IDeployWizardContext } from './IDeployWizardContext';
+import { getDefaultWebAppToDeploy } from './getDefaultWebAppToDeploy';
+import { IDeployWizardContext, WebAppSource } from './IDeployWizardContext';
 
 // tslint:disable-next-line:max-func-body-length cyclomatic-complexity
 export async function deploy(context: IActionContext, confirmDeployment: boolean, target?: vscode.Uri | SiteTreeItem | undefined): Promise<void> {
+    context.telemetry.properties.deployedWithDefaultApp = 'false';
```
EricJizbaMSFT: Can we use the same `webAppSource` property for telemetry? It would be more consistent, easier to write a query for in telemetry, and I think less code since you'd only have to set it once after `webAppSource` has been set

src/commands/deploy/deploy.ts
```diff
@@ -21,23 +21,26 @@ import { getRandomHexString } from "../../utils/randomUtils";
...
-    context.telemetry.properties.deployedWithConfigs = 'false';
+    let node: SiteTreeItem | undefined;
     let deployFsPath: string | undefined;
+    let webAppSource: WebAppSource;
 
     if (target instanceof vscode.Uri) {
-        deployFsPath = target.fsPath;
         context.telemetry.properties.deploymentEntryPoint = 'fileExplorerContextMenu';
+        deployFsPath = target.fsPath;
+        webAppSource = WebAppSource.nodePicker;
```
EricJizbaMSFT: We don't actually know if this will be `nodePicker` until later on. I would prefer to delay setting this value until we actually know what it will be. So instead of this:
```
node = await getDefaultWebAppToDeploy(deployContext) || <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue, context);
```

It might look like this:
```
node = await getDefaultWebAppToDeploy(deployContext);
if (node) {
    webAppSource = WebAppSource.tree;
} else {
    webAppSource = WebAppSource.nodePicker;
    node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue, context);
}
```

src/commands/deploy/deploy.ts
EricJizbaMSFT: As an example, I think you could make `getDefaultWebAppToDeploy` broader to just be `getWebAppToDeploy` similar to how I had `getDeployFsPath`. In other words it takes in whatever the target is and returns a web app after checking all the possible settings, node picker, target, etc.

src/commands/deploy/deploy.ts
EricJizbaMSFT: It wouldn't have to be there _forever_. We could deprecate the old setting and remove it eventually

src/commands/deploy/deploy.ts
nturinski: webAppSource gets overwritten in getDefaultWebApp to `WebAppSource.setting` if we used the setting, though this might not be good coding practice 🤔

src/commands/deploy/deploy.ts
nturinski: Well, webAppSource will always be nodePicker | setting | treeView, but deployedWithDefaultApp is being used to track whether or not people are using the defaultWebAppSetting including if they are using the `none` setting.

nturinski: That being said, we track if people click the "Don't show again" button, so I don't think we necessarily have to track the 'none' setting.

Issue #840, opened 2019-01-28
WEBSITE_NODE_DEFAULT_VERSION is set to 6.9.1 on windows web app


EricJizbaMSFT:
Repro steps:
1. Turn on advanced creation
1. Create new windows web app

By default there seems to be an app setting `WEBSITE_NODE_DEFAULT_VERSION` set to `6.9.1`. This seems out of date. Also if I create a windows web app from the portal, that isn't set at all.
![screen shot 2019-01-28 at 11 03 38 am](https://user-images.githubusercontent.com/11282622/51859669-888bcb80-22ec-11e9-97be-08280d127ec4.png)

In either case, I get this when deploying:
```
11:06:39 AM emjnotportal: Node.js versions available on the platform are: 0.6.20, 0.8.2, 0.8.19, 0.8.26, 0.8.27, 0.8.28, 0.10.5, 0.10.18, 0.10.21, 0.10.24, 0.10.26, 0.10.28, 0.10.29, 0.10.31, 0.10.32, 0.10.40, 0.12.0, 0.12.2, 0.12.3, 0.12.6, 4.0.0, 4.1.0, 4.1.2, 4.2.1, 4.2.2, 4.2.3, 4.2.4, 4.3.0, 4.3.2, 4.4.0, 4.4.1, 4.4.6, 4.4.7, 4.5.0, 4.6.0, 4.6.1, 4.8.4, 5.0.0, 5.1.1, 5.3.0, 5.4.0, 5.5.0, 5.6.0, 5.7.0, 5.7.1, 5.8.0, 5.9.1, 6.0.0, 6.1.0, 6.2.2, 6.3.0, 6.5.0, 6.6.0, 6.7.0, 6.9.0, 6.9.1, 6.9.2, 6.9.4, 6.9.5, 6.10.0, 6.10.3, 6.11.1, 6.11.2, 6.11.5, 6.12.2, 6.12.3, 7.0.0, 7.1.0, 7.2.0, 7.3.0, 7.4.0, 7.5.0, 7.6.0, 7.7.0, 7.7.4, 7.10.0, 7.10.1, 8.0.0, 8.1.4, 8.4.0, 8.5.0, 8.7.0, 8.8.0, 8.8.1, 8.9.0, 8.9.3, 8.9.4, 8.10.0, 8.11.1, 10.0.0, 10.6.0, 10.14.1.
[...]

PR #1106, opened 2019-08-20
Remove website_node_default_version setting

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/840

This was my least preferred solution, but looking at the portal's request to Azure to create a web app, they weren't doing anything special to prevent that setting from appearing.  

I don't think that the setting is used anymore, so it's probably just remnants from their old way of handling node versioning.

### DIFF
--- a/src/explorer/SubscriptionTreeItem.ts
+++ b/src/explorer/SubscriptionTreeItem.ts
@@ -109,6 +109,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         // site is set as a result of SiteCreateStep.execute()
         const siteClient: SiteClient = new SiteClient(nonNullProp(wizardContext, 'site'), this.root);
 
+        // https://github.com/microsoft/vscode-azureappservice/issues/840
         if (wizardContext.newSiteOS === WebsiteOS.windows) {
             const appSettings: StringDictionary = await siteClient.listApplicationSettings();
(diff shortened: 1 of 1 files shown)


src/explorer/SubscriptionTreeItem.ts
```diff
@@ -109,6 +109,15 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         // site is set as a result of SiteCreateStep.execute()
         const siteClient: SiteClient = new SiteClient(nonNullProp(wizardContext, 'site'), this.root);
 
+        // https://github.com/microsoft/vscode-azureappservice/issues/840
+        if (wizardContext.newSiteOS === WebsiteOS.windows) {
+            const appSettings: StringDictionary = await siteClient.listApplicationSettings();
+            if (appSettings.properties && appSettings.properties.WEBSITE_NODE_DEFAULT_VERSION) {
+                delete appSettings.properties.WEBSITE_NODE_DEFAULT_VERSION;
```
EricJizbaMSFT: Are we sure it should be removed? This article seems to imply it has an effect... maybe we just pick a better default than 6.9.1?

https://blogs.msdn.microsoft.com/azureossds/2018/09/25/how-to-set-node-version-for-windows-node-js-webapp-from-azure-portal/

nturinski:
Closing this.  We found the workaround is to set an empty app settings object when creating the web app, so will do that in favor of this.

PR #1109, opened 2019-08-21
Add deployFsPath and leverage it to get subDeployPath

nturinski:
Had to make some changes from the Functions repo, specifically to accept the `fileExtension` parameter.

This is used exclusively for the Java deploy scenario.  I got rid of the file selection for showWorkspaceFolders since it's supposed to be exclusively for folders anyway.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -32,16 +32,15 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
     let node: SiteTreeItem | undefined;
     const newNodes: SiteTreeItem[] = [];
     context.telemetry.properties.deployedWithConfigs = 'false';
+    let siteConfig: WebSiteModels.SiteConfigResource | undefined;
 
     if (target instanceof SiteTreeItem) {
         node = target;
+        // we can only get the siteConfig earlier if the entry point was a treeItem
+        siteConfig = await node.root.client.getSiteConfig();
     }
 
-    let siteConfig: WebSiteModels.SiteConfigResource | undefined;
     let javaFileExtension: string | undefined;
-
-    // we can only get the siteConfig if the entry point was a treeItem
-    siteConfig = node ? await node.root.client.getSiteConfig() : undefined;
     if (siteConfig && javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
         javaFileExtension = javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion);
     }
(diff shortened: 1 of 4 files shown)


src/commands/deploy/getDeployFsPath.ts
```diff
@@ -0,0 +1,77 @@
...
+import * as vscode from 'vscode';
+import { DialogResponses, IActionContext } from 'vscode-azureextensionui';
+import { configurationSettings, extensionPrefix } from '../../constants';
+import { SiteTreeItem } from '../../explorer/SiteTreeItem';
+import { ext } from '../../extensionVariables';
+import { isPathEqual, isSubpath } from '../../utils/pathUtils';
+import * as workspaceUtil from '../../utils/workspace';
+import { getWorkspaceSetting, updateGlobalSetting } from '../../vsCodeConfig/settings';
+
+export async function getDeployFsPath(context: IActionContext, target: vscode.Uri | SiteTreeItem | undefined, fileExtension?: string): Promise<string> {
```
EricJizbaMSFT: Please move this file into the shared package per discussion in [this comment](https://github.com/microsoft/vscode-azureappservice/pull/1060#discussion_r302171653)

package-lock.json
```diff
@@ -7834,10 +7844,29 @@
...
                 "request-promise": "^4.2.2",
                 "simple-git": "~1.92.0",
-                "vscode-azureextensionui": "^0.26.2",
+                "vscode-azureextensionui": "^0.27.0",
                 "vscode-azurekudu": "^0.1.9",
                 "vscode-nls": "^4.0.0",
                 "websocket": "^1.0.25"
+            },
+            "dependencies": {
+                "vscode-azureextensionui": {
```
EricJizbaMSFT: You should update the ui package (and address the breaking changes) at the same time so you don't have multiple different ui packages

src/commands/deploy/deploy.ts
```diff
@@ -31,30 +31,20 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
...
-        // we can only get the siteConfig if the entry point was a treeItem
-        siteConfig = node ? await node.root.client.getSiteConfig() : undefined;
-
-        if (siteConfig && javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
-            const fileExtension: string = javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion);
-            deployFsPath = await workspaceUtil.showWorkspaceFolders(`Select the ${fileExtension} file to deploy...`, context, constants.configurationSettings.deploySubpath, fileExtension);
-        } else {
-            deployFsPath = await workspaceUtil.showWorkspaceFolders("Select the folder to deploy", context, constants.configurationSettings.deploySubpath);
-        }
+    let javaFileExtension: string | undefined;
```
EricJizbaMSFT: nit: rename to just `fileExtension`. Otherwise it seems like all calls to `getDeployFsPath` are for java

PR #1118, opened 2019-08-27
Add .NET Defaults on deployment

### LINKED ISSUE #945: Improve smart defaults for .NET Core apps

nturinski:
Fixes #945 

For now, we are only looking for `csproj` files in the root and its subfolders.  If we find multiple `csproj` files in one project, we assume that the `csproj` is in the user's root path. If a `csproj` file was not found, but the app has a .NET runtime, we also assume the `csproj` is in the root path (it probably won't be, but this will write the tasks and settings for the user so that they can discover/edit as needed)

We don't support multiple frameworks as we only use the `publish` command without a target framework.  Users will see errors when running `publish` with a `csproj` that contains multiple frameworks, so they will be able to edit it accordingly.  

I'd assume a user who is in an advanced scenario and is configuring a project for multiple frameworks is probably already used to this process.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -24,6 +24,7 @@ import { getWorkspaceSetting, updateWorkspaceSetting } from '../../vsCodeConfig/
 import { getDefaultWebAppToDeploy } from '../getDefaultWebAppToDeploy';
 import { startStreamingLogs } from '../startStreamingLogs';
 import { IDeployWizardContext } from './IDeployWizardContext';
+import { setPreDeployTaskForDotnet } from './setPreDeployTaskForDotnet';
 
 // tslint:disable-next-line:max-func-body-length cyclomatic-complexity
 export async function deploy(context: IActionContext, confirmDeployment: boolean, target?: vscode.Uri | SiteTreeItem | undefined): Promise<void> {
@@ -163,6 +164,7 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         node.promptToSaveDeployDefaults(deployContext.workspace.uri.fsPath, deployContext.deployFsPath, deployContext);
     }
 
+    await setPreDeployTaskForDotnet(deployContext, siteConfig);
     await appservice.runPreDeployTask(deployContext, deployContext.deployFsPath, siteConfig.scmType, constants.extensionPrefix);
 
     cancelWebsiteValidation(node);
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -154,7 +154,7 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         node.promptToSaveDeployDefaults(deployContext.workspace.uri.fsPath, deployContext.deployFsPath, deployContext);
     }
 
-    await setPreDeployTaskForDotnet(deployContext, siteConfig);
+    await setPreDeployTaskForDotnet(deployContext);
     await appservice.runPreDeployTask(deployContext, deployContext.deployFsPath, siteConfig.scmType, constants.extensionPrefix);
 
     cancelWebsiteValidation(node);
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -125,6 +125,8 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         }
     }
 
+    await setPreDeployTaskForDotnet(deployContext);
+
     if (confirmDeployment && siteConfig.scmType !== constants.ScmType.LocalGit && siteConfig !== constants.ScmType.GitHub) {
         const warning: string = `Are you sure you want to deploy to "${node.root.client.fullName}"? This will overwrite any previous deployment and cannot be undone.`;
         context.telemetry.properties.cancelStep = 'confirmDestructiveDeployment';
@@ -155,8 +157,6 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
 
     // tslint:disable-next-line:no-floating-promises
     node.promptToSaveDeployDefaults(deployContext, deployContext.workspace.uri.fsPath, deployContext.deployFsPath);
-
-    await setPreDeployTaskForDotnet(deployContext);
     await appservice.runPreDeployTask(deployContext, deployContext.deployFsPath, siteConfig.scmType, constants.extensionPrefix);
 
     cancelWebsiteValidation(node);
--- a/src/commands/deploy/setPreDeployTaskForDotnet.ts
+++ b/src/commands/deploy/setPreDeployTaskForDotnet.ts
@@ -3,16 +3,17 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { SiteConfig } from 'azure-arm-website/lib/models';
 import * as fse from 'fs-extra';
 import * as path from 'path';
+import { MessageItem, TaskDefinition } from 'vscode';
+import { DialogResponses } from 'vscode-azureappservice/node_modules/vscode-azureextensionui';
 import * as constants from '../../constants';
 import { ext } from '../../extensionVariables';
 import { getWorkspaceSetting, updateWorkspaceSetting } from '../../vsCodeConfig/settings';
 import * as tasks from '../../vsCodeConfig/tasks';
 import { IDeployWizardContext } from "./IDeployWizardContext";
 
-export async function setPreDeployTaskForDotnet(context: IDeployWizardContext, siteConfig: SiteConfig): Promise<void> {
+export async function setPreDeployTaskForDotnet(context: IDeployWizardContext): Promise<void> {
     const preDeployTaskSetting: string = 'preDeployTask';
     const configurePreDeployTasksSetting: string = 'configurePreDeployTasks';
     const workspaceFspath: string = context.workspace.uri.fsPath;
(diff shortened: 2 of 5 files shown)


nturinski:
The prompt looks like this

image

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,112 @@
...
+function getTasks(deploySubpath: string, subfolder: string): TaskDefinition[] {
+    // tslint:disable-next-line: no-unsafe-any no-invalid-template-strings
+    const cwd: string = path.posix.join('${workspaceFolder}', subfolder);
+    return [
+        {
+            label: 'clean',
+            command: 'dotnet clean',
+            type: 'shell',
+            problemMatcher: '$msCompile',
+            options: {
```
EricJizbaMSFT: You should keep the tasks clean if there is no subfolder and omit `options` entirely

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,112 @@
...
+                cwd
+            }
+        },
+        {
+            label: 'publish',
```
EricJizbaMSFT: I recommend you follow these steps:
1. Create a new folder
1. Run `dotnet new webapp`
1. Open that folder in VS Code
1. Wait a bit for this notification
Screen Shot 2019-08-29 at 2 05 09 PM
1. Click yes

You'll notice that it adds a `publish` task already. Shouldn't we be using what they created? Seems odd to have two different publish tasks

Also their dialog isn't perfect, but I like it better than ours. It feels more like plain english instead of technical terms/settings.

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,112 @@
...
+
+            await updateWorkspaceSetting(preDeployTaskSetting, 'publish', context.workspace.uri.fsPath);
+            await updateWorkspaceSetting(constants.configurationSettings.deploySubpath, deploySubpath, workspaceFspath);
+
+            // update the deployContext.deployFsPath with the .NET output path since getDeployFsPath is called prior to this
+            // purposely not using posix since this is happening at runtime
+            context.deployFsPath = path.join(context.workspace.uri.fsPath, deploySubpath);
+
+            const defaultTasks = getTasks(deploySubpath, subfolder);
+            tasks.updateTasks(context.workspace, defaultTasks);
```
EricJizbaMSFT: This will overwrite any and all existing tasks, which is not what we want

src/commands/deploy/deploy.ts
```diff
@@ -155,6 +156,7 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
     // tslint:disable-next-line:no-floating-promises
     node.promptToSaveDeployDefaults(deployContext, deployContext.workspace.uri.fsPath, deployContext.deployFsPath);
 
+    await setPreDeployTaskForDotnet(deployContext);
```
EricJizbaMSFT: I feel like I would expect this prompt _before_ the "Are you sure you want to deploy?" prompt. Not 100% sure on this though

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,112 @@
...
+            });
+
+            return filePaths.filter((f: string) => /\.csproj$/i.test(f));
+        }
+    }
+}
+
+function getTasks(deploySubpath: string, subfolder: string): TaskDefinition[] {
+    // tslint:disable-next-line: no-unsafe-any no-invalid-template-strings
+    const cwd: string = path.posix.join('${workspaceFolder}', subfolder);
```
EricJizbaMSFT: Add a note why you're using posix

src/commands/deploy/setPreDeployTaskForDotnet.ts
EricJizbaMSFT: ping @fiveisprime can you also go through this experience and help Nathan decide how his work should integrate with the basic C# experience

src/commands/deploy/deploy.ts
nturinski: Yeah, I agree with that now that you bring it up.

src/commands/deploy/setPreDeployTaskForDotnet.ts
nturinski: What about in the case of a csproj in a subfolder?

The publish task that they create seems to have some extra args that make the logs less verbose?  It also doesn't dependOn the clean task.

I think I would rather write over it.

EricJizbaMSFT: Lol I trust the C# extension for this kind of thing more than I trust us. We can also file issues on them if we truly think something is better from ours

For example - is the clean task actually necessary? Maybe publish cleans by default

fiveisprime: We should rely on the C# extension as much as possible. 

1. They know best how to manage the tasks and what's necessary
1. I believe we're starting to creep out of scope of the app service extension

It seems safe to assume that C# customers will have the extension installed so I'm not so worried about that. I'll schedule a meeting with the C# extension owners next week to work through what we should do and what they might be able to add to help us.

nturinski: Based on the meeting we had with Ravi, I would say our takeaway is that we shouldn't rely on the C# extension for the publish task.  Apparently, they haven't changed their task generator in a while and have no plans to change it.  But I will change our tasks to be formatted in the same way as theirs.

An issue that came up from the meeting is if we don't replace the user's publish task, we can't guarantee that we know where the output bits will be.  We might be able to read `dotnet publish`'s output and find the folder that way?

EricJizbaMSFT: > Based on the meeting we had with Ravi, I would say our takeaway is that we shouldn't rely on the C# extension for the publish task. 

I feel like we're still relying on the C# extension and it's publish task, but perhaps we have different definitions of "rely" so I just wanted to clarify what I expect:
1. If the publish task exists (which it will in most cases IMO) we should use that
1. If the publish task doesn't exist, we would create a task for them, but it will basically be a duplicate of what the C# extension would've created

> An issue that came up from the meeting is if we don't replace the user's publish task, we can't guarantee that we know where the output bits will be.

I feel like this feature is more about the 90% case. Users can always change these settings themselves, so I'm not too worried about guaranteeing the default settings will work 100% of the time.

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,160 @@
...
+            "build",
+            cwd,
+            "/property:GenerateFullPaths=true",
+            "/consoleloggerparameters:NoSummary"
+        ],
+        problemMatcher: "$msCompile"
+    };
+
+    const publishTask: TaskDefinition = {
+        label: "publish",
```
EricJizbaMSFT: Pending discussion on a few things:
1. Should this depend on a clean task
1. Should this be "Release" instead of "Debug"
1. Should we use the "publish" label or create our own name to differentiate from C# extension

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,160 @@
...
+        args: [
+            "publish",
+            cwd,
+            "/property:GenerateFullPaths=true",
+            "/consoleloggerparameters:NoSummary"
+        ],
+        problemMatcher: "$msCompile"
+    };
+
+    const watchTask: TaskDefinition = {
```
EricJizbaMSFT: Do we want to add watch if it's unrelated to deploy?

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,160 @@
...
+        const targetFramework: string | undefined = await tryGetTargetFramework(csprojFile);
+        context.telemetry.properties.tfw = targetFramework ? targetFramework : 'N/A';
+        if (!targetFramework) {
+            // if the target framework cannot be found, don't try to set defaults
+            return;
+        }
+
+        const notConfiguredForDeploy: string = `Required assets to build and deploy are missing from "${context.workspace.name}"? Add them?`;
+        const dontShowAgainButton: MessageItem = { title: "No, and don't show again" };
+        const input: MessageItem = await ext.ui.showWarningMessage(notConfiguredForDeploy, { modal: true }, DialogResponses.yes, dontShowAgainButton);
```
EricJizbaMSFT: I can tell you modeled this off the C# extension, but you know I'm not a fan of "Yes" buttons. As an alternative, you could get rid of `Add them?` and change the button to `Add Assets`

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,160 @@
...
+
+async function tryGetCsprojFile(context: IDeployWizardContext, projectPath: string): Promise<string | undefined> {
+    let projectFiles: string[] = await checkFolderForCsproj(projectPath);
+    // it's a common pattern to have the .csproj file in a subfolder so check one level deeper
+    if (projectFiles.length === 0) {
+        const subfolders: string[] = await fse.readdir(projectPath);
+        await Promise.all(subfolders.map(async folder => {
+            const filePath: string = path.join(projectPath, folder);
+            // check its existence as this will check .vscode even if the project doesn't contain that folder
+            if (fse.existsSync(filePath) && (await fse.stat(filePath)).isDirectory()) {
```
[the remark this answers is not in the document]
nturinski: Just to understand, do we want to use the async methods due to UI reasons?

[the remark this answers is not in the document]
EricJizbaMSFT: Here's a random blog post I found that explains why you should always use `async` if you can:

> When JavaScript is executed, synchronous code has the potential to block further execution until it has finished what it’s doing. In English, long-running JavaScript functions can make the UI or server unresponsive until the function has returned. Obviously this can result in a terrible user-experience.
> 
> For example: if you want to load your latest tweets onto a web page, and you do this synchronously, then a visitor to your site won’t be able to do anything until those tweets are loaded. This could cause a long delay before they even get to see the content of your site!

https://rowanmanning.com/posts/javascript-for-beginners-async/

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,160 @@
...
+            await updateWorkspaceSetting(constants.configurationSettings.deploySubpath, deploySubpath, workspaceFspath);
+
+            // update the deployContext.deployFsPath with the .NET output path since getDeployFsPath is called prior to this
+            context.deployFsPath = path.join(workspaceFspath, deploySubpath);
+
+            const existingTasks: tasks.ITask[] = tasks.getTasks(context.workspace);
+
+            // do not overwrite any dotnet commands the user may have
+            let dotnetTasks: tasks.ITask[] = generateDotnetTasks(subfolder);
+            dotnetTasks = dotnetTasks.filter(t1 => !existingTasks.find(t2 => {
```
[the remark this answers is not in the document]
nturinski: The tasks include the `cwd` as its args which will run the publish command in that source folder.

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,160 @@
...
+        if (!targetFramework) {
+            // if the target framework cannot be found, don't try to set defaults
+            return;
+        }
+
+        const notConfiguredForDeploy: string = `Required assets to build and deploy are missing from "${context.workspace.name}"? Add them?`;
+        const dontShowAgainButton: MessageItem = { title: "No, and don't show again" };
+        const input: MessageItem = await ext.ui.showWarningMessage(notConfiguredForDeploy, { modal: true }, DialogResponses.yes, dontShowAgainButton);
+        if (input === dontShowAgainButton) {
+            await updateWorkspaceSetting(showPreDeployWarningSetting, false, workspaceFspath);
```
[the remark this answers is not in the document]
nturinski: I thought it'd be better as a workspace setting since this seems like it could be project specific.  I might have a project I know that I never want configured, but maybe other apps I create via `dotnet` it'd be nice for VS Code to handle it.

src/commands/deploy/setPreDeployTaskForDotnet.ts
nturinski: > [4:38 PM] Ahmed Metwally
>     1- We should call clean at first.
> 2- We can default to debug to be like everyone else.
> 3- We can reuse the C# extension publish task or create one if it doesn't exist.

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,160 @@
...
+            tasks.updateTasks(context.workspace, newTasks);
+        }
+
+    }
+}
+
+async function tryGetCsprojFile(context: IDeployWizardContext, projectPath: string): Promise<string | undefined> {
+    let projectFiles: string[] = await checkFolderForCsproj(projectPath);
+    // it's a common pattern to have the .csproj file in a subfolder so check one level deeper
+    if (projectFiles.length === 0) {
```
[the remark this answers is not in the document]
nturinski: Wouldn't changing it to `if (projectFiles.length !== 1) {` mean it would check for subfolders if I found 0 or if I found more than 1?

Right now the behavior is check only if there were 0 .csproj files in the root, but if there were  1 or more, skip to to the return statement.

src/commands/deploy/setPreDeployTaskForDotnet.ts
nturinski: Nah, after we talked about it, I think it only makes sense to add "publish" and "clean"

src/commands/deploy/setPreDeployTaskForDotnet.ts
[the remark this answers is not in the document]
nturinski: image

image

src/commands/deploy/setPreDeployTaskForDotnet.ts
nturinski: image

src/commands/deploy/setPreDeployTaskForDotnet.ts
```diff
@@ -0,0 +1,155 @@
...
+
+            await updateWorkspaceSetting(preDeployTaskSetting, 'publish', workspaceFspath);
+            await updateWorkspaceSetting(constants.configurationSettings.deploySubpath, deploySubpath, workspaceFspath);
+
+            // update the deployContext.deployFsPath with the .NET output path since getDeployFsPath is called prior to this
+            context.deployFsPath = path.join(workspaceFspath, deploySubpath);
+
+            const existingTasks: tasks.ITask[] = tasks.getTasks(context.workspace);
+            const publishTask: tasks.ITask | undefined = existingTasks.find(t1 => {
+                return t1.label === 'publish';
```
EricJizbaMSFT: make a const for the publish task id to be used here and when creating the task below. Same for clean

PR #1122, opened 2019-08-29
Nat/detector

### LINKED ISSUE #911: Use detectors to offer insights for failed deployments on Linux

nturinski:
From #1079 

> Fixes #911
> 
> So the idea is after a "completed deployment" if validateWebSite returns a 400 or 500 code, then we should check the detectors to see if there is any troubleshooting available to the user.
> 
> The types all describe how the data comes back, but since it's primarily supposed to be used as a table, the way I had to get the data is a little unconventional.
> 
> Moved to opn because vscode.openUrl does not work for the deeplink that will take them to the portal.
> 
> I can't currently test due to portal issues, but will add screenshots when I'm able.


src/commands/postDeploy/runPostDeployTask.ts
```diff
@@ -0,0 +1,19 @@
...
+import { checkLinuxWebAppDownDetector } from "./checkLinuxWebAppDownDetector";
+import { validateWebSite } from "./validateWebSite";
+
+export const postDeployCancelTokens: Map<string, CancellationTokenSource> = new Map();
+export async function runPostDeployTask(node: SiteTreeItem, correlationId: string, tokenSource: CancellationTokenSource): Promise<void> {
+    await validateWebSite(correlationId, node, tokenSource);
+
+    // this currently only works for Linux apps, so ignore if it's Windows
+    if (!node.root.client.isLinux) {
+        await checkLinuxWebAppDownDetector(correlationId, node, tokenSource);
```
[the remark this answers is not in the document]
nturinski: Oh. Yes, you are completely right, I didn't mean for it to behave like this.

src/commands/postDeploy/validateWebSite.ts
```diff
@@ -60,30 +39,28 @@ export async function validateWebSite(deploymentCorrelationId: string, siteTreeI
...
-            resolveWithFullResponse: true
-        };
         let currentStatusCode: number | undefined = 0;
         const statusCodes: { code: number | undefined, elapsed: number }[] = [];
 
         // tslint:disable-next-line:no-constant-condition
         while (true) {
+            if (tokenSource.token.isCancellationRequested) {
+                // the user cancelled the check by deploying again
+                context.telemetry.properties.canceled = 'true';
```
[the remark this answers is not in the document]
nturinski: Yeah, but Stephen had this being tracked here https://github.com/microsoft/vscode-azureappservice/pull/1122/files#diff-be3a5553e0be71663344e72f967dde08L82-L86 so I didn't want to change the properties.  Thought it might mess with his data visualizations.

test/parseDetectorTables.test.ts
```diff
@@ -0,0 +1,57 @@
...
...nce", dataType: "String", columnType: null }, { columnName: "Facility", dataType: "String", columnType: null }, { columnName: "FailureCount", dataType: "Int64", columnType: null }], rows: [["2019-07-26T20:40:00", "467983_Small_90", "naturins-linux", 1]] }, renderingProperties: { defaultValue: 0, graphType: 1, graphOptions: null, timestampColumnName: null, roleInstanceColumnName: null, seriesColumns: null, type: 2, title: "Number of app Start Failures", description: null } }, { table: { tableName: "", columns: [{ columnName: "Time", dataType: "DateTime", columnType: null }, { columnName: "Instance", dataType: "String", columnType: null }, { columnName: "Facility", dataType: "String", columnType: null }, { columnName: "Message", dataType: "String", columnType: null }, { columnName: "FailureCount", dataType: "Int64", columnType: null }], rows: [["2019-07-26T20:40:00", "467983_Small_90", "naturins-linux", "docker container could not be started: naturins-linux_1", 1]] }, renderingProperties: { displayColumnNames: null, groupByColumnName: null, type: 1, title: null, description: null } }], status: { message: null, statusId: 0 }, dataProvidersMetadata: null, suggestedUtterances: null } };
```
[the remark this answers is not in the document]
nturinski: no-use-before-declare :(

test/parseDetectorTables.test.ts
```diff
@@ -0,0 +1,57 @@
...
+const vsCodeIntegration: string = 'Latest time seen by detector. To be used in VSCode integration.';
+suite('Detector Dataset Parser', () => {
+    test('Find table by column name', async () => {
+        const table: detectorTable | undefined = findTableByColumnName(detectorJSONWithCriticalErrors.properties.dataset, ColumnName.failureCount);
```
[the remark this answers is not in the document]
nturinski: Sorry, I'm not sure what you mean by this comment

test/parseDetectorTables.test.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Turn that rule off - it's actually discouraged

https://palantir.github.io/tslint/rules/no-use-before-declare/

src/commands/deploy/deploy.ts
```diff
@@ -74,6 +75,12 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
         }
     }
 
+    // cancel the previous detector check from the same web app
+    const previousTokenSource: vscode.CancellationTokenSource | undefined = postDeployCancelTokens.get(node.id);
+    if (previousTokenSource) {
+        previousTokenSource.cancel();
```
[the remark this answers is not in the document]
nturinski: Yeah, good point

test/parseDetectorTables.test.ts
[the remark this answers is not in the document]
EricJizbaMSFT: All of this stuff can be removed:
```
	    id: "/subscriptions/9b5c7ccb-9857-4307-843b-8875e83f65e9/resourceGroups/appsvc_linux_centralus/providers/Microsoft.Web/sites/naturins-linux/detectors/LinuxContainerStartFailure",
    name: "LinuxContainerStartFailure",
    type: "Microsoft.Web/sites/detectors",
    location: "West US",
    properties: {
        metadata: {
            id: "LinuxContainerStartFailure",
            name: "Container Issues",
            category: "Availability and Performance",
            author: "",
            description: "Investigate potential reasons why your container failed to start, which could cause deployment and availability issues.",
            type: "Detector",
            supportTopicList: [

            ],
            analysisTypes: null,
            score: 0.0,
            typeId: "Diagnostics.ModelsAndUtils.Attributes.Definition, Diagnostics.ModelsAndUtils, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null"
        },
```

Just define `dataset` instead of the whole `detectorJSONWithCriticalErrors`

src/commands/postDeploy/validateWebSite.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah but you changed it to be `throw new UserCancelledError();`, which affects the telemetry, which means you should update the telemetry and any queries/visualizations we have

Or else change it back to `break;` like it used to be

src/commands/postDeploy/runPostDeployTask.ts
```diff
@@ -0,0 +1,33 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { CancellationTokenSource } from "vscode";
+import { SiteTreeItem } from "../../explorer/SiteTreeItem";
+import { checkLinuxWebAppDownDetector } from "./checkLinuxWebAppDownDetector";
+import { validateWebSite } from "./validateWebSite";
+
+export const postDeployCancelTokens: Map<string, CancellationTokenSource> = new Map();
```
EricJizbaMSFT: Seems like an odd place to store `postDeployCancelTokens` when you don't even use it in this file

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,176 @@
...
+        const deployment: DeployResult = await kuduClient.deployment.getResult('latest');
+        if (!deployment.startTime) {
+            // if there's no deployment detected, nothing can be done
+            context.telemetry.properties.cancelStep = 'noDeployResult';
+            return;
+        }
+
+        const deployResultTime: Date = new Date(deployment.startTime);
+        const enableDetectorsSetting: string = 'enableDetectors';
+        const showOutput: boolean | undefined = getGlobalSetting<boolean>(enableDetectorsSetting);
```
EricJizbaMSFT: You should almost always use `getWorkspaceSetting`. There's no reason to ignore the value if it's set at the workspace level and `getWorkspaceSetting` will defer to the global setting if the workspace setting isn't defined

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,176 @@
...
+        const detectorUri: string = `${node.id}/detectors/${linuxContainerStartFailureId}`;
+        const requestOptions: requestUtils.Request = await requestUtils.getDefaultAzureRequest(detectorUri, node.root);
+
+        // these parameters were specified to retrieve the timestamp from the detector by Detectors team
+        // string "Latest time seen by detector. To be used in VSCode integration."" is added to response
+        // when val: 'vscode' is added
+        requestOptions.qs = {
+            'api-version': "2015-08-01",
+            fId: "1",
+            btnId: "2",
```
EricJizbaMSFT: What do all these ids mean - add some comments in the code to clarify

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,176 @@
...
+                // the user cancelled the check by deploying again
+                context.telemetry.properties.cancelStep = 'cancellationToken';
+                throw new UserCancelledError();
+            }
+
+            if (Date.now() > detectorTimeoutMs) {
+                if (showOutput) {
+                    const noIssuesFound: string = `Diagnosing for "${node.root.client.siteName}" has timed out.`;
+                    ext.outputChannel.appendLine(noIssuesFound);
+                    context.telemetry.properties.failureCount = '0';
```
EricJizbaMSFT: This seems disingenuous - we don't know the failure count if there's a timeout

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,176 @@
...
+            requestOptions.qs.endTime = new Date().toISOString();
+            detectorErrorMessage = await getLinuxDetectorError(context, requestOptions, deployResultTime, node.root.client.fullName);
+
+            // poll every minute
+            await delay(1000 * 60);
+        } while (!detectorErrorMessage);
+
+        if (showOutput) {
+            // this detector has more valuable information to show in the portal than 'LinuxContainerStartFailure'
+            const linuxAppDownId: string = 'LinuxAppDown';
```
EricJizbaMSFT: Why make a `const` for this? Couldn't you just put it in the queryPrefix directly

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,176 @@
...
+            return undefined;
+        }
+
+        const errorMessages: string[] = getValuesByColumnName(criticalErrorTable, ColumnName.value);
+        const failureCount: string = getValuesByColumnName(failureCountTable, ColumnName.failureCount)[0];
+
+        context.telemetry.properties.errorMessages = JSON.stringify(errorMessages);
+        context.telemetry.properties.failureCount = failureCount;
+
+        return `Critical insights found for "${siteName}": ${failureCount} as of ${new Date(timestamp).toLocaleString()}. Critical insight preview: "${errorMessages[0]}". Click here to access "App Service Diagnostics" for recommendations.`;
```
EricJizbaMSFT: Can I see a screenshot of this

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    return undefined;
+}
+
+export function validateTimestamp(detectorTime: Date, deployResultTime: Date): boolean {
+    const timeBetweenDeployAndDetector: number = Math.abs(detectorTime.getSeconds() - deployResultTime.getSeconds());
+
+    // there's usually around a 5-20 second difference between the deploy result and time reported by the detector
+    if (timeBetweenDeployAndDetector <= 20) {
```
EricJizbaMSFT: I still don't trust this. Can you at least add telemetry tracking the time between deploy and detector so that we can better understand how far apart these times actually are

Also add some sort of telemetry so that we know if `validateTimestamp` is causing us not to find any detectors

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,176 @@
...
+        const criticalErrorTable: detectorTable | undefined = findTableByRowValue(datasets, 'Critical');
+        const failureCountTable: detectorTable | undefined = findTableByColumnName(datasets, ColumnName.failureCount);
+
+        // if the criticalError table or failureCounts aren't defined, we can't display anything to the user
+        if (!criticalErrorTable || !failureCountTable) {
+            return undefined;
+        }
+
+        const errorMessages: string[] = getValuesByColumnName(criticalErrorTable, ColumnName.value);
+        const failureCount: string = getValuesByColumnName(failureCountTable, ColumnName.failureCount)[0];
```
EricJizbaMSFT: So how often do we expect multiple rows vs. just the first row? Since we only ever show the first one... Is it possible the first row will have 0 failures and not be useful? Can we somehow find the most useful row?

Should probably add telemetry for how many rows there are.

Also maybe just change `getValuesByColumnName` to always return the first row. I feel like that would be simpler than always doing `[0]`

src/commands/postDeploy/parseDetectorResponse.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    return undefined;
+}
+
+export function getValuesByColumnName(table: detectorTable, columnName: ColumnName): string[] {
+    // -1 indicates that findIndex returned nothing
+    let rowIndex: number = -1;
+    const values: string[] = [];
+
+    for (let i = 0; i < table.columns.length; i++) {
```
EricJizbaMSFT: Something simpler like this should work:
```
const rowIndex: number = table.columns.findIndex(column => column.columnName === columnName);
```

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -0,0 +1,176 @@
...
+            // this detector has more valuable information to show in the portal than 'LinuxContainerStartFailure'
+            const linuxAppDownId: string = 'LinuxAppDown';
+            await ext.ui.showWarningMessage(detectorErrorMessage, { title: 'Open in Portal' });
+            await openInPortal(node.root, `${node.root.client.id}/troubleshoot`, { queryPrefix: `websitesextension_ext=asd.featurePath%3Ddetectors%2F${linuxAppDownId}` });
+            context.telemetry.properties.didClick = 'true';
+        }
+    });
+}
+
+async function getLinuxDetectorError(context: IActionContext, requestOptions: WebResource, deployResultTime: Date, siteName: string): Promise<string | undefined> {
```
EricJizbaMSFT: Suggest moving this function to its own file. This file is only going to get more and more complex if/when the feature ever gets enabled

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
nturinski: image

nturinski:
I couldn't remember how to fix the tests when a function requires an ActionContext, so all of the tests are currently broken, but I will fix that asap.

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
EricJizbaMSFT: ping on this one

PR #1129, opened 2019-09-04
Add AppInsights steps to site creation

### LINKED ISSUE #222: Enable Application Insights on creation
Create an Azure Monitor node in the tree with Application Insights under the node that links out to the portal.

See spec doc.

nturinski:
Fixes #222 

For Linux apps, there are a couple of settings we need 
 
> `APPINSIGHTS_INSTRUMENTATIONKEY` I think you’re already setting this
> `APPLICATIONINSIGHTSAGENT_EXTENSION_ENABLED` set to “true”

I copied the settings from the portal for Windows app when AI is enabled.

### DIFF
--- a/src/explorer/SubscriptionTreeItem.ts
+++ b/src/explorer/SubscriptionTreeItem.ts
@@ -105,6 +105,9 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         if (!context.advancedCreation) {
             await setDefaultRgAndPlanName(wizardContext, siteStep);
             wizardContext.newAppInsightsName = await wizardContext.relatedNameTask;
+            if (!wizardContext.newAppInsightsName) {
+                throw new Error('Failed to generate unique name for resources. Use advanced creation to manually enter resource names.');
+            }
         }
 
         await wizard.execute();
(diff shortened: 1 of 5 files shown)


src/explorer/SubscriptionTreeItem.ts
```diff
@@ -123,7 +123,19 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             }
 
             appSettings.properties.APPINSIGHTS_INSTRUMENTATIONKEY = wizardContext.appInsightsComponent.instrumentationKey;
-            appSettings.properties.APPLICATIONINSIGHTSAGENT_EXTENSION_ENABLED = 'true';
+            if (wizardContext.newSiteOS === WebsiteOS.windows) {
+                // all these settings are set on the portal if AI is enabled for Windows apps
+                appSettings.properties.APPINSIGHTS_PROFILERFEATURE_VERSION = 'disabled';
```
EricJizbaMSFT: nit: would prefer to pull `disabled` into a `const` since it's used so often

nturinski:
Oh, I'll have to fix the tests.  I'll do that in this PR.

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -112,6 +116,30 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         const createdNewAppMsg: string = `Created new web app "${siteClient.fullName}": https://${siteClient.defaultHostName}`;
         ext.outputChannel.appendLine(createdNewAppMsg);
 
+        if (wizardContext.appInsightsComponent && wizardContext.appInsightsComponent.instrumentationKey) {
```
EricJizbaMSFT: Wait turns out I only reviewed one commit not the whole PR 😂

This should be a part of SiteCreateStep

https://github.com/microsoft/vscode-azuretools/blob/master/appservice/src/createAppService/SiteCreateStep.ts#L63

Rather than making a separate call to update app settings after the web app is created

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -100,6 +103,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
 
         if (!context.advancedCreation) {
             await setDefaultRgAndPlanName(wizardContext, siteStep);
+            wizardContext.newAppInsightsName = wizardContext.newSiteName;
```
EricJizbaMSFT: This should use the generated related name or whatever from the wizardContext

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -111,7 +116,62 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+
+    if (context.aiInstrumentationKey) {
+        appSettings.push({
+            name: 'APPINSIGHTS_INSTRUMENTATIONKEY',
+            value: context.aiInstrumentationKey
+        });
+
+        // all these settings are set on the portal if AI is enabled for Windows apps
+        if (context.os === WebsiteOS.windows) {
+            appSettings.push({
```
EricJizbaMSFT: nit: You can simplify this by only calling `push` once:
```
appSettings.push(
    {
        name: 'APPINSIGHTS_PROFILERFEATURE_VERSION',
        value: disabled
    },
    {
        name: 'APPINSIGHTS_SNAPSHOTFEATURE_VERSION',
        value: disabled
    },
    ...
);
```

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -100,6 +104,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
 
         if (!context.advancedCreation) {
             await setDefaultRgAndPlanName(wizardContext, siteStep);
+            wizardContext.newAppInsightsName = await wizardContext.relatedNameTask;
```
EricJizbaMSFT: `relatedNameTask` can technically return undefined. Suggest something like this:
https://github.com/microsoft/vscode-azurefunctions/blob/master/src/tree/SubscriptionTreeItem.ts#L136

PR #1138, opened 2019-09-11
Change for test case structure

v-wuzhai:
Divide the "createAzureResource.test.ts" file into multiple files, and the similar actions are grouped into the same file.

### DIFF
--- a/test/nightly/azureResourceWebAppActions.test.ts
+++ b/test/nightly/azureResourceWebAppActions.test.ts
@@ -14,20 +14,18 @@ import { resourceGroupsToDelete, webSiteClient } from './global.resource.test';
 // tslint:disable-next-line: max-func-body-length
 suite('Web App actions', async function (this: ISuiteCallbackContext): Promise<void> {
     this.timeout(350 * 1000);
-    let regExpLTS: RegExp;
     let resourceName: string;
 
     suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
         if (!longRunningTestsEnabled) {
             this.skip();
         }
-        regExpLTS = /LTS/g;
-        resourceName = getRandomHexString().toLowerCase();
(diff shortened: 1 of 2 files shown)


test/nightly/azureResourceWebAppActions.test.ts
```diff
@@ -4,51 +4,25 @@
...
-suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Promise<void> {
-    this.timeout(1200 * 1000);
-    const resourceGroupsToDelete: string[] = [];
-    const testAccount: TestAzureAccount = new TestAzureAccount(vscode);
-    const regExpLTS: RegExp = /LTS/g;
-    const resourceName: string = getRandomHexString().toLowerCase();
-    let webSiteClient: WebSiteManagementClient;
+suite('Web App actions', async function (this: ISuiteCallbackContext): Promise<void> {
+    this.timeout(350 * 1000);
+    let regExpLTS: RegExp;
```
nturinski: Why was this changed?  I think `regExpLTS` will be constant.

test/nightly/azureResourceWebAppActions.test.ts
```diff
@@ -4,51 +4,25 @@
...
-                await client.resourceGroups.deleteMethod(resourceGroup);
-                console.log(`Resource group "${resourceGroup}" deleted.`);
-            } else {
-                // If the test failed, the resource group might not actually exist
-                console.log(`Ignoring resource group "${resourceGroup}" because it does not exist.`);
-            }
-        }));
-        ext.azureAccountTreeItem.dispose();
+        regExpLTS = /LTS/g;
+        resourceName = getRandomHexString().toLowerCase();
```
nturinski: Lower-case letters are not a limitation of resource groups, so you shouldn't use that here.

PR #1148, opened 2019-09-12
Add support for remote debugging Python web apps

mrcrane:
This PR will fail to build until it is updated with the changes in microsoft/vscode-azuretools#585 See that PR for more context.

This PR introduces the `enablePythonRemoteDebugging` configuration. This is in place because Python remote debugging support is not released to production. When that is released and fully tested, we can remove this config.


nturinski:
@mrcrane Did you want to try to squeeze this into the next App Service release going out next week?  Is Python debugging actually working on the platform yet?

test/remoteDebug/getRemoteDebugLanguage.test.ts
```diff
@@ -0,0 +1,68 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { RemoteDebugLanguage } from 'vscode-azureappservice';
+import { IActionContext } from 'vscode-azureextensionui';
+import { getRemoteDebugLanguage } from '../../extension.bundle';
+
+suite('getRemoteDebugLanguage', () => {
```
nturinski: Could you update the tests to include getRemoteDebugLanguage with the Python flag enabled?  I'm adding this helper function that should aid in that https://github.com/microsoft/vscode-azureappservice/pull/1149

mrcrane:
I added the tests for the configuration setting and bumped the vscode-azureappservice dependency so this is should be good to go now.

Presumably Python debugging is supported on Azure by now. If there are any issues with platform support (or any bugs in general), this should still be good to go because it is hidden behind the experimental flag.

src/commands/remoteDebug/getRemoteDebugLanguage.ts
```diff
@@ -26,13 +30,22 @@ export function checkForRemoteDebugSupport(siteConfig: SiteConfigResource, conte
         if (version.startsWith('node')) {
             const splitVersion = version.split('|');
             if (splitVersion.length > 1 && isNodeVersionSupported(splitVersion[1])) {
-                // Node version is supported, so return successfully
-                return;
+                return RemoteDebugLanguage.Node;
+            } else {
+                throw new Error('Azure Remote Debugging is currently only supported for Node.js version >= 8.11 on Linux.');
```
nturinski: Not blocking, but this error and the error on line 47 should ideally be the same message.

PR #1151, opened 2019-09-13
Detect if workspace is Java project and allow .jar/.war deploy

### LINKED ISSUE #1113: Detect if project is Java before prompting to select folder to deploy
Java users won't be able to select a .jar file to deploy if their entry point is the blue arrow, so detect the workspace first.

wwlorey:
Fixes #1113 

It is still possible to deploy a directory from an auto-detected Java project (from the blue arrow entry point)


wwlorey:
**Offline feedback**: Since `selectWorkspaceFile` isn't consumed externally, the `fileExtensions` argument type can be simplified to just optional `string[]`

src/commands/deploy/deploy.ts
```diff
@@ -101,11 +103,13 @@ export async function deploy(context: IActionContext, confirmDeployment: boolean
     siteConfig = siteConfig ? siteConfig : await node.root.client.getSiteConfig();
 
     if (javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
-        const javaArtifactFiles: vscode.Uri[] = await workspaceUtil.findFilesByFileExtension(deployContext.deployFsPath, javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion));
-        if (javaArtifactFiles.length > 0) {
-            const javaArtifactQp: IAzureQuickPickItem<string>[] = workspaceUtil.mapFilesToQuickPickItems(javaArtifactFiles);
-            // check if there is a jar/war file in the fsPath that was provided
-            deployContext.deployFsPath = <string>(await ext.ui.showQuickPick(javaArtifactQp, { placeHolder: `Select the ${javaUtils.getArtifactTypeByJavaRuntime(siteConfig.linuxFxVersion)} file to deploy...` })).data;
+        if (!fileExtensions) {
```
nturinski: I don't think we need the check or any code inside of it anymore.  This was originally used to prompt the user a second time for a .jar/.war file IF we detected that they were deploying to a Java app.  We do still need the `javaUtils.configureJavaSEAppSettings` though

PR #1176, opened 2019-09-18
Added azure-pipelines as dependency and enabled browse/configure pipeline experience

hiyadav:
### Changes done: 
- _Added azure-pipelines as dependency_
- _Enabled configure/browse pipeline option for webapps_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Change Log
 All notable changes to the "azure-appservice" extension will be documented in this file.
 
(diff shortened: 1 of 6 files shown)


hiyadav:
@ericJizbaMSFT 

We want to enable users to be able to Configure/browse Azure pipelines on a webapp (or GitHub actions in the future). While on a call with Matt we thought of implementing by getting added as a dependency.
We would love to discuss the best way possible to do so. I'll setup a call for tomorrow for the same.

PR #1189, opened 2019-09-25
Use CLI configured defaults when creating an app in basic mode

### LINKED ISSUE #743: Use CLI configured defaults when creating an app in basic mode
There's a way to configure defaults (resource group and plan) within the CLI and I believe that's stored in a JSON file somewhere. The extension should pick those defaults up and use them in basic mode.

wwlorey:
Fixes #743

This applies to the resource group, plan, and location configuration settings.

### DIFF
--- a/src/commands/createWebApp/readAzConfig.ts
+++ b/src/commands/createWebApp/readAzConfig.ts
@@ -1,0 +1,42 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
+import * as os from 'os';
+import * as path from 'path';
+
+export async function readAzConfig(...properties: AzConfigProperties[]): Promise<AzConfig> {
+    const config: AzConfig = {};
+    const configPath: string = path.join(os.homedir(), '.azure', 'config');
+
+    if (await fse.pathExists(configPath)) {
+        const configStr: string = await fse.readFile(configPath, 'utf-8');
+
+        const defaultsIndex = configStr.search('[defaults]');
+        if (defaultsIndex !== -1) {
+            for (const property of properties) {
+                const match: RegExpMatchArray | null = configStr.substring(defaultsIndex).match(RegExp(`${property}[ \\t]?=[ \\t]?(.+)`));
+
+                if (match) {
+                    config[property] = match[1].trim();
+                }
+            }
+        }
+    }
+
+    return config;
+}
+
+export type AzConfig = {
+    group?: string;
+    plan?: string;
+    location?: string;
+};
+
+export enum AzConfigProperties {
+    group = 'group',
+    plan = 'plan',
+    location = 'location'
+}
--- a/src/commands/createWebApp/readAzConfig.ts
+++ b/src/commands/createWebApp/readAzConfig.ts
@@ -27,7 +27,7 @@ export async function readAzConfig(wizardContext: IAppServiceWizardContext & Par
         // Default to local config values if they exist
         readAzConfigListing(wizardContext, propertiesToRead, azLocalConfigListing, config);
     } catch (error) {
-        // Azure CLI not installed so no config to read
+        // Suppress errors - this functionality should not block creating a web app
     }
 
     return config;
--- a/src/commands/createWebApp/readAzConfig.ts
+++ b/src/commands/createWebApp/readAzConfig.ts
@@ -13,8 +13,8 @@ export async function readAzConfig(...properties: AzConfigProperties[]): Promise
 
     if (await fse.pathExists(configPath)) {
         const configStr: string = await fse.readFile(configPath, 'utf-8');
+        const defaultsIndex: number = configStr.search('[defaults]');
 
-        const defaultsIndex = configStr.search('[defaults]');
         if (defaultsIndex !== -1) {
             for (const property of properties) {
                 const match: RegExpMatchArray | null = configStr.substring(defaultsIndex).match(RegExp(`${property}[ \\t]?=[ \\t]?(.+)`));
--- a/src/commands/createWebApp/readAzConfig.ts
+++ b/src/commands/createWebApp/readAzConfig.ts
@@ -7,6 +7,11 @@ import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
 
+/**
+ * Takes any number of Azure config properties and returns their configured values
+ * found in `$AZURE_CONFIG_DIR/config`.
+ * Azure CLI configuration docs: https://aka.ms/AA64syh
+ */
 export async function readAzConfig(...properties: AzConfigProperties[]): Promise<AzConfig> {
     const config: AzConfig = {};
     const configPath: string = path.join(os.homedir(), '.azure', 'config');
--- a/src/commands/createWebApp/readAzConfig.ts
+++ b/src/commands/createWebApp/readAzConfig.ts
@@ -3,47 +3,59 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import * as fse from 'fs-extra';
-import * as os from 'os';
-import * as path from 'path';
+import { workspace } from 'vscode';
 import { IAppServiceWizardContext } from 'vscode-azureappservice';
+import { cpUtils } from '../../utils/cpUtils';
 
 /**
- * Takes any number of Azure config properties and returns their configured values
- * found in `$AZURE_CONFIG_DIR/config`.
+ * Takes any number of Azure config properties to read and returns their configured values.
  * Azure CLI configuration docs: https://aka.ms/AA64syh
  */
-export async function readAzConfig(wizardContext: IAppServiceWizardContext, ...properties: AzConfigProperties[]): Promise<AzConfig> {
+export async function readAzConfig(wizardContext: IAppServiceWizardContext, ...propertiesToRead: AzConfigProperties[]): Promise<AzConfig> {
     const config: AzConfig = {};
-    const configPath: string = path.join(os.homedir(), '.azure', 'config');
+    const azCommandCheck: cpUtils.ICommandResult = await cpUtils.tryExecuteCommand(undefined, undefined, 'az --version');
 
-    if (await fse.pathExists(configPath)) {
-        const configStr: string = await fse.readFile(configPath, 'utf-8');
-        const defaultsIndex: number = configStr.search('[defaults]');
+    if (azCommandCheck.code === 0) {
+        const workspacePath: string | undefined = workspace.workspaceFolders ? workspace.workspaceFolders[0].uri.fsPath : undefined;
+        const azGlobalConfigListing = <IAzConfigListing | undefined>JSON.parse(await cpUtils.executeCommand(undefined, undefined, 'az configure --list-defaults --output json'));
+        const azLocalConfigListing = <IAzConfigListing | undefined>JSON.parse(await cpUtils.executeCommand(undefined, workspacePath, 'az configure --list-defaults --output json --scope local'));
 
-        if (defaultsIndex !== -1) {
-            for (const property of properties) {
-                const match: RegExpMatchArray | null = configStr.substring(defaultsIndex).match(RegExp(`${property}[ \\t]?=[ \\t]?(.+)`));
+        readAzConfigListing(wizardContext, propertiesToRead, azGlobalConfigListing, config);
 
-                if (match) {
-                    config[property] = match[1].trim();
-                    wizardContext.telemetry.properties[`azConfig-${property}`] = 'true';
-                }
-            }
-        }
+        // Default to local config values if they exist
+        readAzConfigListing(wizardContext, propertiesToRead, azLocalConfigListing, config);
     }
 
     return config;
 }
 
+function readAzConfigListing(wizardContext: IAppServiceWizardContext, propertiesToRead: AzConfigProperties[], azConfigListing: IAzConfigListing | undefined, config: AzConfig): void {
+    if (azConfigListing) {
+        for (const configItem of azConfigListing) {
+            for (const propertyToRead of propertiesToRead) {
+                if (configItem.name === propertyToRead) {
+                    config[propertyToRead] = configItem.value;
+                    wizardContext.telemetry.properties[`hasAz${propertyToRead.charAt(0).toUpperCase() + propertyToRead.slice(1)}Default`] = 'true';
+                }
+            }
+        }
+    }
+}
+
 export type AzConfig = {
     group?: string;
-    plan?: string;
     location?: string;
 };
 
 export enum AzConfigProperties {
     group = 'group',
-    plan = 'plan',
     location = 'location'
 }
+
+interface IAzConfigItem {
+    name: string;
+    source: string;
+    value: string;
+}
+
+interface IAzConfigListing extends Array<IAzConfigItem> { }
--- a/src/commands/createWebApp/readAzConfig.ts
+++ b/src/commands/createWebApp/readAzConfig.ts
@@ -3,59 +3,60 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { workspace } from 'vscode';
+import { WorkspaceFolder } from 'vscode';
 import { IAppServiceWizardContext } from 'vscode-azureappservice';
 import { cpUtils } from '../../utils/cpUtils';
+import { getSingleRootWorkspace } from '../../utils/workspace';
+import { IDeployWizardContext } from '../deploy/IDeployWizardContext';
 
 /**
  * Takes any number of Azure config properties to read and returns their configured values.
  * Azure CLI configuration docs: https://aka.ms/AA64syh
  */
-export async function readAzConfig(wizardContext: IAppServiceWizardContext, ...propertiesToRead: AzConfigProperties[]): Promise<AzConfig> {
+export async function readAzConfig(wizardContext: IAppServiceWizardContext & Partial<IDeployWizardContext>, ...propertiesToRead: AzConfigProperty[]): Promise<AzConfig> {
     const config: AzConfig = {};
-    const azCommandCheck: cpUtils.ICommandResult = await cpUtils.tryExecuteCommand(undefined, undefined, 'az --version');
+    const workspaceFolder: WorkspaceFolder | undefined = getSingleRootWorkspace(wizardContext);
+    const workspacePath: string | undefined = workspaceFolder ? workspaceFolder.uri.fsPath : undefined;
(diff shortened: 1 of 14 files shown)


src/commands/createWebApp/readAzConfig.ts
```diff
@@ -0,0 +1,42 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
+import * as os from 'os';
+import * as path from 'path';
+
+export async function readAzConfig(...properties: AzConfigProperties[]): Promise<AzConfig> {
```
nturinski: A comment clarifying what this is doing and where it is reading the config from would be helpful, I think.

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -12,28 +12,32 @@ import { createAzureClient, DialogResponses, IActionContext } from "vscode-azure
...
+    const config: AzConfig = await readAzConfig(AzConfigProperties.group, AzConfigProperties.plan);
+
     // this should always be set when in the basic creation scenario
     const location: Location = nonNullProp(wizardContext, 'location');
-    const defaultName: string = `appsvc_${wizardContext.newSiteOS}_${location.name}`;
+    const defaultGroupName: string = config.group ? config.group : `appsvc_${wizardContext.newSiteOS}_${location.name}`;
+    const defaultPlanName: string = config.plan ? config.plan : defaultGroupName;
 
-    const asp: AppServicePlan | null = await getAppServicePlan(wizardContext, defaultName, defaultName);
+    const asp: AppServicePlan | null = await getAppServicePlan(wizardContext, defaultGroupName, defaultPlanName);
```
nturinski: Have you tried this test case?
- Create a plan with your default azconfig plan name in a random rg
- Try to create a new app with your azconfig plan name, but don't specific an azconfig rgName

wwlorey: This case passes. It allows the same plan name in different resource groups. Setting the azconfig plan but not the rg defaults to use the `appsvc_<os>_<location>` rg name.

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -12,28 +12,32 @@ import { createAzureClient, DialogResponses, IActionContext } from "vscode-azure
 import { ext } from "../../extensionVariables";
 import { nonNullProp } from "../../utils/nonNull";
 import { getWorkspaceSetting, updateGlobalSetting } from "../../vsCodeConfig/settings";
+import { AzConfig, AzConfigProperties, readAzConfig } from "./readAzConfig";
 
 const maxNumberOfSites: number = 3;
 
 export async function setDefaultRgAndPlanName(wizardContext: IAppServiceWizardContext, siteNameStep: SiteNameStep): Promise<void> {
+    const config: AzConfig = await readAzConfig(AzConfigProperties.group, AzConfigProperties.plan);
```
nturinski: I think it's worth getting some telemetry on how often this is used.

src/commands/createWebApp/setAppWizardContextDefault.ts
```diff
@@ -43,7 +44,8 @@ export async function setAppWizardContextDefault(wizardContext: IAppServiceWizar
 
     if (!wizardContext.advancedCreation) {
         if (!wizardContext.location) {
-            await LocationListStep.setLocation(wizardContext, 'centralus');
+            const config: AzConfig = await readAzConfig(wizardContext, AzConfigProperties.location);
+            await LocationListStep.setLocation(wizardContext, config.location ? config.location : 'centralus');
         }
 
         if (!wizardContext.newPlanSku) {
```
EricJizbaMSFT: Are there az cli defaults for sku, runtime, or OS?

src/commands/createWebApp/readAzConfig.ts
```diff
@@ -0,0 +1,49 @@
...
+import { IAppServiceWizardContext } from 'vscode-azureappservice';
+
+/**
+ * Takes any number of Azure config properties and returns their configured values
+ * found in `$AZURE_CONFIG_DIR/config`.
+ * Azure CLI configuration docs: https://aka.ms/AA64syh
+ */
+export async function readAzConfig(wizardContext: IAppServiceWizardContext, ...properties: AzConfigProperties[]): Promise<AzConfig> {
+    const config: AzConfig = {};
+    const configPath: string = path.join(os.homedir(), '.azure', 'config');
```
EricJizbaMSFT: The az cli actually provides a command to list out defaults. For example:
```
> az configure --list-defaults --output json
[
  {
    "name": "group",
    "source": "/Users/erijiz/.azure/config",
    "value": "myRG"
  },
  {
    "name": "location",
    "source": "/Users/erijiz/.azure/config",
    "value": "westus"
  }
]
```

First of all, this just looks easier to parse than all of your RegExp stuff. Second of all, it is more robust if the config isn't stored in the default manner. For example, the following command will add a config file in my current folder (similar to how VS Code has workspace settings instead of user settings):
```
> az configure --defaults web='emjtestdefault1' --scope local
> az configure --list-defaults --output json --scope local
[
  {
    "name": "web",
    "source": "/Users/erijiz/Repos/vscode-azurefunctions/.azure/config",
    "value": "emjtestdefault1"
  },
  {
    "name": "group",
    "source": "/Users/erijiz/.azure/config",
    "value": "myRG"
  },
  {
    "name": "location",
    "source": "/Users/erijiz/.azure/config",
    "value": "westus"
  }
]
```

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -12,28 +12,32 @@ import { createAzureClient, DialogResponses, IActionContext } from "vscode-azure
...
 
 const maxNumberOfSites: number = 3;
 
 export async function setDefaultRgAndPlanName(wizardContext: IAppServiceWizardContext, siteNameStep: SiteNameStep): Promise<void> {
+    const config: AzConfig = await readAzConfig(wizardContext, AzConfigProperties.group, AzConfigProperties.plan);
+
     // this should always be set when in the basic creation scenario
     const location: Location = nonNullProp(wizardContext, 'location');
-    const defaultName: string = `appsvc_${wizardContext.newSiteOS}_${location.name}`;
+    const defaultGroupName: string = config.group ? config.group : `appsvc_${wizardContext.newSiteOS}_${location.name}`;
```
EricJizbaMSFT: So if I have a default plan set:
```
> az configure --list-defaults --output json
[
  {
    "name": "plan",
    "source": "/Users/erijiz/.azure/config",
    "value": "emjtest1"
  }
]
```

I still see this when I try to create a webapp:
```
> az webapp create --name emjtest123 --resource-group emjtest123
az webapp create: error: the following arguments are required: --plan/-p
```

It seems like the cli doesn't use that default when creating? Can you double check the az cli's behavior and make sure you match it in terms of new vs. existing resources

src/commands/createWebApp/readAzConfig.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Takes any number of Azure config properties and returns their configured values
+ * found in `$AZURE_CONFIG_DIR/config`.
+ * Azure CLI configuration docs: https://aka.ms/AA64syh
+ */
+export async function readAzConfig(...properties: AzConfigProperties[]): Promise<AzConfig> {
+    const config: AzConfig = {};
+    const configPath: string = path.join(os.homedir(), '.azure', 'config');
+
+    if (await fse.pathExists(configPath)) {
+        const configStr: string = await fse.readFile(configPath, 'utf-8');
```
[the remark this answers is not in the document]
wwlorey: If we don't put an encoding, `fse.readFile` returns an array of ASCII numbers that have to be translated into characters. We could do that processing but I thought this was easier

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -12,28 +12,32 @@ import { createAzureClient, DialogResponses, IActionContext } from "vscode-azure
...
+    const defaultPlanName: string = config.plan ? config.plan : defaultGroupName;
```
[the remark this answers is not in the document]
wwlorey: I think the latter makes more sense

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -48,16 +52,16 @@ export async function setDefaultRgAndPlanName(wizardContext: IAppServiceWizardCo
             }
 
             // otherwise create a new rg and asp
-            wizardContext.newResourceGroupName = wizardContext.newResourceGroupName || await siteNameStep.getRelatedName(wizardContext, defaultName);
+            wizardContext.newResourceGroupName = wizardContext.newResourceGroupName || await siteNameStep.getRelatedName(wizardContext, defaultGroupName);
```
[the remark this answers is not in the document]
wwlorey: A new rg wouldn't be created until the app service quota is met which looks to be 3 app services in one rg

src/commands/createWebApp/setAppWizardContextDefault.ts
wwlorey: There are not

src/commands/createWebApp/readAzConfig.ts
wwlorey: Is there a way to get the output of that command from within the extension?

src/commands/createWebApp/setDefaultRgAndPlanName.ts
wwlorey: So plans aren't supported in the docs but the CLI lets you configure the plan anyway. We can remove support for checking plans if it's going to cause issues with commands like `list-defaults`

src/commands/createWebApp/readAzConfig.ts
EricJizbaMSFT: Yeah something like this: https://github.com/microsoft/vscode-azurefunctions/blob/master/src/utils/cpUtils.ts

If app service repo doesn't have any child_process utils yet, it's fine to just copy that file

src/commands/createWebApp/setDefaultRgAndPlanName.ts
EricJizbaMSFT: I mean "--list-defaults" works just fine (it listed the plan). The problem is it wasn't actually used (from my brief testing) when I ran `az webapp create`

src/commands/createWebApp/setDefaultRgAndPlanName.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Good point. That being said, I believe it's 3 app services in one _plan_ - I don't think the rg matters as much. Which means it would be more important to use `defaultPlanName` when calling `getRelatedName`

src/commands/createWebApp/readAzConfig.ts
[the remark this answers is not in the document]
EricJizbaMSFT: I think it returns a buffer - which you can just call `toString` on (without specifying the encoding)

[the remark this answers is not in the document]
EricJizbaMSFT: Like this:
https://github.com/microsoft/vscode-azurefunctions/blob/45e2aebdd3b384ed385dc57254b4c36b2f5e1d17/src/funcConfig/local.settings.ts#L75

src/commands/createWebApp/setDefaultRgAndPlanName.ts
wwlorey: Yeah default plans aren't formally supported. I can remove that

src/commands/createWebApp/setDefaultRgAndPlanName.ts
```diff
@@ -10,54 +10,64 @@ import { MessageItem } from "vscode";
...
             const allAppServicePlans: AppServicePlan[] = await client.appServicePlans.list();
-            const defaultPlans: AppServicePlan[] = allAppServicePlans.filter(plan => { return plan.name && plan.name.includes(defaultName); });
+            const defaultPlans: AppServicePlan[] = allAppServicePlans.filter(plan => { return plan.name && plan.name.includes(defaultPlanName); });
 
             // when using appServicePlans.list, the numOfSites are all set to 0 so individually get each plan and look for one with less than 3 sites
             for (const plan of defaultPlans) {
                 if (plan.name) {
-                    const fullPlanData: AppServicePlan | null = await getAppServicePlan(wizardContext, plan.name, plan.name);
+                    const groupName: string = getResourceGroupFromId(nonNullProp(plan, 'id'));
+                    const fullPlanData: AppServicePlan | null = await getAppServicePlan(wizardContext, groupName, plan.name);
```
EricJizbaMSFT: I feel like this should check `if (groupName.includes(defaultGroupName))` - otherwise you might run into weird behavior if you have a plan named the same thing in different resource groups. Here are some repro steps:
1. Remove az cli defaults
1. Create a web app
1. Add az cli defaults for group name 'myGroup'
1. Create 4 web apps

Expected:
The 4th web app will create a new plan in group 'myGroup'

Actual:
The 4th web app creates a new plan in 'appsvc_windows_eastus' (or whatever) from step 1

src/utils/azure.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function getSubscriptionFromId(id: string): string {
+    return parseResourceId(id)[1];
+}
+
+export function getNameFromId(id: string): string {
+    return parseResourceId(id)[4];
+}
+
+export interface IBaseResourceWithName extends BaseResource {
```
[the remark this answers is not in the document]
EricJizbaMSFT: To be clear, I'm okay with you copying over functions like `getSubscriptionFromId` because they're related to `getResourceGroupFromId` which you actually need, but `IBaseResourceWithName` and `IResourceResult` are only used by `promptForResource` from the functions repo and I don't think app service will ever need that

src/commands/createWebApp/readAzConfig.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Takes any number of Azure config properties to read and returns their configured values.
+ * Azure CLI configuration docs: https://aka.ms/AA64syh
+ */
+export async function readAzConfig(wizardContext: IAppServiceWizardContext, ...propertiesToRead: AzConfigProperties[]): Promise<AzConfig> {
+    const config: AzConfig = {};
+    const azCommandCheck: cpUtils.ICommandResult = await cpUtils.tryExecuteCommand(undefined, undefined, 'az --version');
+
+    if (azCommandCheck.code === 0) {
+        const workspacePath: string | undefined = workspace.workspaceFolders ? workspace.workspaceFolders[0].uri.fsPath : undefined;
```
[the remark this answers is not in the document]
wwlorey: Unless I'm missing something, that logic always results in undefined in the context of basic create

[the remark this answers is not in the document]
EricJizbaMSFT: It's only relevant when you do basic create _during_ a deploy

But also that code is supposed to handle the case when it's undefined, and I'm not sure why it's not - which is what I was asking @nturinski about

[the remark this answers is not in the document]
nturinski: For some reason, I removed that code in this PR: https://github.com/microsoft/vscode-azureappservice/commit/2aeb49340e4a90154e7e16d9f1bbed41cacb1571

Not entirely sure what I was thinking 😓 

As for it always being undefined, it would have a value if the user started with "Deploy" and then clicked `+ Create Web App...` when prompted for a web app.

wwlorey:
`getRelatedName` won't always return a unique plan name across resource groups. Is it worth fixing this?

src/commands/createWebApp/readAzConfig.ts
```diff
@@ -0,0 +1,62 @@
...
+    try {
+        const azGlobalConfigListing = <IAzConfigItem[] | undefined>JSON.parse(await cpUtils.executeCommand(undefined, undefined, 'az configure --list-defaults --output json'));
+        const azLocalConfigListing = <IAzConfigItem[] | undefined>JSON.parse(await cpUtils.executeCommand(undefined, workspacePath, 'az configure --list-defaults --output json --scope local'));
+
+        readAzConfigListing(wizardContext, propertiesToRead, azGlobalConfigListing, config);
+
+        // Default to local config values if they exist
+        readAzConfigListing(wizardContext, propertiesToRead, azLocalConfigListing, config);
+    } catch (error) {
+        // Azure CLI not installed so no config to read
```
EricJizbaMSFT: This comment isn't entirely accurate - errors could happen for a lot of reasons (for example maybe they're on an old version of the cli that doesn't support `az configure`)

What about something like this:
```
// Suppress errors - this functionality should not block creating a web app
```

src/commands/createWebApp/setDefaultRgPlanNameLocation.ts
```diff
@@ -9,55 +9,71 @@ import { AppServicePlan } from "azure-arm-website/lib/models";
...
+import { getResourceGroupFromId } from '../../utils/azureUtils';
 import { nonNullProp } from "../../utils/nonNull";
 import { getWorkspaceSetting, updateGlobalSetting } from "../../vsCodeConfig/settings";
+import { AzConfig, AzConfigProperty, readAzConfig } from "./readAzConfig";
 
 const maxNumberOfSites: number = 3;
 
-export async function setDefaultRgAndPlanName(wizardContext: IAppServiceWizardContext, siteNameStep: SiteNameStep): Promise<void> {
-    // this should always be set when in the basic creation scenario
+export async function setDefaultRgPlanNameLocation(wizardContext: IAppServiceWizardContext, siteNameStep: SiteNameStep): Promise<void> {
```
EricJizbaMSFT: Not a big fan of this function/file name because we'll have to change it every time we change which defaults are included in this. Also it's confusing how we have this _and_ `setAppWizardContextDefault` and the names don't line up at all.

A few ideas:
- Option 1
    - setAppWizardContextDefault -> setPrePromptDefaults
    - setDefaultRgPlanNameLocation -> setPostPromptDefaults
- Option 2
    - setAppWizardContextDefault -> setCreateDefaults
    - setDefaultRgPlanNameLocation -> setBasicCreateDefaults
- Option 3
    - setAppWizardContextDefault -> setWebAppRecommendations
    - setDefaultRgPlanNameLocation -> setWebAppDefaults

src/commands/createWebApp/setDefaultRgPlanNameLocation.ts
```diff
@@ -9,55 +9,71 @@ import { AppServicePlan } from "azure-arm-website/lib/models";
...
+    const config: AzConfig = await readAzConfig(wizardContext, AzConfigProperty.group, AzConfigProperty.location);
```
EricJizbaMSFT: I would add a comment above this line saying something like this:
```
// Reading az config should always happen after prompting because it can cause a few seconds delay
```

Seems like something we could forget

nturinski:
🎉

Issue #1185, opened 2019-09-20
Don't prompt to "Add Config" when deploying .NET Core project with Local Git


XofoSol:
Repro steps:

1. Do a commit in your local git
2. Select Local git as your deployment source
3. Do the deploy to webapp.
4. VSCode tells you that you need a new configuration to do the deploy (it doesn't matter if you had already done another deploy before)
6. Pum.  Error.

Action: appService.Deploy
Error type: Error
Error Message: Cannot read property 'constructor' of undefined

Version: 0.16.0
OS: darwin
Product: Visual Studio Code
Product Version: 1.38.1
Language: es

<details>
<summary>Call Stack</summary>

```
Object.<anonymous> extension.bundle.js:323:357962
a extension.bundle.js:323:357598
process._tickCallback next_tick.js:68:7
```

</details>

XofoSol:
Update:
VSCode is creating a settings.json file with this info:
{
"appService.defaultWebAppToDeploy": "/subscriptions/{{subscriptionId}}/resourceGroups/{{ResourceGroup}}/providers/Microsoft.Web/sites/{{WebAppName}}",
    "appService.deploySubpath": "{{ProjectName}}/bin/Release/netcoreapp2.2/publish",
    "appService.preDeployTask": "publish-release"
}

If I delete this data or replace it with an empty json (like {}) it works again.

EricJizbaMSFT:
@XofoSol as a workaround, you can try disabling the warning by setting `appService.showPreDeployWarning` to `false`

XofoSol:
Nop, it does not work. Because is not a warning, is an error.

EricJizbaMSFT:
Sorry I wasn't clear - I'm talking about this warning:
![Screen Shot 2019-09-20 at 1 18 28 PM](https://user-images.githubusercontent.com/11282622/65356387-37955400-dba9-11e9-9eb8-5f76830f49ec.png)

If you click "Add Config" it adds the settings you mentioned ("appService.deploySubpath" and "appService.preDeployTask"). I believe those settings are causing your error, so you'll want to remove them and set `appService.showPreDeployWarning` to `false` to prevent adding that config again

XofoSol:
Ok, Thank you very much.  
Apparently it works.  But, like you said, you have to add this configuration in the file only after the first fail.

EricJizbaMSFT:
Yeah that's only meant as a temporary workaround to make sure you're unblocked.

[...]

PR #1190, opened 2019-09-27
Leverage new getDeployPath. Ignores subpath for local git deploys

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1185

Relies on https://github.com/microsoft/vscode-azuretools/pull/595

### DIFF
--- a/src/commands/deploy/IDeployWizardContext.ts
+++ b/src/commands/deploy/IDeployWizardContext.ts
@@ -8,8 +8,8 @@ import { IActionContext } from "vscode-azureextensionui";
 
 export interface IDeployWizardContext extends IActionContext {
     workspace: WorkspaceFolder;
-    deployFsPath: string;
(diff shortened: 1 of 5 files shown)


nturinski:
Should I have a message saying something like:
`WARNING: Ignoring subDeployPath "${subdeployPath}" for non-zip deploy.`

src/commands/deploy/IDeployWizardContext.ts
```diff
@@ -9,6 +9,7 @@ import { IActionContext } from "vscode-azureextensionui";
 export interface IDeployWizardContext extends IActionContext {
     workspace: WorkspaceFolder;
     deployFsPath: string;
+    deployFsSubpath: string;
```
EricJizbaMSFT: Can you update these names to make more sense as well

PR #1199, opened 2019-10-01
Add util function to grab singleRootWorkspace

nturinski:
@wwlorey You can use this for your PR as well.

### DIFF
--- a/src/utils/workspace.ts
+++ b/src/utils/workspace.ts
@@ -6,6 +6,7 @@
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { IDeployWizardContext } from '../commands/deploy/IDeployWizardContext';
 import { ext } from '../extensionVariables';
 import { isPathEqual, isSubpath } from '../utils/pathUtils';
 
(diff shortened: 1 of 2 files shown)


src/utils/workspace.ts
```diff
@@ -77,3 +77,8 @@ export function mapFilesToQuickPickItems(files: vscode.Uri[]): IAzureQuickPickIt
         };
     });
 }
+
+export function getSingleRootWorkspace(): vscode.WorkspaceFolder | undefined {
```
EricJizbaMSFT: How about you add `context: Partial<IDeployWizardContext>` as a parameter and then Will could re-use even more code

nturinski: Are you saying to set the context directly instead?

EricJizbaMSFT: Change this function to something like this:
```
return wizardContext.workspace || (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length === 1 ? vscode.workspace.workspaceFolders[0] : undefined);
```

And call it above like this:
```
const workspaceForRecommendation: WorkspaceFolder | undefined = getSingleRootWorkspace(wizardContext);
```

Issue #1209, opened 2019-10-07
 Error: ETag does not represent the latest state of the resource.


straightcore:
This error has been thrown during an updating '"App Service" azure resource. 
I just created this webapp by Azure portal. And I updated the hostingstart.html file directly by VS Code.

Repro steps:

1. Go to Azure (VS Code menu)
2. Go to App Service tab
 2.1 Go to Subcription
 2.2 Go to WebAPP ( myWebAppInAzure)
3. Open file: myWebAppInAzure -> Files -> hostingstart.html
4. Modify the file
5. Save the file
6. Click on Updating button on the popup window.

Action: appService.fileEditor.onDidSaveTextDocument
Error type: 412
Error Message: ETag does not represent the latest state of the resource.

Version: 0.16.0
OS: win32
Product: Visual Studio Code
Product Version: 1.38.1
Language: en

<details>
<summary>Call Stack</summary>

```
module.exports.o.pipeline extension.bundle.js:323:803050
s extension.bundle.js:218:111699
s extension.bundle.js:157:26069
extension.bundle.js:157:27311extension.bundle.js:157:27311
o extension.bundle.js:164:7075
extension.bundle.js:157:24982extension.bundle.js:157:24982
$.module.exports.t.requestLibrarySink.t.headersOnly.t.streamedResponse.t.body.function.t.body.pipe.s.pipe.a [as _callback] extension.bundle.js:1:576747
$.module.exports.$.init.t._callback.t.callback.t.callback extension.bundle.js:194:5449
$.emit events.js:182:13
$.EventEmitter.emit domain.js:442:20
$.<anonymous> extension.bundle.js:194:18309
$.emit events.js:187:15
$.EventEmitter.emit domain.js:442:20
IncomingMessage.<anonymous> extension.bundle.js:194:17141
Object.onceWrapper events.js:273:13
IncomingMessage.emit events.js:187:15
IncomingMessage.EventEmitter.emit domain.js:442:20
endReadableNT _stream_readable.js:1092:12
asyncWrap async-listener/glue.js:188:31
process._tickCallback next_tick.js:63:19
```

</details>

nturinski:
Hi @straightcore

Did you happen to edit this file anywhere else?  Or how long did you wait between steps?  The Etag was stale which means you were not working with the most update to date file.  If you redownload the file and try saving it again, does that resolve the issue?

PR #1240, opened 2019-10-17
Leverage shared file tree items 🗄🌲

EricJizbaMSFT:
Depends on https://github.com/microsoft/vscode-azuretools/pull/613

Almost exactly the same experience, just from the shared package.


src/commands/startStreamingLogs.ts
```diff
@@ -6,19 +6,16 @@
...
 import { WebAppTreeItem } from '../explorer/WebAppTreeItem';
 import { ext } from '../extensionVariables';
 import { enableFileLogging } from './enableFileLogging';
 
-export async function startStreamingLogs(context: IActionContext, node?: SiteTreeItem | LogStreamTreeItem): Promise<void> {
+export async function startStreamingLogs(context: IActionContext, node?: SiteTreeItem): Promise<void> {
     let siteTreeItem: SiteTreeItem;
 
     if (!node) {
         siteTreeItem = <WebAppTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue, context);
```
nturinski: Was it okay to remove the check for getting a SiteTreeItem?  What if they enter through the LogStreamTreeItem?

src/explorer/editors/FileEditor.ts
```diff
@@ -40,7 +41,9 @@ export class FileEditor extends BaseEditor<FileTreeItem> {
             throw new Error('Cannot update file after it has been closed.');
         }
         const localFile: Readable = fs.createReadStream(vscode.window.activeTextEditor.document.uri.fsPath);
-        node.etag = await putFile(node.root.client, localFile, node.path, nonNullProp(node, 'etag'));
+        let etag: string = nonNullValue(this._etags.get(node.fullId), 'etag');
+        etag = await putFile(node.root.client, localFile, node.path, etag);
```
nturinski: Is this for https://github.com/microsoft/vscode-azureappservice/issues/1209?

src/extension.ts
```diff
@@ -224,19 +221,19 @@ export async function activateInternal(
...
         registerEvent('appService.fileEditor.onDidSaveTextDocument', vscode.workspace.onDidSaveTextDocument, async (actionContext: IActionContext, doc: vscode.TextDocument) => { await fileEditor.onDidSaveTextDocument(actionContext, context.globalState, doc); });
-        registerCommand('appService.EnableFileLogging', async (actionContext: IActionContext, node?: SiteTreeItem | FolderTreeItem) => {
+        registerCommand('appService.EnableFileLogging', async (actionContext: IActionContext, node?: SiteTreeItem | LogFilesTreeItem) => {
             if (!node) {
                 node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue, actionContext);
             }
 
-            if (node instanceof FolderTreeItem) {
-                // If the entry point was the Files/Log Files node, pass the parent as that's where the logic lives
+            if (node instanceof LogFilesTreeItem) {
```
nturinski: Is this not the case anymore?

src/commands/startStreamingLogs.ts
EricJizbaMSFT: I modified commandArgs so that it would pass the parent instead of the LogStreamTreeItem itself:
https://github.com/microsoft/vscode-azuretools/pull/613/files#diff-81088792bebe528022540e6bb9a4512aR34

src/explorer/editors/FileEditor.ts
EricJizbaMSFT: I moved the etag stuff into this file because I don't need it in the shared package (everything is read only for functions). Plus it was always a little weird on the tree item (hence the [comment](https://github.com/microsoft/vscode-azureappservice/pull/1240/files#diff-9392209e5d5f170c97511a539a415a06L14) "cannot create etag on creation due to Kudu API calls"). The bug still repros after my changes

src/extension.ts
EricJizbaMSFT: Can you expand on what you mean by "this"? If you're talking about how I deleted "Files/" from the comment below, I don't think we ever had "Enable File Logging" on that node - only the "Logs" node. See https://github.com/microsoft/vscode-azureappservice/blob/master/package.json#L546

EricJizbaMSFT:
> When is ext.prefix actually set?

https://github.com/microsoft/vscode-azureappservice/pull/1240/files#diff-9f20fc1ad477b17ef4d963edfe6c7066R20

src/extension.ts
nturinski: Yeah, that's what I meant.  That makes sense-- FolderTreeItem used to be both the Files and Log Files, I believe.

PR #1253, opened 2019-10-24
Refactor deploy command to get rid of tslint disable

### LINKED ISSUE #1112: Refactor Deploy code to get rid of `max-func-body-length cyclomatic-complexity`

nturinski:
A few things were refactored out of `deploy.ts` to get rid of the cyclomatic-complexity rule.

Pretty much anything that required UI has been moved out.  `getDeployNode` has been refactored from `getDefaultWebApp` and follows similar behavior to Azure Functions, though we have to account for the `defaultWebApp` setting.

Fixes #1112


nturinski:
bump

src/commands/deploy/getDeployNode.ts
```diff
@@ -0,0 +1,51 @@
...
+    node: SiteTreeItem;
+    isNewWebApp: boolean;
+}
+
+export async function getDeployNode(context: IDeployContext, target: Uri | string | SiteTreeItem | undefined, confirmDeployment: boolean): Promise<IDeployNode> {
+    const defaultWebAppId: string | undefined = getWorkspaceSetting(configurationSettings.defaultWebAppToDeploy, context.workspace.uri.fsPath);
+    let node: SiteTreeItem | undefined;
+
+    // if the entry point for deploy was via "Deploy" after creating, it is also a new web app (so confirmDeployment would be false)
+    let isNewWebApp: boolean = !confirmDeployment;
```
EricJizbaMSFT: Can we change the signature of `deploy` to be something like this:
```typescript
export async function deploy(context: IActionContext, target?: vscode.Uri | SiteTreeItem | undefined, isTargetNewWebApp?: boolean): Promise<void> {
```
(I tried using `isNewWebApp` as the name first, but that causes a weird conflict)

For two reasons:
1. I don't like the name `confirmDeployment` in this function. Perhaps there are other reasons we let users skip confirming deployment in the future and thus it wouldn't be accurate to do `isNewWebApp: boolean = !confirmDeployment;`
1. Then you can simplify extension.ts like this:
```typescript
// before
registerCommand('appService.Deploy', async (actionContext: IActionContext, target?: vscode.Uri | WebAppTreeItem | undefined) => {
    await deploy(actionContext, true, target);
});


// after
registerCommand('appService.Deploy', deploy);
```

Issue #1154, opened 2019-09-16
ETag does not represent the latest state of the resource (Auto upload on save error)


piphansdk:
Repro steps:
1. Open a document from the AzureAppService integration in visual studio code (Mac) 
2. Edit and save the document a few times
3. VScode cannot upload the file due to error message: ETag does not represent the latest state of the resource.
4. File must be closed and reopened/redownloaded. 

(Bug is found for the following filetypes: ".js, .php, .html, .css, .htaccess, .xml". No further filetypes has been testet).

Action: appService.fileEditor.onDidSaveTextDocument
Error type: 412
Error Message: ETag does not represent the latest state of the resource.

Version: 0.15.0
OS: darwin
Product: Visual Studio Code
Product Version: 1.38.1
Language: en

<details>
<summary>Call Stack</summary>

```
module.exports.o.pipeline extension.bundle.js:332:788567
s extension.bundle.js:262:112563
s extension.bundle.js:157:26036
extension.bundle.js:157:27276
o extension.bundle.js:208:7054
extension.bundle.js:157:24957
$.module.exports.t.requestLibrarySink.t.headersOnly.t.streamedResponse.t.body.function.t.body.pipe.s.pipe.a [as _callback] extension.bundle.js:1:534379
$.module.exports.$.init.t._callback.t.callback.t.callback extension.bundle.js:238:5461
$.emit events.js:182:13
$.<anonymous> extension.bundle.js:238:18274
$.emit events.js:187:15
IncomingMessage.<anonymous> extension.bundle.js:238:17116
Object.onceWrapper events.js:273:13
IncomingMessage.emit events.js:187:15
endReadableNT _stream_readable.js:1092:12
process._tickCallback next_tick.js:63:19
```

</details>

nturinski:
@piphansdk Thanks for reporting and providing detailed repro steps.  I'll look into this soon.

Issue #287, opened 2018-01-31
Save All (Ctrl+K S) will upload all your files to the cloud


nturinski:
Repro:
1.) Click on 2+ files on the tree view to open multiple files
2.) Edit 1 or more files
3.) Save All
4.) All files that are opened are now replaced with the active text document

Expected:
Each file will upload the appropriate data to the cloud
Actual:
Each file is updated with the current active document, overwriting the old files

PR #1264, opened 2019-10-31
Fix stale 🍞 etag 🏷️ and updating with wrong text 📜

nturinski:
Turns out updateData was always called with two parameters-- the text of the document itself.  So it made all that activeEditor logic unnecessary.

Fixes https://github.com/microsoft/vscode-azureappservice/issues/1154
Fixes https://github.com/microsoft/vscode-azureappservice/issues/287

### DIFF
--- a/src/explorer/editors/FileEditor.ts
+++ b/src/explorer/editors/FileEditor.ts
@@ -4,10 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { FileTreeItem, getFile, IFileResult, putFile } from 'vscode-azureappservice';
-import { BaseEditor } from 'vscode-azureextensionui';
+import { BaseEditor, IParsedError, parseError } from 'vscode-azureextensionui';
 import { ext } from '../../extensionVariables';
+import { nonNullValue } from '../../utils/nonNull';
 
 export class FileEditor extends BaseEditor<FileTreeItem> {
+    private _etags: Map<string, string> = new Map<string, string>();
 
     constructor() {
         super(`${ext.prefix}showSavePrompt`);
@@ -23,6 +25,7 @@ export class FileEditor extends BaseEditor<FileTreeItem> {
 
     public async getData(node: FileTreeItem): Promise<string> {
         const result: IFileResult = await getFile(node.root.client, node.path);
+        this._etags.set(node.fullId, result.etag);
         return result.data;
     }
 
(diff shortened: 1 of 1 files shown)


src/explorer/editors/FileEditor.ts
```diff
@@ -3,18 +3,14 @@
...
 import { BaseEditor } from 'vscode-azureextensionui';
-import { nonNullValue } from '../../utils/nonNull';
+import { ext } from '../../extensionVariables';
 
 export class FileEditor extends BaseEditor<FileTreeItem> {
-    private _etags: Map<string, string> = new Map<string, string>();
 
     constructor() {
-        super('appService.showSavePrompt');
+        super(`${ext.prefix}showSavePrompt`);
```
EricJizbaMSFT: Missing a dot in between

src/explorer/editors/FileEditor.ts
```diff
@@ -36,14 +31,10 @@ export class FileEditor extends BaseEditor<FileTreeItem> {
...
-    public async updateData(node: FileTreeItem): Promise<string> {
-        if (!vscode.window.activeTextEditor) {
-            throw new Error('Cannot update file after it has been closed.');
-        }
-        const localFile: Readable = fs.createReadStream(vscode.window.activeTextEditor.document.uri.fsPath);
-        let etag: string = nonNullValue(this._etags.get(node.fullId), 'etag');
-        etag = await putFile(node.root.client, localFile, node.path, etag);
-        this._etags.set(node.fullId, etag);
+    public async updateData(node: FileTreeItem, data: string): Promise<string> {
+        // because the etag can become stale, get the latest before updating
```
EricJizbaMSFT: Discussed offline. Need to be more careful with etag because we don't want to overwrite changes that might've already been made in the cloud

nturinski:
So I think the issue with the etag was coming from the fact that were awaiting the [kudu.putFile](https://github.com/microsoft/vscode-azuretools/blob/31f5d7f87dd0a56047b1951ca02534ed38652508/appservice/src/putFile.ts#L28) command, but it didn't seem like that was guaranteed to be the same etag as Azure.  Now I just get the file directly from Azure again after updating to ensure that we have the most updated ETag,

Also changed the error messaging to this:
image

src/explorer/editors/FileEditor.ts
```diff
@@ -36,13 +34,19 @@ export class FileEditor extends BaseEditor<FileTreeItem> {
...
-        }
-        const localFile: Readable = fs.createReadStream(vscode.window.activeTextEditor.document.uri.fsPath);
+    public async updateData(node: FileTreeItem, data: string): Promise<string> {
         let etag: string = nonNullValue(this._etags.get(node.fullId), 'etag');
-        etag = await putFile(node.root.client, localFile, node.path, etag);
+        try {
+            await putFile(node.root.client, data, node.path, etag);
+        } catch (error) {
+            const parsedError: IParsedError = parseError(error);
+            if (parsedError.message === 'ETag does not represent the latest state of the resource.') {
```
EricJizbaMSFT: This is too dependent on the exact error message. It could change by one letter or be different for other languages. One or both of these checks would be better IMO:
```
parsedError.errorType === '412' && /etag/i.test(parsedError.message)
```

Issue #1228, opened 2019-10-11
SCM_DO_BUILD warning looks for, generates .deployment in wrong folder


nickwalkmsft:
Discovered when using Web Template Studio (generates a full stack app, with `appService.deploySubpath` in the workspace settings set to `server`).

Deploying a newly generated WebTS Node app pops the notification about running build commands on the server, but shouldn't, because the folder being zipdeployed (`/server`) already contains a `.deployment` with `SCM_DO_BUILD_DURING_DEPLOYMENT=true`.

If you click Yes on the notification, it creates a new `.deployment` at the root of the workspace that doesn't get used for anything.

Linking #1167 - no direct relation but same feature and the last comment indicates that there might be some potentially related discussion.

nturinski:
@nickwalkmsft Thanks for reporting this.  We should look/generate the `.deployment` file in the deploySubpath

nturinski:
This should be resolved with the latest release, v0.16.2.

Please let me know if it hasn't! 😃

nickwalkmsft:
Confirmed fixed. Thanks!

PR #1272, opened 2019-11-05
Use effectiveDeployFsPath rather than workspace

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1228

We were using the old method of getting the workspace's fsPath, but because of subdeploy path settings, this could be inaccurate (and then not deployed).

Though that being said, I've been seeing the messaging `Ignoring .deployment file...` when deploying to a Linux node.js web app, so it doesn't seem like it's being respected anymore?


nturinski:
Tracking here: https://github.com/microsoft/vscode-azureappservice/issues/1274

PR #1273, opened 2019-11-05
Add WebAppCreateStep

EricJizbaMSFT:
Based on https://github.com/microsoft/vscode-azuretools/pull/625

Should be mostly copying code with no functional differences.


nturinski:
🙇

Issue #797, opened 2019-01-03
Detect which folder is a virtual environment to be ignored for Python projects with scm build enabled


nturinski:
We're using a list of general names for virtual environments provided by GitHub .gitignore file to generate this list, but we should be able to detect if the user has a virtual environment by seeing if the activate script exists and then ignoring that folder for deployment.

Issue #1277, opened 2019-11-07
Refactor scmBuild commands out of SiteTreeItem


nturinski:
SiteTreeItem is getting kind of cluttered and while only SiteTreeItems will be using this function, I don't think it's immediately obvious why it's living in that file.

PR #1278, opened 2019-11-07
🕵  Detect venvs and add ➕ them to the gitignore 🙈

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/797

Rather than just using a generic list of virtual environments, now we look for venvs by seeing if the activate command works.

I think it's time to refactor most of this scm stuff out of SiteTreeItem though https://github.com/microsoft/vscode-azureappservice/issues/1277

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -31,9 +31,3 @@ export namespace AppServiceDialogResponses {
 }
 
 export const envFileName: string = '.env';
-
-export enum Platform {
-    Windows = 'win32',
-    MacOS = 'darwin',
-    Linux = 'linux'
-}
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -31,3 +31,9 @@ export namespace AppServiceDialogResponses {
 }
 
 export const envFileName: string = '.env';
+
+export enum Platform {
+    Windows = 'win32',
+    MacOS = 'darwin',
+    Linux = 'linux'
+}
--- a/src/explorer/SiteTreeItem.ts
+++ b/src/explorer/SiteTreeItem.ts
@@ -220,15 +220,31 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
                 ignoredFolders = ['node_modules{,/**}'];
                 break;
             case LinuxRuntimes.python:
-                const venvsfsPaths: string[] = (await venvUtils.getExistingVenvs(fsPath)).map(venvPath => `${venvPath}{,/**}`);
+                let venvFsPaths: string[];
+                try {
+                    venvFsPaths = (await venvUtils.getExistingVenvs(fsPath)).map(venvPath => `${venvPath}{,/**}`);
+                } catch (error) {
+                    // if there was an error here, don't block-- just assume none could be detected
+                    venvFsPaths = [];
+                }
+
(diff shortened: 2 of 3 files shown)


src/constants.ts
```diff
@@ -31,3 +31,9 @@ export namespace AppServiceDialogResponses {
 }
 
 export const envFileName: string = '.env';
+
+export enum Platform {
```
EricJizbaMSFT: Yeah I've been meaning to get rid of this constant in Functions. The intellisense/validation are a lot better than I remember in the past:
Screen Shot 2019-11-07 at 2 14 40 PM
Screen Shot 2019-11-07 at 2 15 11 PM

src/explorer/SiteTreeItem.ts
```diff
@@ -212,21 +213,22 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
 
-    private getIgnoredFoldersForDeployment(runtime: string): string[] {
+    private async getIgnoredFoldersForDeployment(fsPath: string, runtime: string): Promise<string[]> {
         let ignoredFolders: string[];
         switch (runtime) {
             case LinuxRuntimes.node:
                 ignoredFolders = ['node_modules{,/**}'];
                 break;
             case LinuxRuntimes.python:
+                const venvsfsPaths: string[] = (await venvUtils.getExistingVenvs(fsPath)).map(venvPath => `${venvPath}{,/**}`);
```
EricJizbaMSFT: If this fails (maybe user doesn't have access to a directory?) I feel like we should catch that rather than block deploy

src/explorer/SiteTreeItem.ts
```diff
@@ -212,21 +213,22 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
                 // Byte-compiled / optimized / DLL files
                 ignoredFolders = ['__pycache__{,/**}', '*.py[cod]', '*$py.class',
                     // Distribution / packaging
                     '.Python{,/**}', 'build{,/**}', 'develop-eggs{,/**}', 'dist{,/**}', 'downloads{,/**}', 'eggs{,/**}', '.eggs{,/**}', 'lib{,/**}', 'lib64{,/**}', 'parts{,/**}', 'sdist{,/**}', 'var{,/**}',
-                    'wheels{,/**}', 'share/python-wheels{,/**}', '*.egg-info{,/**}', '.installed.cfg', '*.egg', 'MANIFEST',
-                    // Environments
-                    '.env{,/**}', '.venv{,/**}', 'env{,/**}', 'venv{,/**}', 'ENV{,/**}', 'env.bak{,/**}', 'venv.bak{,/**}'];
+                    'wheels{,/**}', 'share/python-wheels{,/**}', '*.egg-info{,/**}', '.installed.cfg', '*.egg', 'MANIFEST'];
+                // Virtual Environments
+                ignoredFolders = ignoredFolders.concat(venvsfsPaths);
```
EricJizbaMSFT: I would include the previous list of virtual environments that you had. It will help if you fail to detect a venv, if the venv doesn't exist yet, or if someone on a different machine (with the same repo) uses a different venv name

nturinski:
@EricJizbaMSFT There's enough new logic that I'd like to double-check it's okay 👍

Issue #1161, opened 2019-09-17
There is no auto-refresh for web app after installing the Cosmos DB extension


v-ruizh:
**OS:** All
**Build Version:** [20190916.20](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=10047)
**Regression:** Not a Regression

**Repro Steps:**

1. Make sure there is no Cosmos DB Extension installed.
2. Expand one Web app->Connections->Cosmos DB node.
3. Click 'Install Cosmos DB Extension...' to install Cosmos DB extension.
4. Refresh the Web App and check the Cosmos DB node.

**Expect:**
It changes to 'Add Cosmos DB Extension...'.

**Actual:**
Refresh web app doesn't work, it still shows 'Install Cosmos DB Extension...'.
![image](https://user-images.githubusercontent.com/45614674/65035013-c708f000-d97a-11e9-8c35-d91af403314d.png)

**More Info:**
Refresh the App Service explorer can fix this issue.

nturinski:
Offline discussion: Maybe poll for a little bit after.

v-dantch:
Verified on the [20191125.3](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=12677). Now App Service Explorer auto refreshes when the Cosmos DB Extension has been installed, so this bug has been fixed.

PR #1279, opened 2019-11-07
Poll 🗳️ for Cosmos 👨‍🚀⭐️ installation 

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1161

This more importantly adds a `refreshImpl` to `CosmosDBTreeItem` which was absent before.  Previously when you refreshed the tree item, it would not update the installation status of Cosmos, leading to confusion as to whether or not the extension was actually installed.

### DIFF
--- a/src/commands/installCosmosDBExtension.ts
+++ b/src/commands/installCosmosDBExtension.ts
@@ -7,19 +7,10 @@ import { commands } from 'vscode';
 import { AzureTreeItem, IActionContext } from 'vscode-azureextensionui';
 import { CosmosDBTreeItem } from '../explorer/CosmosDBTreeItem';
 import { delay } from '../utils/delay';
-import { openUrl } from "../utils/openUrl";
 
 export async function installCosmosDBExtension(context: IActionContext, treeItem: AzureTreeItem): Promise<void> {
-    const commandToRun = 'extension.open';
-    const listOfCommands = await commands.getCommands();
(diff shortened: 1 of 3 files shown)


src/commands/installCosmosDBExtension.ts
```diff
@@ -0,0 +1,39 @@
...
+import { commands } from 'vscode';
+import { AzureTreeItem, IActionContext } from 'vscode-azureextensionui';
+import { CosmosDBTreeItem } from '../explorer/CosmosDBTreeItem';
+import { delay } from '../utils/delay';
+import { openUrl } from "../utils/openUrl";
+
+export async function installCosmosDBExtension(context: IActionContext, treeItem: AzureTreeItem): Promise<void> {
+    const commandToRun = 'extension.open';
+    const listOfCommands = await commands.getCommands();
+    if (listOfCommands.find((x: string) => x === commandToRun)) {
```
EricJizbaMSFT: You can just use the command without doing all this extra validation. Our minimum engine is 1.31.0 in package.json and the command was added in September 2018 (https://github.com/microsoft/vscode/issues/58642) which looks like it was 1.28.0

nturinski:
Do you think it's worth finding all CosmosDBTreeItem nodes and refreshing them?  If the user had multiple connection nodes opened, only the one that they clicked would get refreshed.

nturinski:
🐫

Issue #1298, opened 2019-11-21
Node v12.13.1 is on local PC, Node 12 LTS is running on Azure


AlDjo:
Repro steps:

1.
2.

Action: appService.StartRemoteDebug
Error type: Error
Error Message: Azure Remote Debugging is currently only supported for Node.js version >= 8.11 on Linux.

Version: 0.16.1
OS: win32
Product: Visual Studio Code
Product Version: 1.40.1
Language: en

<details>
<summary>Call Stack</summary>

```
Object.t.checkForRemoteDebugSupport extension.bundle.js:157:366496
extension.bundle.js:323:880961extension.bundle.js:323:880961
o extension.bundle.js:323:880314
processTicksAndRejections task_queues.js:89:5
```

</details>

nturinski:
This should be resolved with the latest release, v0.16.2.

Please let me know if it hasn't! 😃

PR #1305, opened 2019-11-26
Use a regex for find LTS in version rather than equals operator

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1298

Back when we implemented this, there was only Node LTS.  Since then, the runtime team has added Node 12 LTS, Node 8 LTS, etc.

### DIFF
--- a/src/commands/remoteDebug/getRemoteDebugLanguage.ts
+++ b/src/commands/remoteDebug/getRemoteDebugLanguage.ts
@@ -54,7 +54,7 @@ function isNodeVersionSupported(nodeVersion: string): boolean {
     if (/lts/i.test(nodeVersion)) {
         const ltsVersionNumber: number = Number(nodeVersion.split('-')[0]);
         // if there is no version number, ltsVersionNumber will equal 'lts' which isNan
-        if (ltsVersionNumber < 8 && !isNaN(ltsVersionNumber)) {
+        if (!isNaN(ltsVersionNumber) && ltsVersionNumber < 8) {
             return false;
         }
 
--- a/src/commands/remoteDebug/getRemoteDebugLanguage.ts
+++ b/src/commands/remoteDebug/getRemoteDebugLanguage.ts
@@ -50,7 +50,7 @@ export function getRemoteDebugLanguage(siteConfig: SiteConfigResource, context: 
 
 // Remote debugging is currently only supported for Node.js >= 8.11
 function isNodeVersionSupported(nodeVersion: string): boolean {
-    // the portal's runtime includes several LTS versions i.e. (12-lts)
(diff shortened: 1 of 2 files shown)


src/commands/remoteDebug/getRemoteDebugLanguage.ts
```diff
@@ -50,8 +50,8 @@ export function getRemoteDebugLanguage(siteConfig: SiteConfigResource, context:
 
 // Remote debugging is currently only supported for Node.js >= 8.11
 function isNodeVersionSupported(nodeVersion: string): boolean {
-    // the portal's new default node runtime is LTS
-    if (nodeVersion.toLocaleLowerCase() === 'lts') {
+    // the portal's Node runtimes includes several LTS versions i.e. (12-lts)
+    if (/lts/.test(nodeVersion)) {
```
EricJizbaMSFT: Add the `i` flag for case insensitivity

src/commands/remoteDebug/getRemoteDebugLanguage.ts
```diff
@@ -50,8 +50,8 @@ export function getRemoteDebugLanguage(siteConfig: SiteConfigResource, context:
 
 // Remote debugging is currently only supported for Node.js >= 8.11
```
EricJizbaMSFT: I assume the portal doesn't support 6-lts? And that 8-lts actually resolves to a version greater than 8.11?

src/commands/remoteDebug/getRemoteDebugLanguage.ts
```diff
@@ -50,8 +50,14 @@ export function getRemoteDebugLanguage(siteConfig: SiteConfigResource, context:
...
 function isNodeVersionSupported(nodeVersion: string): boolean {
-    // the portal's new default node runtime is LTS
-    if (nodeVersion.toLocaleLowerCase() === 'lts') {
+    // the portal's Node runtimes includes several LTS versions i.e. (lts, 8-lts, 12-lts)
+    if (/lts/i.test(nodeVersion)) {
+        const ltsVersionNumber: number = Number(nodeVersion.split('-')[0]);
+        // if there is no version number, ltsVersionNumber will equal 'lts' which isNan
+        if (ltsVersionNumber < 8 && !isNaN(ltsVersionNumber)) {
```
EricJizbaMSFT: Shouldn't this be `||`?

nturinski: Nah, it needs to be a number and less than 8.  If it isn't a number, then it's LTS, and that means we want it to skip this condition and return true.

PR #1358, opened 2020-01-15
Add test case to deploy Node.js LTS project

v-wuzhai:


test/nightly/deploy.test.ts
```diff
@@ -0,0 +1,67 @@
...
+import { WebSiteManagementModels } from 'azure-arm-website';
+import { IHookCallbackContext, ISuiteCallbackContext } from 'mocha';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { DialogResponses, getRandomHexString, requestUtils } from '../../extension.bundle';
+import { longRunningTestsEnabled, testUserInput } from '../global.test';
+import { resourceGroupsToDelete, webSiteClient } from './global.resource.test';
+
+suite('Create Web App and deploy', async function (this: ISuiteCallbackContext): Promise<void> {
+    this.timeout(20 * 6 * 1000);
```
nturinski: Let's increase the timeout for this particular test.  I've run it twice, and once Linux timed out and the other time, Mac OS timed out with:

`Error: Timeout of 120000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.`

I'm confident that the test is fine, so it probably just needs like 200000 ms.

test/nightly/deploy.test.ts
```diff
@@ -0,0 +1,67 @@
...
+});
+
+// The workspace folder that vscode is opened against for tests
+async function getWorkspacePath(testWorkspaceName: string): Promise<string> {
+    let workspacePath: string = '';
+    const workspaceFolders: vscode.WorkspaceFolder[] | undefined = vscode.workspace.workspaceFolders;
+    if (!workspaceFolders || workspaceFolders.length === 0) {
+        throw new Error("No workspace is open");
+    } else {
+        assert.equal(workspaceFolders.length, 1, "Expected three workspace to be open.");
```
nturinski: I think there is a mistake in this error message.

test/nightly/deploy.test.ts
```diff
@@ -0,0 +1,67 @@
...
+            await vscode.commands.executeCommand('appService.CreateWebAppAdvanced');
+        });
+        const createdApp: WebSiteManagementModels.Site = await webSiteClient.webApps.get(resourceGroupName, resourceName);
+        assert.ok(createdApp);
+
+        // Verify that the deployment is successful
+        await testUserInput.runWithInputs([workspacePath, resourceName, 'Deploy', DialogResponses.yes.title], async () => {
+            await vscode.commands.executeCommand('appService.Deploy');
+        });
+        const request: requestUtils.Request = await requestUtils.getDefaultRequest(`https://${resourceName}.azurewebsites.net`);
```
nturinski: Rather than hardcoding the url, could you retrieve the node and use `this.root.client.defaultHostUrl` to get the website?  This will help if we start testing with other endpoints like Mooncake.

Alternatively, we can update createWebApp and createWebAppAdvanced to return the `WebAppTreeItem` after it is called or return the WebAppTreeItem's ID and use that to retrieve the node like Functions https://github.com/microsoft/vscode-azurefunctions/blob/f641c1166f74ad11be2fc67b3e689d26ae8e40b0/src/commands/createFunctionApp/createFunctionApp.ts#L12-L29

test/nightly/deploy.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+// The workspace folder that vscode is opened against for tests
+async function getWorkspacePath(testWorkspaceName: string): Promise<string> {
+    let workspacePath: string = '';
+    const workspaceFolders: vscode.WorkspaceFolder[] | undefined = vscode.workspace.workspaceFolders;
+    if (!workspaceFolders || workspaceFolders.length === 0) {
+        throw new Error("No workspace is open");
+    } else {
+        assert.equal(workspaceFolders.length, 1, "Expected one workspace to be open.");
+        for (const obj of workspaceFolders) {
```
nturinski: You should be able to just check if workspaceFolders[0].name === testWorkspaceName since we know that the length of the array is 1.

Issue #1357, opened 2020-01-15
Update defaults when creating new apps


fiveisprime:
- [ ] Enable verbose/application logging (related to #1287)
- [x] Enable HTTP logging
- [ ] Set `SCM_DO_BUILDING_DURING_DEPLOYMENT` to `true` (need this when creating Windows too)
- [ ] Set `logsDirectorySizeLimit` to `100`

I still need to verify that `WEBSITE_HTTPLOGGING_RETENTION_DAYS` is being honored in Linux so that may need to be set as well.

nturinski:
Is `WEBSITE_HTTPLOGGING_RETENTION_DAYS` not honored in Windows?  We've been setting it to 7 whenever we enabled log streaming in the past.

nturinski:
@fiveisprime

Can you confirm that all logging isn't enabled by this one setting?
![image](https://user-images.githubusercontent.com/5290572/73314880-75535a80-41e3-11ea-8f27-3eeac09c9a28.png)

https://docs.microsoft.com/en-us/azure/app-service/troubleshoot-diagnostic-logs

This documentation seems to imply that to get all this logging, you just enable Application Logging, but I haven't been able to confirm that.

This PR https://github.com/microsoft/vscode-azureappservice/pull/1370 only enables httpLogs in the code, but when I go to the portal, Application Logging seems to be enabled.

Also, we might not want to default `retentionInDays` to 7.  According to this tooltip, by default, logs are never deleted:

![image](https://user-images.githubusercontent.com/5290572/73315090-0aeeea00-41e4-11ea-81e7-cd4066184470.png)

Maybe we should pass in undefined for that property?

fiveisprime:
I'm not sure about the UI element, but you should be able to see this in the web app object (these should be defaults that the extension sets): 

```js
{
  "httpLoggingEnabled": true
  "logsDirectorySizeLimit": 100
  "virtualApplications": ["preloadEnabled": true]
}
```

fiveisprime:
I added a note that the SCM_DO_BUILD... app setting should also be set for Windows apps. This comes from feedback from the docs team where the Windows workflow (via local git) run npm install by default.

fiveisprime:
Discussed offline: check and re-enable file logging on deploy.

PR #1370, opened 2020-01-27
Enable Http logs by default

nturinski:
Related to https://github.com/microsoft/vscode-azureappservice/issues/1357

The new detectors that I want to consume require HTTP logging to be enabled too.

### DIFF
--- a/src/commands/createWebApp/createWebApp.ts
+++ b/src/commands/createWebApp/createWebApp.ts
@@ -5,7 +5,6 @@
 
 import { AzureParentTreeItem, IActionContext, ICreateChildImplContext } from "vscode-azureextensionui";
 import { SubscriptionTreeItem } from '../../explorer/SubscriptionTreeItem';
-import { WebAppTreeItem } from "../../explorer/WebAppTreeItem";
 import { ext } from "../../extensionVariables";
 
 export async function createWebApp(context: IActionContext & Partial<ICreateChildImplContext>, node?: AzureParentTreeItem | undefined): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/commands/createWebApp/createWebApp.ts
```diff
@@ -14,6 +14,7 @@ export async function createWebApp(context: IActionContext & Partial<ICreateChil
     }
 
     const newSite: WebAppTreeItem = <WebAppTreeItem>await node.createChild(context);
+    await newSite.enableHttpLogs();
```
EricJizbaMSFT: This won't happen if people create from the blue up arrow, which is the more common case

nturinski:
He lists that as verbose/application logging in #1357

src/commands/createWebApp/createWebApp.ts
```diff
@@ -5,16 +5,14 @@
...
-import { WebAppTreeItem } from "../../explorer/WebAppTreeItem";
 import { ext } from "../../extensionVariables";
 
 export async function createWebApp(context: IActionContext & Partial<ICreateChildImplContext>, node?: AzureParentTreeItem | undefined): Promise<void> {
     if (!node) {
         node = <AzureParentTreeItem>await ext.tree.showTreeItemPicker(SubscriptionTreeItem.contextValue, context);
     }
 
-    const newSite: WebAppTreeItem = <WebAppTreeItem>await node.createChild(context);
-    newSite.promptToDeploy(context);
```
EricJizbaMSFT: I don't think you meant to remove this line?

nturinski: Yeah, you're right =_= I'm being stupid, I was thinking that we didn't use `newSite` for anything for some reason

PR #1391, opened 2020-02-13
Fix tests on linux build machine

EricJizbaMSFT:
As of today linux tests are silently failing in Azure Pipelines. All I did was port over the fix implemented by VS Code folks and it seems to work: https://github.com/microsoft/vscode/pull/89323

cc @haniamr


haniamr:
> As of today linux tests are silently failing in Azure Pipelines. All I did was port over the fix implemented by VS Code folks and it seems to work: [microsoft/vscode#89323](https://github.com/microsoft/vscode/pull/89323)
> 
> cc @haniamr

Thanks @EricJizbaMSFT, I'll port the fix for Docker as well.

Issue #1249, opened 2019-10-22
Switch over to application log detectors


nturinski:
Apparently, this detector will be almost instant as it is getting the application log rather than kusto which results in much better performance.  It still provides insights as to why the deployment has failed, so it might serve our purposes much better than Linux Web App Down Detector.

nturinski:
Fixed with #1444

PR #1444, opened 2020-03-17
New LinuxLogDetector

nturinski:
This leverages the new LinuxLogDetector which needs to have FileLogging enabled to work.  I believe right now, if it is not enabled, the logging will just report no critical errors.

The detector will see if the app started, and if not, offer insights.  If there are no app insights, it's possible that the Docker container itself.

I'll fix the tests in this PR, but it's a little hefty so I wanted to get the ball rolling (and prove that I'm actually working remotely 😅)


src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -58,18 +57,32 @@ export async function checkLinuxWebAppDownDetector(correlationId: string, node:
                 return undefined;
             }
 
-            // tslint:disable-next-line: no-unsafe-any
-            detectorErrorMessage = await getLinuxDetectorError(context, linuxContainerStartFailureId, node, deployResultTime, node.root.client.fullName);
+            // time constraint of being less than the current time by 14.5 minutes is enforced system wide for all detectors
+            // however LinuxLogViewer detector ignores it and gets the most recent data
+            const nowTime: Date = new Date();
+            const startTimeDate: Date = new Date(nowTime.getTime() - (60 * 60 * 1000));
```
EricJizbaMSFT: Instead of `nowTime.getTime()`, would `Date.now()` give you the same thing?

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -58,18 +57,32 @@ export async function checkLinuxWebAppDownDetector(correlationId: string, node:
...
+            const endTimeDate: Date = new Date(nowTime.getTime() - (30 * 60 * 1000));
```
EricJizbaMSFT: I don't get this - is it saying the end time is 30 minutes ago? Before I deployed? Is this related to the above comment? If so - please explain how the 30/60 is related to 14.5

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -58,18 +57,32 @@ export async function checkLinuxWebAppDownDetector(correlationId: string, node:
...
+
+            // need to remove the miliseconds from the ISOstring for request
+            const startTimeArray: string[] = startTimeDate.toISOString().split(':');
+            const startTime: string = `${startTimeArray[0]}:${startTimeArray[1]}`;
+
+            const endTimeArray: string[] = endTimeDate.toISOString().split(':');
+            const endTime: string = `${endTimeArray[0]}:${endTimeArray[1]}`;
```
EricJizbaMSFT: How about you use `moment` to format exactly the date you want instead of doing this splitting stuff

Same thing as here: https://github.com/microsoft/vscode-azuretools/pull/658/files#diff-e1557ab1c14775da7a5262caf71fee73R22

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -58,18 +57,32 @@ export async function checkLinuxWebAppDownDetector(correlationId: string, node:
...
+                await delay(1000 * 10);
             }
         }
 
         if (showOutput) {
-            await ext.ui.showWarningMessage(detectorErrorMessage, { title: 'Open in Portal' });
-            await openInPortal(node.root, `${node.root.client.id}/troubleshoot`, { queryPrefix: `websitesextension_ext=asd.featurePath%3Ddetectors%2F${linuxContainerStartFailureId}` });
+            ext.outputChannel.appendLog(detectorErrorMessage);
+            const openInPortalMsg: string = localize('openInPortal', 'Open in Portal for more details.');
+            await ext.ui.showWarningMessage(`${detectorErrorMessage} ${openInPortalMsg}`, { title: 'Open in Portal' });
```
EricJizbaMSFT: How about we get rid of "Open in Portal for more details" from the message and just make the button "View details"

The duplicate "Open in Portal" bothers me and I don't think we have to tell them beforehand that it'll be in the portal

src/commands/postDeploy/getLinuxDetectorError.ts
```diff
@@ -1,83 +1,95 @@
+
```
EricJizbaMSFT: 🗑

src/commands/postDeploy/getLinuxDetectorError.ts
```diff
@@ -1,83 +1,95 @@
...
     solutions = "Solutions",
     time = "Time",
     instance = "Instance",
     facility = "Facility",
-    failureCount = "FailureCount"
+    failureCount = "FailureCount",
+    dataValue = 'Data.Value',
+    dataName = 'Data.Name'
 }
+const timestampFormat: RegExp = /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}/;
```
EricJizbaMSFT: prefer to have this variable closer to where it's used

src/commands/postDeploy/getLinuxDetectorError.ts
```diff
@@ -1,83 +1,95 @@
...
+    const criticalError: string = localize('insightError', '"{0}" reported a critical error:', node.root.client.siteName);
 
-        context.telemetry.properties.errorMessages = JSON.stringify(errorMessages);
-        context.telemetry.properties.failureCount = failureCount;
+    if (getValuesByColumnName(context, insightTable, ColumnName.status) === 'Critical') {
+        const insightError: string = getValuesByColumnName(context, insightTable, ColumnName.dataValue);
+        context.telemetry.properties.errorMessages = JSON.stringify(insightError);
 
-        return `Critical insights found for "${siteName}": ${failureCount} as of ${new Date(timestamp).toLocaleString()}. Critical insight preview: "${errorMessages}". Click here to access "App Service Diagnostics" for recommendations.`;
+        return `${criticalError} ${insightError}".`;
```
EricJizbaMSFT: `insightError` should be added to the string inside of `localize` just like you do for `node.root.client.siteName` above.

There could be cases where languages read right-to-left instead of left-to-right and thus `insightError` would go first.

src/commands/postDeploy/getLinuxDetectorError.ts
```diff
@@ -1,83 +1,95 @@
...
+    const insightDataset: detectorDataset[] = <detectorDataset[]>JSON.parse(rawApplicationLog);
+
+    let insightTable: detectorTable;
+    let detectorTimestamp: RegExpMatchArray | null;
+
+    const appInsightTable: detectorTable | undefined = findTableByName(insightDataset, 'application/insight');
+    if (appInsightTable) {
+        insightTable = appInsightTable;
+        context.telemetry.properties.insight = 'app';
+        detectorTimestamp = getValuesByColumnName(context, appInsightTable, ColumnName.dataName).match(timestampFormat);
```
EricJizbaMSFT: Can you give me example data of what's going on here? I'm a little confused what the regexp is doing and why it's necessary

src/commands/postDeploy/parseDetectorResponse.ts
```diff
@@ -48,6 +31,7 @@ export function getValuesByColumnName(context: IActionContext, table: detectorTa
     context.telemetry.properties.columnName = columnName;
     context.telemetry.properties.numberOfValues = values.length.toString();
 
+    // the last one should be the most recent
     // tslint:disable-next-line: strict-boolean-expressions
-    return values[0] || '';
+    return values[values.length - 1] || '';
```
EricJizbaMSFT: How about `values.pop()` instead of `values[values.length - 1]`

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -9,11 +9,12 @@ import { KuduClient } from "vscode-azurekudu";
 import { DeployResult } from "vscode-azurekudu/lib/models";
```
EricJizbaMSFT: This isn't directly related to this PR, but in your demo the other day I saw this message and it bothered me:
> "Diagnosing web app "${node.root.client.siteName}" for critical errors..."

It makes it seem like I already have critical errors even on a successful deploy. It also makes it feel like I have to wait for the diagnostics to finish before I can truly know if my site deployed, but we never "finish" diagnostics, we just time out. Can we just get rid of the message?

Another reason is because you might not actually be diagnosing if FileLogging isn't enabled.

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -39,8 +39,7 @@ export async function checkLinuxWebAppDownDetector(correlationId: string, node:
         }
 
         let detectorErrorMessage: string | undefined;
-        // for telemetry's sake, wait for 30 minutes from after "deployment complete" prompt
-        const detectorTimeoutMs: number = Date.now() + 30 * 60 * 1000;
+        const detectorTimeoutMs: number = Date.now() + 5 * 60 * 1000;
```
EricJizbaMSFT: 🎉🙏🎊🍻

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1249

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
nturinski: I picked 60/30 because it's way beyond the 14.5 minute restraint that the detectors enforce.  I don't understand why they do it _(probably because Kusto query injection takes forever so you can't retrieve any detector information that isn't at least 14.5 minutes old)._

I could do start/end as 15/14.5, I guess but I didn't wanna mess with configuring the times since I asked Michimune and he said the LinuxLogViewer detector completely ignores that and just returns the most recent entry regardless.  My guess is because it's using app logs that populate while the deployment is happening, the time constraint doesn't apply to them but is a platform-wide constraint that they can't just ignore.

src/commands/postDeploy/getLinuxDetectorError.ts
nturinski: image

I'm looking for the timestamp from the insight and comparing it to the deployEndTime to make sure that we're not looking at stale data.

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
nturinski: Yeah, I totally get that.  And the message saying that it "timed out" also makes it sound like that there are critical errors, but we just couldn't find them.  I wanted some way to alert the user that more information may be incoming to keep their attention, but I think I agree with you.

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -50,26 +41,33 @@ export async function checkLinuxWebAppDownDetector(correlationId: string, node:
...
-                // poll every minute
-                await delay(1000 * 60);
+                // poll every 10 seconds
+                await delay(1000 * 10);
             }
         }
 
+        const enableDetectorsSetting: string = 'enableDetectors';
+        const showOutput: boolean | undefined = getWorkspaceSetting<boolean>(enableDetectorsSetting);
         if (showOutput) {
```
nturinski, line 67: I was thinking it could be time to remove this flag.

EricJizbaMSFT: Personally I'm still waiting to see examples of common, real-life errors that this finds and displays to the user in an easy-to-understand, actionable way

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -50,26 +41,34 @@ export async function checkLinuxWebAppDownDetector(correlationId: string, node:
...
-            await ext.ui.showWarningMessage(detectorErrorMessage, { title: 'Open in Portal' });
-            await openInPortal(node.root, `${node.root.client.id}/troubleshoot`, { queryPrefix: `websitesextension_ext=asd.featurePath%3Ddetectors%2F${linuxContainerStartFailureId}` });
+            ext.outputChannel.appendLog(detectorErrorMessage);
+            await ext.ui.showWarningMessage(detectorErrorMessage, { title: 'View details' });
```
EricJizbaMSFT, line 70: You'll want to do something like this where it doesn't await the message and creates a new event for the result of the message:
https://github.com/microsoft/vscode-azureappservice/blob/d6fbafdb8c8aac22c3310ec25f932fe62894db99/src/commands/deploy/showDeployCompletedMessage.ts#L22

Otherwise we won't get any telemetry for users when an error is found, but they never interact with the notification

src/commands/postDeploy/getLinuxDetectorError.ts
EricJizbaMSFT: Why does it have "[1] " in front of the time? I'm just worried about the format of this changing in the future because it seems really weird. Ideally the RegExp we use is super simple and mostly just passes along the time and lets `new Date` parse it in whatever format it comes in

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
```diff
@@ -50,26 +41,34 @@ export async function checkLinuxWebAppDownDetector(correlationId: string, node:
...
+                context.telemetry.properties.timedOut = 'true';
                 return undefined;
             }
 
-            // tslint:disable-next-line: no-unsafe-any
-            detectorErrorMessage = await getLinuxDetectorError(context, linuxContainerStartFailureId, node, deployResultTime, node.root.client.fullName);
+            // time constraint of being less than the current time by 14.5 minutes is enforced system wide for all detectors
+            // however LinuxLogViewer detector ignores it and gets the most recent data
+            const nowTime: number = Date.now();
+            const startTimeDate: Date = new Date(nowTime - (60 * 60 * 1000));
```
EricJizbaMSFT, line 51: Okay I think I understand this now - is this right?

> There's a requirement (outside the control of the LinuxLogDetector team) that these times have to be less than the current time by 14.5 minutes. That being said, the LinuxLogDetector team ignores the times and always gets the most recent entry. We'll use 30/60 minutes just to make sure we're well over 14.5

If so, can we make that the comment in the code? The most important piece is I want the comment to explicitly reference 30/60 otherwise I feel like it's really confusing

src/commands/postDeploy/getLinuxDetectorError.ts
nturinski: > I added the [1] because the insight table is internally Dictionary<string,object> and the key string needs to be unique. So if there are multiple issues happened at the same timestamp, then it will throw an exception.

EricJizbaMSFT: That makes very little sense to me. What's the key and what's the value? Why would a timestamp be a key?

That aside, instead of a regexp for the time part, seems like you could have a simpler regexp to cut out the `[1] ` part

src/commands/postDeploy/checkLinuxWebAppDownDetector.ts
EricJizbaMSFT: Okay demos in standup have satisfied me - 🚫🚩

nturinski: Emoji game ✊

src/commands/postDeploy/getLinuxDetectorError.ts
```diff
@@ -3,83 +3,97 @@
...
-        // if the criticalError table or failureCounts aren't defined, we can't display anything to the user
-        if (!criticalErrorTable || !failureCountTable) {
+    const rawApplicationLog: string = getValuesByColumnName(context, insightLogTable, ColumnName.value);
+    const insightDataset: detectorDataset[] = <detectorDataset[]>JSON.parse(rawApplicationLog);
+
+    let insightTable: detectorTable;
+    let detectorTimestamp: string;
+
+    // [1] 2020-04-21T18:23:50 is the format of the timestamp in the insight response, this RegExp will remove the brackets
+    const bracketsAndSpace: RegExp = /\[.*?\]\s/;
```
EricJizbaMSFT, line 56: Please modify the above comment to explain why there is a bracket thing in the timestamp. Basically what you explained to me in the GitHub comments

src/commands/postDeploy/getLinuxDetectorError.ts
```diff
@@ -3,83 +3,97 @@
...
-        const errorMessages: string | undefined = getValuesByColumnName(context, criticalErrorTable, ColumnName.message);
-        const failureCount: string | undefined = getValuesByColumnName(context, failureCountTable, ColumnName.failureCount);
+        insightTable = dockerInsightTable;
+        context.telemetry.properties.insight = 'docker';
+        detectorTimestamp = getValuesByColumnName(context, dockerInsightTable, ColumnName.dataName).replace(bracketsAndSpace, '');
+    }
 
-        context.telemetry.properties.errorMessages = JSON.stringify(errorMessages);
-        context.telemetry.properties.failureCount = failureCount;
+    detectorTimestamp = moment.utc(detectorTimestamp).format(timeFormat);
```
EricJizbaMSFT, line 74: Lol `timeFormat` makes me laugh. It's such a generic name and you put it in the `constants` file where the whole extension could use it. Makes me think "IF YOU EVER USE A TIME IT HAS TO BE IN THIS FORMAT OR ELSE"

Suggest you at least rename it to "detectorTimeFormat"

EricJizbaMSFT, line 74: Rather than doing `.replace(bracketsAndSpace, '');` twice above, just do it here.

src/commands/postDeploy/getLinuxDetectorError.ts
```diff
@@ -3,83 +3,97 @@
...
-
-    // we don't know how long it takes for the deployResult startTime and the detector timestamp
+export function validateTimestamp(context: IActionContext, detectorTime: string, deployResultTime: string): boolean {
+    const secondsBetweenTimes: number = (new Date(detectorTime).getTime() - new Date(deployResultTime).getTime()) / 1000;
+    // the detector can be as fast as ~20 seconds for app errors, but Docker container errors seem to timeout at
+    // about 5 minutes
     context.telemetry.properties.timeBetweenDeployAndDetector = secondsBetweenTimes.toString();
-    if (secondsBetweenTimes <= 60) {
+
+    // detector time must be more recent than deployResultTime
```
EricJizbaMSFT, line 95: So this is only an issue if they start another deploy while we're still checking the last deploy for detector errors. But this gets canceled if they start another deploy, right? I mean...at least if they're deploying from the same VS Code window

nturinski: Yeah, correct. We cancel it if we detect that the post-deployment task is going on.  But yeah, someone could screw it up if they deploy it from another window, but technically, they'd probably still want the most recent error anyway.

main.js
```diff
@@ -16,10 +16,10 @@ let perfStats = {
 
 Object.defineProperty(exports, "__esModule", { value: true });
 
-const extension = require('./out/src/extension');
```
EricJizbaMSFT, line 19: Revert changes to this file. Related to https://github.com/microsoft/vscode-azurefunctions/pull/1973#issue-387997953

PR #1457, opened 2020-03-25
Engineering work to make our repos more consistent 👷🏻‍♂️

EricJizbaMSFT:
See https://github.com/microsoft/vscode-cosmosdb/pull/1387 for more info


test/tslint.json
```diff
@@ -4,7 +4,25 @@
...
         "no-implicit-dependencies": [
             true,
-            "dev"
-        ]
+            "dev",
+            [
+                "vscode"
+            ]
+        ],
+        "typedef": [
```
nturinski: Are we enforcing tslint in our test files too?

EricJizbaMSFT: We always have and this file actually _disables_ rules that I feel are unnecessary to run in test files. At the top of the file you can see it `"extends": "../tslint.json",` and all of the stuff in here is essentially set to false

nturinski:
How come you're not merging this?

EricJizbaMSFT:
I expanded this PR to include more stuff and wanted to get a fresh approval before merging. I submitted almost exactly the same PR in 3 other repos (functions, cosmos, storage), so I figured I'd count an approval there as a "fresh approval" rather than pinging for another review in this PR

PR #1464, opened 2020-03-30
Add test case to deploy .NET Core LTS project

v-wuzhai:


test/nightly/deploy.test.ts
```diff
@@ -41,6 +41,11 @@ suite('Create Web App and deploy', async function (this: ISuiteCallbackContext):
         await testCreateWebAppAndDeploy(['Linux', 'Node 12 LTS'], testFolderPath);
     });
 
+    test('.NET Core LTS', async () => {
```
nturinski: Could you make a new file to deploy.dotnet and deploy.nodejs rather than having them all in one file?  This will get big very quickly 😅

test/nightly/deploy.test.ts
```diff
@@ -67,11 +72,16 @@ suite('Create Web App and deploy', async function (this: ISuiteCallbackContext):
 
 // The workspace folder that vscode is opened against for tests
 async function getWorkspacePath(testWorkspaceName: string): Promise<string> {
+    let workspacePath: string = '';
     const workspaceFolders: vscode.WorkspaceFolder[] | undefined = vscode.workspace.workspaceFolders;
     if (!workspaceFolders || workspaceFolders.length === 0) {
         throw new Error("No workspace is open");
     } else {
-        const workspacePath: string = workspaceFolders[0].uri.fsPath;
+        for (const obj of workspaceFolders) {
```
nturinski: Technically having multi-root workspaces is a different test case, though this is a limitation I'm not sure what to do about.  @EricJizbaMSFT any ideas?

test/nightly/deploy.test.ts
EricJizbaMSFT: As an alternative, we could keep everything in this file and prevent it from getting too big with a strategy like [this](https://github.com/microsoft/vscode-azurefunctions/blob/master/test/nugetUtils.test.ts#L10). I expect `ITestCase` would look something like this:
```
interface ITestCase {
    workspaceFolder: string;
    runtimes: string[];
}
```

And the `for` loop would look something like this:
```
for (const testCase of testCases) {
    for (const runtime of testCase.runtimes) {
        test(runtime, async () => {
            const testFolderPath: string = await getWorkspacePath(testCase.workspaceFolder);
            await testCreateWebAppAndDeploy(['Linux', runtime], testFolderPath);
        });
    }
}
```

test/nightly/deploy.test.ts
EricJizbaMSFT: I see what you mean, but I think this is the easiest way to get the test done so I'm fine with it. This actually isn't unique to this PR. It was already testing a multi-root workspace since it was using a `.code-workspace` file (even though that workspace file only happened to have one folder)

test/test.code-workspace
```diff
@@ -1,5 +1,8 @@
 {
     "folders": [
+        {
+            "path": "./nightly/testFolder/dotnet-hello-world"
```
EricJizbaMSFT, line 4: I would suggest putting this project in a "2.1" folder since we'll likely want to test a "3.1" project in the future

"./nightly/testFolder/dotnet-hello-world/2.1"

PR #1477, opened 2020-04-09
Exclude test apps from being displayed in VS Code

EricJizbaMSFT:
These errors were showing up for me and I want to hide them:
Screen Shot 2020-04-09 at 11 36 22 AM

cc @v-wuzhai

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,9 @@
     "editor.formatOnSave": true,
     "editor.insertSpaces": true,
     "editor.tabSize": 4,
+    "files.exclude": {
+        "test/nightly/testFolder/*hello-world": true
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -7,6 +7,9 @@
     "editor.formatOnSave": true,
     "editor.insertSpaces": true,
     "editor.tabSize": 4,
+    "files.exclude": {
```
nturinski, line 10: Should you just ignore the entire testFolder instead of any directory that includes "hello-world"?

EricJizbaMSFT: I want the README to display otherwise I feel like people might be confused why the folder shows as empty

PR #1517, opened 2020-04-29
Add new setting test case

v-wuzhai:


test/nightly/appSettings.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import * as assert from 'assert';
+import { WebSiteManagementModels as Models } from 'azure-arm-website';
+import { IHookCallbackContext, ISuiteCallbackContext } from 'mocha';
+import * as vscode from 'vscode';
+import { getRandomHexString } from '../../extension.bundle';
+import { longRunningTestsEnabled, testUserInput } from '../global.test';
+import { resourceGroupsToDelete, webSiteClient } from './global.resource.test';
+
+suite('Application settings', async function (this: ISuiteCallbackContext): Promise<void> {
```
nturinski, line 14: I think test is worth putting in the webAppActions file.  Creating and deleting a resource kind of takes a while and there's really no reason to create another app just to test app settings.

In fact, I think it'd be cool if we could find a way to make it alternate between testing it on a Linux/Windows app service.  Maybe it could be something like today's date % 2 to determine whether or not to test it for Windows or Linux?  I don't know, there's probably a better way to do that 😅

test/nightly/azureResourceWebAppActions.test.ts
```diff
@@ -10,21 +10,25 @@ import { constants, DialogResponses, getRandomHexString } from '../../extension.
 import { longRunningTestsEnabled, testUserInput } from '../global.test';
 import { resourceGroupsToDelete, webSiteClient } from './global.resource.test';
 
-// tslint:disable-next-line: max-func-body-length
 suite('Web App actions', async function (this: Mocha.Suite): Promise<void> {
     this.timeout(6 * 60 * 1000);
     let resourceName: string;
+    let appSettingKey: string;
+    let appSettingValue: string;
+    const os0: string = (new Date().getDate()) % 2 === 0 ? 'Linux' : 'Windows';
```
nturinski, line 18: Could you use the `WebsiteOS enum` for this?

test/nightly/azureResourceWebAppActions.test.ts
```diff
@@ -100,6 +104,15 @@ suite('Web App actions', async function (this: Mocha.Suite): Promise<void> {
         assert.equal(createdApp.scmType, constants.ScmType.None, `Web App scmType's property value should be ${constants.ScmType.None} rather than ${createdApp.scmType}.`);
     });
 
+    test('Add new setting', async () => {
```
nturinski, line 107: Could you include which OS is being used to run this action test?  Along with the other tests above?  It'll be helpful if the test fails to know which OS it failed on 😄

test/nightly/azureResourceWebAppActions.test.ts
```diff
@@ -15,8 +15,8 @@ suite('Web App actions', async function (this: Mocha.Suite): Promise<void> {
     let resourceName: string;
     let appSettingKey: string;
     let appSettingValue: string;
-    const os0: string = (new Date().getDate()) % 2 === 0 ? 'Linux' : 'Windows';
-    const os1: string = os0 === 'Windows' ? 'Linux' : 'Windows';
+    const WebsiteOS0: string = (new Date().getDate()) % 2 === 0 ? 'Linux' : 'Windows';
```
nturinski, line 18: Sorry, I should been more clear.  I meant change the type of string to `WebsiteOS`.  Like this: https://github.com/microsoft/vscode-azureappservice/blob/5398a7dd58a5726f9fa1625b903e3fba5176f596/src/commands/createWebApp/setPrePromptDefaults.ts#L55

PR #1526, opened 2020-05-06
Implement getResourceName to enable more verbose logs

alexweininger:
Related to PR: [vscode-azuretools#701](https://github.com/microsoft/vscode-azuretools/pull/701)

- Implement getResourceName so the resource name can be included in logs.
- Bumped the version of the shared UI package to 0.32.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8417,6 +8417,26 @@
                 "vscode-azurekudu": "^0.1.9",
                 "vscode-nls": "^4.1.1",
                 "websocket": "^1.0.31"
+            },
+            "dependencies": {
+                "vscode-azureextensionui": {
+                    "version": "0.31.2",
+                    "resolved": "https://registry.npmjs.org/vscode-azureextensionui/-/vscode-azureextensionui-0.31.2.tgz",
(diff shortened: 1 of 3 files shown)


nturinski:
I would say congrats on the first PR, but I guess you've been submitting PRs from even before your internship 😁 

🎉 🎉 🎉

package-lock.json
```diff
@@ -8417,6 +8417,26 @@
                 "vscode-azurekudu": "^0.1.9",
                 "vscode-nls": "^4.1.1",
                 "websocket": "^1.0.31"
+            },
+            "dependencies": {
+                "vscode-azureextensionui": {
```
EricJizbaMSFT, line 8422: You see how this is using a different version of the UI package than the one below? We want to avoid that 😬 Whenever you make a breaking change to the UI package you should also update the reference [here](https://github.com/microsoft/vscode-azuretools/blob/master/appservice/package.json#L50) and push out an update to the app service package as well

alexweininger: Ok, I'll fix that now. Thanks!

PR #1555, opened 2020-05-20
Suppress create pick for several commands

EricJizbaMSFT:
I can't think of any reason it would be useful in these cases

### DIFF
--- a/src/commands/connections/removeCosmosDBConnection.ts
+++ b/src/commands/connections/removeCosmosDBConnection.ts
@@ -9,7 +9,7 @@ import { ext } from "../../extensionVariables";
 
 export async function removeCosmosDBConnection(context: IActionContext, node?: CosmosDBConnection): Promise<void> {
     if (!node) {
-        node = <CosmosDBConnection>await ext.tree.showTreeItemPicker(CosmosDBConnection.contextValue, context);
+        node = <CosmosDBConnection>await ext.tree.showTreeItemPicker(CosmosDBConnection.contextValue, { ...context, suppressCreatePick: true });
     }
 
     await node.deleteTreeItem(context);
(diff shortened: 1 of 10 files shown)


src/commands/connections/removeCosmosDBConnection.ts
```diff
@@ -9,7 +9,7 @@ import { ext } from "../../extensionVariables";
 
 export async function removeCosmosDBConnection(context: IActionContext, node?: CosmosDBConnection): Promise<void> {
     if (!node) {
-        node = <CosmosDBConnection>await ext.tree.showTreeItemPicker(CosmosDBConnection.contextValue, context);
+        node = await ext.tree.showTreeItemPicker<CosmosDBConnection>(CosmosDBConnection.contextValue, { ...context, suppressCreatePick: true });
```
wwlorey, line 12: Why did the build complain about the placement of `<CosmosDBConnection>`?

EricJizbaMSFT: Because some of them did this (with parens):
```
<DeploymentTreeItem>(await ext.tree.showTreeItemPicker(DeploymentTreeItem.contextValue, context));
```

and some of them did this (without parens):
```
<DeploymentTreeItem>await ext.tree.showTreeItemPicker(DeploymentTreeItem.contextValue, context);
```

And when I copy/pasted `, { ...context, suppressCreatePick: true });` I was missing some parens 😂

EricJizbaMSFT: The main reason I moved it was just to make them consistent. I've had problems related to this before: https://github.com/microsoft/vscode-azurefunctions/pull/1273#discussion_r284046503

I don't fully remember why (could be related to tslint/typescript rules), but the way it is now seems to cause the least amount of problems

wwlorey: I gotcha 👌

PR #1557, opened 2020-05-20
Import & remove trial apps

alexweininger:
Depends on https://github.com/microsoft/vscode-azuretools/pull/725

Since we are about to release, I don't expect any of my changes to get merged until after we release. But I wanted to get feedback on this sooner rather than later.

For the sake of small PRs. This PR includes the smallest import/remove trial app experience. So everything other than importing and removing a trial app is left out.

- Added import trial app command
- Added remove trial app command
- Added ISiteTreeItem interface that TrialAppTreeItem and SiteTreeItem implement.
- Changed the browseWebsite command, so that it uses the browse functions on the node itself, so logic is contained within the node. (this change is a small change towards larger ones to make these tree items generic).


src/commands/trialApp/importTrialApp.ts
```diff
@@ -0,0 +1,25 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { commands, ProgressLocation, window } from 'vscode';
+import { IActionContext } from 'vscode-azureextensionui';
+import { TrialAppLoginSession } from '../../constants';
+import { ext } from '../../extensionVariables';
+import { localize } from '../../localize';
+
+export async function importTrialApp(_context: IActionContext, loginSession?: string): Promise<void> {
```
nturinski, line 12: If loginSession is required, I don't think you should make it optional parameter.  You should have the logic to make sure that it is a string when you call it anyway.

src/ExternalUriHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as querystring from 'querystring';
+import { Disposable, Uri, UriHandler, window } from 'vscode';
+import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
+import { importTrialApp } from './commands/trialApp/importTrialApp';
+
+export class ExternalUriHandler implements UriHandler {
```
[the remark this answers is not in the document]
alexweininger: Moved class to `importTrialApp`.

src/ExternalUriHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+    private async importTrialApp(uri: Uri): Promise<void> {
+        const data = querystring.parse(uri.query);
+
+        if (!data.url) {
+            console.warn('Failed to import URI:', uri);
+        }
+
+        if (typeof data.loginSession === 'string') {
+            await callWithTelemetryAndErrorHandling<void>('importTrialApp', async (context: IActionContext): Promise<void> => {
+                if (typeof data.loginSession === 'string') {
```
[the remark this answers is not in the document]
alexweininger: Removed extra check ✔

package.json
```diff
@@ -316,6 +329,14 @@
         },
         "menus": {
             "commandPalette": [
+                {
```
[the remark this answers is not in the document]
alexweininger: I changed the when clauses for both the commands to: `view == azureAppService && viewItem == trialApp`. I realize this might be too specific (user has to be focused on the trial app item?) for these commands to show up in the palette now. 

Reading from when-clause docs [here](https://code.visualstudio.com/docs/getstarted/keybindings#_when-clause-contexts)

I think this is the *simple* only way to:

> disallow depending on when there is/is not a trial app attached

without adding an extension setting and referencing it with "config." in the when clause.

src/commands/trialApp/importTrialApp.ts
alexweininger: No longer optional 👍

package.json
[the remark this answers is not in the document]
alexweininger: As for the import command, I think that this should never be in the palette because the only way to import a trial app (for now) is through the browser and an external Uri.

src/commands/trialApp/importTrialApp.ts
```diff
@@ -0,0 +1,65 @@
...
+            }
+        });
+    }
+}
+
+interface IDisposable {
+    dispose(): void;
+}
+
+function dispose<T extends IDisposable>(disposables: T[]): T[] {
```
nturinski, line 62: I think you can just overload the dispose method above with this code, kind of like this: https://github.com/microsoft/vscode-azurefunctions/blob/master/src/tree/AzureAccountTreeItemWithProjects.ts

src/commands/trialApp/importTrialApp.ts
```diff
@@ -0,0 +1,65 @@
...
+    private async importTrialApp(uri: Uri): Promise<void> {
+        const data = querystring.parse(uri.query);
+
+        if (!data.url) {
+            await ext.ui.showWarningMessage(`Failed to import URI: ${uri}`);
+            return;
+        }
+
+        await callWithTelemetryAndErrorHandling<void>('importTrialApp', async (context: IActionContext): Promise<void> => {
+            if (typeof data.loginSession === 'string') {
```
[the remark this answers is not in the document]
alexweininger: Sadly, there is no way to import based on the hostname URL, because we need their login session from when they authenticate in the browser.

package.json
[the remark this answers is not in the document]
alexweininger: Sadly, there is no way to import based on the hostname URL, because we need their login session from when they authenticate in the browser.

src/commands/trialApp/importTrialApp.ts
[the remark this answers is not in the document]
alexweininger: Done

src/explorer/trialApp/TrialAppTreeItemBase.ts
```diff
@@ -0,0 +1,44 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzExtParentTreeItem, AzExtTreeItem, TreeItemIconPath } from 'vscode-azureextensionui';
+import { localize } from '../../localize';
+import { getIconPath } from '../../utils/pathUtils';
+import { ITrialAppMetadata } from './ITrialAppMetadata';
+
+export abstract class TrialAppTreeItemBase extends AzExtTreeItem {
```
nturinski, line 11: I think you should have a SiteTreeItemBase that are both extend by SiteTreeItem and TrialAppTreeItem.  TrialAppTreeItemBase is only used to extend TrialAppTreeItem.

I think since you've been making the interfaces in the tools package, you probably have a better idea of what I'm trying to get you to do, but let me know if you need any clarification?

src/explorer/trialApp/TrialAppTreeItemBase.ts
```diff
@@ -0,0 +1,44 @@
...
+    public get iconPath(): TreeItemIconPath {
+        return getIconPath('WebApp');
+    }
+
+    public get label(): string {
+        return this.metadata.siteName ? this.metadata.siteName : localize('nodeJsTrialApp', 'NodeJS Trial App');
+    }
+
+    private get minutesLeft(): number {
+        return (this.metadata?.timeLeft / 60);
```
nturinski, line 22: Why is this one the only one that checks that metadata is undefined?

src/commands/trialApp/importTrialApp.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export class ImportUriHandler implements UriHandler {
+    private disposables: Disposable[] = [];
+
+    constructor() {
+        this.disposables.push(window.registerUriHandler(this));
+    }
+
+    public async handleUri(uri: Uri): Promise<void> {
+        switch (uri.path) {
```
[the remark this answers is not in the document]
alexweininger: I will change the switch to an if/else.

As for handling other Uris, we register this Uri handler which doesn't stop the other Uri from handling Uris. You can still use git clone Uri from Chrome etc.

src/constants.ts
```diff
@@ -3,7 +3,7 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { MessageItem } from "vscode";
+import { MessageItem } from 'vscode';
```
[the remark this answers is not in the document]
alexweininger: I think it's my formatter 😩. I need to fix that.

src/explorer/trialApp/TrialAppTreeItemBase.ts
alexweininger: Makes sense. However, I think I need to wait until we publish the tools package to make this change. Once it's published, it will allow SiteTreeItem to no longer extend AzureParentTreeItem.

src/commands/ImportUriHandler.ts
```diff
@@ -0,0 +1,41 @@
+import * as querystring from 'querystring';
+import { Disposable, Uri, UriHandler, window } from 'vscode';
+import { ext } from 'vscode-azureappservice/out/src/extensionVariables';
+import { callWithTelemetryAndErrorHandling, IActionContext } from '../../extension.bundle';
+import { localize } from '../localize';
+import { importTrialApp } from './trialApp/importTrialApp';
+
+export class ImportUriHandler implements UriHandler {
```
nturinski, line 8: nit: Let's name this to be `ImportTrialAppUriHandler`.  I think it should be more specific since it's being used for only that purpose.

src/commands/deployments/connectToGitHub.ts
```diff
@@ -19,7 +20,14 @@ export async function connectToGitHub(context: IActionContext, target?: GenericT
         node = <DeploymentsTreeItem>target.parent;
     }
 
-    await editScmType(node.root.client, node, context, ScmType.GitHub);
+    if (node instanceof WebAppTreeItem) {
+        node.deploymentsNode = new DeploymentsTreeItem(node.parent, node.client, await node.client.getSiteConfig(), await node.client.getSourceControl());
```
nturinski, line 23: You shouldn't be reassigning the deploymentsNode here.  You can either change the `SiteTreeItem` class to have an async create method like so: https://github.com/microsoft/vscode-azurefunctions/blob/master/src/tree/remoteProject/RemoteFunctionTreeItem.ts#L24-L31  This will ensure that the `deploymentsNode` is already defined.

The other thing you can do is check to see if `node.deploymentsNode` is undefined.  If it is, you can just refresh the treeItem and then proceed as normal.

Also, you should be checking if it's a `SiteTreeItem` rather than `WebAppTreeItem`, which extends `SiteTreeItem`.  This command can also be used on `SlotTreeItems` (which also extends `SiteTreeItem`)

src/commands/deployments/disconnectRepo.ts
```diff
@@ -3,13 +3,20 @@
...
 import { ext } from "../../extensionVariables";
+import { localize } from '../../localize';
 
 export async function disconnectRepo(context: IActionContext, node?: DeploymentsTreeItem): Promise<void> {
     if (!node) {
         node = <DeploymentsTreeItem>await ext.tree.showTreeItemPicker(DeploymentsTreeItem.contextValueConnected, context);
     }
-    await node.disconnectRepo(context);
+
+    if (node.parent instanceof WebAppTreeItem) {
```
nturinski, line 17: Same here, `instanceof SiteTreeItem`.  Otherwise you get the Action not supported error when trying to disconnect the repo from a deployment slot.

src/commands/deployments/editScmType.ts
```diff
@@ -18,7 +18,11 @@ export async function editScmType(context: IActionContext, node?: SiteTreeItem |
         node = <SiteTreeItem>node.parent;
     }
 
-    await appservice.editScmType(node.root.client, node, context, newScmType, showToast);
+    if (node instanceof DeploymentsTreeItem && node.parent instanceof AzureParentTreeItem) {
+        await appservice.editScmType(context, node.root.client, node.root, newScmType, showToast);
+    } else if (node instanceof SiteTreeItem && node.deploymentsNode) {
```
nturinski, line 23: You'll have to do a refresh on the node if the `deploymentsNode` doesn't exist.  If you make an async create method, you won't have to do this check.

Right now, if you try to call this from the command palette, it just doesn't do anything.

src/commands/registerCommands.ts
```diff
@@ -3,49 +3,56 @@
...
+import { startSsh } from './startSsh';
+import { startWebApp } from './startWebApp';
+import { stopWebApp } from './stopWebApp';
+import { swapSlots } from './swapSlots';
+import { importTrialApp } from './trialApp/importTrialApp';
+import { removeTrialApp } from './trialApp/removeTrialApp';
 
 export function registerCommands(): void {
+    // tslint:disable-next-line: no-unused-expression
+    new ImportUriHandler();
```
nturinski, line 53: Could you move this to `extension.ts`?  It feels a bit out of place in `registerCommands`.

src/commands/ImportUriHandler.ts
```diff
@@ -0,0 +1,41 @@
...
+    private disposables: Disposable[] = [];
+
+    constructor() {
+        this.disposables.push(window.registerUriHandler(this));
```
nturinski, line 12: I might be wrong, but could you not do this in the `extension.ts` file?

`context.subscriptions.push(new ImportUriHandler());`

`context.subscriptions` is an array of disposals that get cleaned up after the extension is deactivated.  You could remove some of this extra code then.

src/explorer/ISiteTreeItem.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface ISiteTreeItem {
+
+    defaultHostUrl: string;
```
nturinski, line 8: nit: Unless you're gonna add descriptions, I'd prefer there's no empty space between properties.  Just takes up a lot of space and makes the file look cluttered.

src/explorer/SiteTreeItem.ts
```diff
@@ -39,12 +39,28 @@ export abstract class SiteTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+    public get defaultHostUrl(): string {
+        return this.root.client.defaultHostUrl;
+    }
+
+    public get defaultHostName(): string {
+        return this.root.client.defaultHostName;
+    }
+
+    public async openInPortal(): Promise<void> {
+        await openInPortal(this._root, this.id);
```
nturinski, line 59: nit: Just use `this.root` since you're doing that every else.

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    public metadata: ITrialAppMetadata;
+
+    public defaultHostName: string;
+
+    public defaultHostUrl: string;
+
+    private readonly _defaultHostName: string;
+
+    public get iconPath(): TreeItemIconPath {
```
nturinski, line 29: You should be able to move this into `SiteTreeItemBase` with the idea that this is a shared UI element.  But leave the overloaded `get iconPath` in `DeploymentSlotTreeItem`

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -0,0 +1,105 @@
...
+            if (e instanceof RequestError) {
+                throw Error(localize('errorMetadataRequest', 'Could not get trial app metadata: RequestError.'));
+            } else if (e instanceof SyntaxError) {
+                throw Error(localize('errorMetadataParse', 'Could not get trial app metadata. Could not parse response body.'));
+            } else {
+                throw e;
+            }
+        }
+    }
+    public loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<import("vscode-azureextensionui").AzExtTreeItem[]> {
```
nturinski, line 87: I don't think you have to import `vscode-azureextensionui`, just include it in your import at the top of the file.

src/commands/deployments/connectToGitHub.ts
```diff
@@ -19,13 +20,18 @@ export async function connectToGitHub(context: IActionContext, target?: GenericT
         node = <DeploymentsTreeItem>target.parent;
     }
 
-    await editScmType(node.root.client, node, context, ScmType.GitHub);
-
     if (node instanceof SiteTreeItem) {
+        if (node.deploymentsNode === undefined) {
```
nturinski, line 24: Rather than doing all this logic here, I think you can just use `editScmType` and pass whatever node you want.  Note right now that `connectToGitHub` is importing `editScmType` from '`vscode-azureappservice`, but you can use the one that lives in the extension.

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -0,0 +1,100 @@
...
+            "sec-fetch-dest": "empty",
+            "sec-fetch-mode": "cors",
+            "sec-fetch-site": "same-origin",
+            cookie: `loginsession=${loginSession}`
+        };
+
+        try {
+            const result: string = await requestUtils.sendRequest<string>(metadataRequest);
+            return <ITrialAppMetadata>JSON.parse(result);
+        } catch (e) {
```
nturinski, line 72: I'm not really sure what the value added here is.  Are the errors that Azure throws really human unfriendly?

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -0,0 +1,100 @@
...
+                throw Error(localize('errorMetadataRequest', 'Could not get trial app metadata: RequestError.'));
+            } else if (e instanceof SyntaxError) {
+                throw Error(localize('errorMetadataParse', 'Could not get trial app metadata. Could not parse response body.'));
+            } else {
+                throw e;
+            }
+        }
+    }
+    public loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtTreeItem[]> {
+        return Promise.resolve([]);
```
nturinski, line 83: You can just do return [] if you make it an async function.

src/extension.ts
```diff
@@ -51,6 +52,9 @@ export async function activateInternal(
         ext.fileEditor = new FileEditor();
         context.subscriptions.push(ext.fileEditor);
 
+        const uriHandler = new ImportTrialAppUriHandler();
```
nturinski, line 55: No reason to declare a variable here since you don't use it for anything else (except to push into the disposables array)

src/commands/deployments/editScmType.ts
```diff
@@ -18,7 +18,14 @@ export async function editScmType(context: IActionContext, node?: SiteTreeItem |
         node = <SiteTreeItem>node.parent;
     }
 
-    await appservice.editScmType(node.root.client, node, context, newScmType, showToast);
+    if (node instanceof DeploymentsTreeItem && node.parent instanceof AzureParentTreeItem) {
```
nturinski, line 21: You only need to check if the node is an instanceof AzureParentTreeItem (or just SiteTreeItem alternatively) since we check if it's a deploymentsTreeItem in L15:18

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    public async browse(): Promise<void> {
+        await openUrl(this.defaultHostUrl);
+    }
+
+    public async refreshImpl(): Promise<void> {
+        this.metadata = await TrialAppTreeItem.getTrialAppMetaData(this.metadata.loginSession);
+    }
+
+    public isAncestorOfImpl?(_contextValue: string | RegExp): boolean {
```
nturinski, line 97: Underscores are typically only used in typescript for private properties or unused parameters (but are required because of the definition).  Since this is neither, you should remove the _

src/explorer/trialApp/TrialAppTreeItem.ts
alexweininger: Yeah, I see what you're getting at. They are a little unfriendly, however in the event that an error occurs it would probably be more useful to just pass the error to the user.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -14,4 +17,26 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
         return new SubscriptionTreeItem(this, root);
     }
+
+    public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
+        const children: AzExtTreeItem[] = await super.loadMoreChildrenImpl(clearCache, context);
+
+        const loginSession: string | undefined = ext.context.globalState.get(TrialAppLoginSession);
+
+        if (loginSession) {
+            children.push(await TrialAppTreeItem.createTrialAppTreeItem(this, loginSession));
```
nturinski, line 27: Discussed offline: Use treeErrorHandling code here

src/commands/ImportTrialAppUriHandler.ts
```diff
@@ -0,0 +1,22 @@
+import * as querystring from 'querystring';
+import { Uri, UriHandler } from 'vscode';
+import { callWithTelemetryAndErrorHandling, IActionContext } from '../../extension.bundle';
+import { localize } from '../localize';
+import { importTrialApp } from './trialApp/importTrialApp';
+
+export class ImportTrialAppUriHandler implements UriHandler {
```
nturinski, line 7: Write a comment about how the uriHandler only activates when the uri has our extension id in it.  Should also link this https://code.visualstudio.com/api/references/activation-events#onUri

src/commands/deployments/disconnectRepo.ts
```diff
@@ -3,13 +3,20 @@
...
 export async function disconnectRepo(context: IActionContext, node?: DeploymentsTreeItem): Promise<void> {
     if (!node) {
         node = await ext.tree.showTreeItemPicker<DeploymentsTreeItem>(DeploymentsTreeItem.contextValueConnected, { ...context, suppressCreatePick: true });
     }
-    await node.disconnectRepo(context);
+
+    if (node.parent instanceof SiteTreeItem) {
+        await disconnectRepository(context, node.parent.root.client, node.parent.root);
+    } else {
+        throw Error(localize('actionNotSupported', 'Action not supported.'));
```
nturinski, line 20: Could you just make the error like this? https://github.com/microsoft/vscode-azureresourcegroups/pull/15/files?diff=split&w=1#diff-7c2b04b667e5fe72febaf888c6ee1a7cR12

I like to try to keep consistent with whatever Eric's doing so our extensions feel like they're from the same company haha.

EricJizbaMSFT: Lol tbf, I should be copying Alex's message since he wrote it first

alexweininger: Lol. I will change it since that error is already in master.

src/commands/deployments/editScmType.ts
```diff
@@ -18,9 +18,11 @@ export async function editScmType(context: IActionContext, node?: SiteTreeItem |
         node = <SiteTreeItem>node.parent;
     }
 
-    await appservice.editScmType(node.root.client, node, context, newScmType, showToast);
-
-    if (node.deploymentsNode) {
-        await node.deploymentsNode.refresh();
+    if (node instanceof SiteTreeItem) {
```
nturinski, line 21: I'm not sure we need this check.  You could have an else and throw the not supported error, but honestly we have the typings here to prevent people passing in the wrong types of treeItems so I think it's more on the coder to ensure that they pass in the right treeItem rather than throwing an error here.

src/commands/deployments/disconnectRepo.ts
```diff
@@ -3,13 +3,20 @@
...
 import { ext } from "../../extensionVariables";
+import { localize } from '../../localize';
 
 export async function disconnectRepo(context: IActionContext, node?: DeploymentsTreeItem): Promise<void> {
     if (!node) {
         node = await ext.tree.showTreeItemPicker<DeploymentsTreeItem>(DeploymentsTreeItem.contextValueConnected, { ...context, suppressCreatePick: true });
     }
-    await node.disconnectRepo(context);
+
+    if (node.parent instanceof SiteTreeItem) {
```
nturinski, line 17: I do think this one might be worth keeping though since DeploymentsTreeItems will exist under the TrialApp, so it seems likely that we might accidentally pass up the wrong treeItem.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -14,4 +17,38 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
+
+        if (loginSession) {
+            const ti: AzExtTreeItem[] = await this.createTreeItemsWithErrorHandling(
+                [loginSession],
+                'trialAppInvalid',
+                async (source: string): Promise<AzExtTreeItem> => {
+                    return await TrialAppTreeItem.createTrialAppTreeItem(this, source);
+                },
+                (_source: unknown): string => {
+                    return 'Could not import trial app';
```
nturinski, line 34: Is there not an error message you could return here instead?

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -0,0 +1,89 @@
...
+import { requestUtils } from '../../utils/requestUtils';
+import { AzureAccountTreeItem } from '../AzureAccountTreeItem';
+import { ISiteTreeItem } from '../ISiteTreeItem';
+import { SiteTreeItemBase } from '../SiteTreeItemBase';
+import { ITrialAppMetadata } from './ITrialAppMetadata';
+
+export class TrialAppTreeItem extends SiteTreeItemBase implements ISiteTreeItem {
+    public static contextValue: string = 'trialApp';
+    public contextValue: string = TrialAppTreeItem.contextValue;
+
```
nturinski, line 18: nit: Could you reduce the number of empty lines?  Just leave what is currently line 24, seperating the pubic and private properties.

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -0,0 +1,89 @@
...
+    public hasMoreChildrenImpl(): boolean {
+        return false;
+    }
+
+    public async browse(): Promise<void> {
+        await openUrl(this.defaultHostUrl);
+    }
+
+    public async refreshImpl(): Promise<void> {
+        this.metadata = await TrialAppTreeItem.getTrialAppMetaData(this.metadata.loginSession);
```
nturinski, line 83: I don't think these properties are likely to change given the context of the trial apps, but I wonder if there is a situation where the user's hostName might change?  Well, either way, I think it makes sense to also make these get properties:
```
this.defaultHostName = this.metadata.hostName;
this.defaultHostUrl = `https://${this.defaultHostName}`;
```

That way when you refresh and update metadata, these will automatically be updated as well.  Or you can update them in refreshImpl.

src/explorer/trialApp/ITrialAppMetadata.ts
```diff
@@ -0,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface ITrialAppMetadata {
```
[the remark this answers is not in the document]
alexweininger: Good point. How about I remove the unused ones after implementing the child nodes? Since about half of these will be used by the children.

src/explorer/AzureAccountTreeItem.ts
alexweininger: Looking at other examples of `createTreeItemsWithErrorHandling` it doesn't look like we are supposed to return an error message there. Instead the error is put as a child of the node. 

For example: 

https://github.com/microsoft/vscode-azureappservice/blob/83c6f0df8ad80e6d1775fade38070b32b3b32fc6/src/explorer/SubscriptionTreeItem.ts#L60

PR #1561, opened 2020-05-22
Refactor commands out of SiteTreeItem

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1277

I changed promptToSaveDeployDefault to actually collect telemetry now since we weren't awaiting it.

Ignore those settings, I don't know how they got past code review XD


src/commands/promptScmDoBuildDeploy.ts
```diff
@@ -0,0 +1,92 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
```
EricJizbaMSFT, line 6: Probably makes sense to put these in the deploy folder

PR #1581, opened 2020-06-05
Clone trial app command

alexweininger:
Simple command that clones a trial app.

For now doesn't show up in command pallet, only shows up in context menu for trial apps.

### DIFF
--- a/src/commands/trialApp/cloneTrialApp.ts
+++ b/src/commands/trialApp/cloneTrialApp.ts
@@ -6,12 +6,7 @@
 import { commands } from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
 import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
-import { ext } from '../../extensionVariables';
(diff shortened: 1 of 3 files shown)


src/commands/trialApp/cloneTrialApp.ts
```diff
@@ -0,0 +1,17 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { commands } from 'vscode';
+import { IActionContext } from 'vscode-azureextensionui';
+import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
+import { ext } from '../../extensionVariables';
+
+export async function cloneTrialApp(context: IActionContext, node?: TrialAppTreeItem): Promise<void> {
+    if (!node) {
```
nturinski, line 12: So you only need this if you enter through the command palette, which you've disabled.  I think you can just remove the ? in the parameter.

PR #1583, opened 2020-06-05
Add logs and files nodes

alexweininger:
Using the changes in https://github.com/microsoft/vscode-azuretools/pull/726 to add the logs node and files node as children to the trial app node.

One of the major additions here is the `TrialAppClient` which implements the `IFilesClient` to enable the files node and logs node.

### DIFF
--- a/src/commands/logstream/enableFileLogging.ts
+++ b/src/commands/logstream/enableFileLogging.ts
@@ -7,6 +7,7 @@ import * as vscode from 'vscode';
 import { LogFilesTreeItem } from 'vscode-azureappservice';
 import { DialogResponses, IActionContext } from 'vscode-azureextensionui';
 import { SiteTreeItem } from "../../explorer/SiteTreeItem";
+import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
 import { WebAppTreeItem } from '../../explorer/WebAppTreeItem';
 import { ext } from '../../extensionVariables';
 
@@ -14,7 +15,7 @@ export interface IEnableFileLoggingContext extends IActionContext {
     suppressAlreadyEnabledMessage?: boolean;
 }
 
-export async function enableFileLogging(context: IEnableFileLoggingContext, node?: SiteTreeItem | LogFilesTreeItem): Promise<void> {
+export async function enableFileLogging(context: IEnableFileLoggingContext, node?: SiteTreeItem | LogFilesTreeItem | TrialAppTreeItem): Promise<void> {
     if (!node) {
         node = await ext.tree.showTreeItemPicker<SiteTreeItem>(WebAppTreeItem.contextValue, context);
     }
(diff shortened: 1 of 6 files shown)


src/commands/logstream/enableFileLogging.ts
```diff
@@ -24,14 +25,14 @@ export async function enableFileLogging(context: IEnableFileLoggingContext, node
...
-    const siteNode: SiteTreeItem = node;
+    const siteNode: SiteTreeItem | TrialAppTreeItem = node;
 
     const isEnabled = await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification }, async p => {
         p.report({ message: 'Checking container diagnostics settings...' });
         return await siteNode.isHttpLogsEnabled();
     });
 
-    if (!isEnabled) {
+    if (!isEnabled && siteNode instanceof SiteTreeItem && node instanceof SiteTreeItem) {
```
EricJizbaMSFT, line 35: Why do you have to check both `siteNode` and `node`? I would think you should only be using `siteNode` after line 28

src/explorer/trialApp/TrialAppClient.ts
```diff
@@ -0,0 +1,80 @@
...
+    public async listHostKeys(): Promise<IHostKeys> {
+        throw new Error('Method not implemented.');
+    }
+
+    public async getKuduClient(): Promise<KuduClient> {
+        if (!this.metadata.scmHostName) {
+            throw new Error(localize('notSupportedLinux', 'This operation is not supported by this app service plan.'));
+        }
+
+        const kuduClient: KuduClient = new KuduClient(this.credentials, `https://${this.metadata.scmHostName}`);
```
EricJizbaMSFT, line 76: Use `this.kuduUrl` instead of re-constructing the url

src/explorer/trialApp/TrialAppClient.ts
```diff
@@ -0,0 +1,80 @@
...
+        return <ITrialAppMetadata>JSON.parse(result);
+    }
+
+    public async listHostKeys(): Promise<IHostKeys> {
+        throw new Error('Method not implemented.');
+    }
+
+    public async getKuduClient(): Promise<KuduClient> {
+        if (!this.metadata.scmHostName) {
+            throw new Error(localize('notSupportedLinux', 'This operation is not supported by this app service plan.'));
```
EricJizbaMSFT, line 73: This was only needed for linux consumption function apps. You shouldn't need to check it for trial web apps

src/explorer/trialApp/TrialAppClient.ts
```diff
@@ -0,0 +1,80 @@
...
+            "sec-fetch-site": "same-origin",
+            cookie: `loginsession=${loginSession}`
+        };
+
+        const result: string = await requestUtils.sendRequest<string>(metadataRequest);
+        return <ITrialAppMetadata>JSON.parse(result);
+    }
+
+    public async listHostKeys(): Promise<IHostKeys> {
+        throw new Error('Method not implemented.');
```
EricJizbaMSFT, line 68: You're never going to implement this, right? I think it's function app-specific. In which case, it would be better to throw a "not supported" error

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -79,7 +82,7 @@ export class TrialAppTreeItem extends SiteTreeItemBase implements ISiteTreeItem
     }
 
     public async refreshImpl(): Promise<void> {
-        this.metadata = await TrialAppTreeItem.getTrialAppMetaData(this.metadata.loginSession);
+        this.client.metadata = await TrialAppClient.getTrialAppMetaData(this.metadata.loginSession);
```
EricJizbaMSFT, line 85: I think it would be better to completely re-create the client instead of just setting the metadata. For example, if `publishingPassword` changed, we want to pick up that change when they refresh - and it's only set in the constructor of `TrialAppClient`.

At that point, I might just move all of the code from `getTrialAppMetaData` into `createTrialAppClient` and delete  `getTrialAppMetaData` since I think it's only used once.

src/explorer/trialApp/TrialAppClient.ts
alexweininger: Made it optional in the `IFilesClient` interface, and removed it's implementation from `TrialAppClient`

src/explorer/trialApp/TrialAppClient.ts
```diff
@@ -0,0 +1,67 @@
...
+import KuduClient from 'vscode-azurekudu';
+import { requestUtils } from '../../utils/requestUtils';
+import { ITrialAppMetadata } from './ITrialAppMetadata';
+
+export class TrialAppClient implements IFilesClient {
+
+    public isFunctionApp: boolean = false;
+    public metadata: ITrialAppMetadata;
+
+    private credentials: ServiceClientCredentials;
```
EricJizbaMSFT, line 18: nit: private variables should be prefixed with `_`

PR #1585, opened 2020-06-08
Add the test case to delete application settings

v-wuzhai:


test/nightly/azureResourceWebAppActions.test.ts
```diff
@@ -114,6 +114,14 @@ suite('Web App actions', async function (this: Mocha.Suite): Promise<void> {
         assert.equal(await getAppSettingValue(resourceName, resourceName, appSettingKey), appSettingValue);
     });
 
+    test(`Delete setting for ${WebsiteOS0} Web App`, async () => {
```
EricJizbaMSFT, line 117: I think I'd prefer if this was combined with "Add new setting for ${WebsiteOS0} Web App". They're both pretty simple tests and this one won't work if that one fails.

PR #1588, opened 2020-06-09
Add AppSettingsTreeItem to trial app children

alexweininger:
* Add app settings node to trial apps
* Don't show `openInPortal` action in the `AppSettingsTreeItem` for trial apps
* Don't show `toggleSlotSetting` action in the `AppSettingTreeItem` for trial apps


src/explorer/trialApp/TrialAppClient.ts
```diff
@@ -51,17 +53,50 @@ export class TrialAppClient implements IFilesClient {
...
     }
+
+    public async listApplicationSettings(): Promise<StringDictionary> {
+        const kuduClient: KuduClient = await this.getKuduClient();
+        const settings: StringDictionary = {};
+        settings.properties = <{ [name: string]: string }>await kuduClient.settings.getAll();
+        return settings;
+    }
+
+    public async updateApplicationSettings(appSettings: StringDictionary): Promise<StringDictionary> {
```
nturinski, line 73: This definitely all seems kind of weird, but after your offline discussion, I get why you have to do it.

package.json
```diff
@@ -566,32 +566,32 @@
                 },
                 {
                     "command": "appService.Refresh",
-                    "when": "view == azureAppService && viewItem == applicationSettings",
+                    "when": "view == azureAppService && viewItem =~ /^(applicationSettings|applicationSettingsTrialApp)$/",
```
EricJizbaMSFT, line 569: Bruh let's reduce that character count

/^applicationSettings(|TrialApp)$/

package.json
```diff
@@ -566,32 +566,32 @@
...
                 {
                     "command": "appService.appSettings.Upload",
-                    "when": "view == azureAppService && viewItem == applicationSettings",
+                    "when": "view == azureAppService && viewItem =~ /^(applicationSettings|applicationSettingsTrialApp)$/",
                     "group": "1_appSettingEnv@2"
                 },
                 {
                     "command": "appService.appSettings.Edit",
-                    "when": "view == azureAppService && viewItem == applicationSettingItem",
+                    "when": "view == azureAppService && viewItem =~ /^(applicationSettingItem|applicationSettingItemNoSlots)$/",
```
EricJizbaMSFT, line 584: /^applicationSettingItem(|NoSlots)$/

src/explorer/trialApp/TrialAppClient.ts
```diff
@@ -51,17 +53,50 @@ export class TrialAppClient implements IFilesClient {
         return `https://${this.metadata.scmHostName}`;
     }
 
-    public get defaultHostName(): string {
-        return this.metadata.hostName;
-    }
-
     public get defaultHostUrl(): string {
-        return `https://${this.metadata.hostName}`;
+        return `https://${this.fullName}`;
```
EricJizbaMSFT, line 57: Usually "fullName" looks like `emjnode1`, "hostName" looks like `emjnode1.azurewebsites.net`, and "hostUrl" looks like `https://emjnode1.azurewebsites.net`. Something seems amiss with the naming here.

Does `IAppSettingsClient` even need `fullName`?

src/explorer/trialApp/TrialAppClient.ts
EricJizbaMSFT: Yeah this is not great. For example, there's a possibility of getting into a weird half-updated state if one of the deletes fails before you even try the update.

I think the only reason I'm okay with this is because it's a trial app that'll get thrown away after an hour anyways (aka it's not a big deal if settings get messed up by us). That being said, if you're doing "weird" things in the code, you should generally add comments (to the TypeScript file itself, not on GitHub) describing why you're doing it.

alexweininger: Ok, I can add a comment explaining the weird code. I also think we are going to filter out many of the app settings in the future because lots of the trial app settings are not things we want to show to users.

alexweininger: I actually already have a comment explaining the code [here](https://github.com/microsoft/vscode-azureappservice/pull/1588/files/5eff536c5352854732d17d227913860733e88618#diff-6b9c6c17562552f98bb7ff2484af2216R76).

src/explorer/trialApp/TrialAppClient.ts
alexweininger: > Does `IAppSettingsClient` even need `fullName`?
If you're asking if it's used, then yes. But if you're asking if it could be removed from `IAppSettingsClient` and refactor the code that references `client.fullName` be refactored then it could probably be removed.

I am fixing those properties. There is actually a url property in the metadata I should have been using. And a siteName prop for fullName.

src/explorer/trialApp/TrialAppClient.ts
EricJizbaMSFT: That comment is _what_ you're doing, not _why_ you're doing it. A few questions the comment should answer:

- Why are you making your own requests instead of using KuduClient or SiteClient?
- Why isn't rename automatically handled for you like normal apps?

Your mentor needed an offline discussion to understand why you're doing this, so obviously I'm not the only one that wants more context :P

EricJizbaMSFT: Also it's fine to say something along the lines of "I don't know the root cause, but I tested X and Y happened" in your comment. This is all meant to help someone reading your code down the line. Maybe they read your comment, think of something you didn't try, and fix it. Maybe they try what you tried, get a different result, and realize the platform must've changed

Worst case is they look at your code, think it looks ugly, refactor to follow the same practice as normal apps, don't test a trial app (because it's annoying to set up and why would it be different? lol), and now we have broken app settings for trial apps

Issue #1587, opened 2020-06-09
Deploy to trial app


alexweininger:
From offline discussion:

How and if we should allow users to deploy their own repos/code to their trial app by pressing the blue deploy arrow.

PR #1595, opened 2020-06-12
Deploy trial apps

alexweininger:
Implements #1587
Depends on https://github.com/microsoft/vscode-azuretools/pull/739, and https://github.com/microsoft/vscode-azuretools/pull/743

* Created `deployTrialApp`
* Make `showDeployCompletedMessage` function generic
* Use `ISimplifySiteClient` instead of `IFilesClient` and `IDeploymentsClient` from https://github.com/microsoft/vscode-azuretools/pull/743
* Properly update the `trialAppNode` property in `AzureAccountTreeItem` when removing, importing and refreshing trial apps.
* Implement `pickTreeItemImpl` to change the childTypeLabel. See below image for that change

This is the behavior that we decided on for when the user hits the blue deploy arrow  if the user is signed in and has a trial app:
image


src/commands/deploy/deployTrialApp.ts
```diff
@@ -0,0 +1,52 @@
...
+    const workspaceFolders: WorkspaceFolder[] | undefined = workspace.workspaceFolders;
+
+    let path: string;
+    let commit: boolean = false;
+
+    if (workspaceFolders?.length === 1) {
+        path = workspaceFolders[0].uri.fsPath;
+        commit = workspaceFolders[0].name === trialAppTreeItem.metadata.siteName;
+    } else {
+        path = await selectWorkspaceFolder('Select folder containing a repository to deploy');
```
nturinski, line 28: `Select a folder containing a repository to deploy`

src/commands/deploy/deployTrialApp.ts
```diff
@@ -0,0 +1,52 @@
...
+        const workspaceFolder: WorkspaceFolder | undefined = workspace.getWorkspaceFolder(vscode.Uri.file(path));
```
nturinski, line 29: Should be able to move this out of the if/else clause so you only have to assign it once.

src/commands/deploy/deploy.ts
```diff
@@ -39,6 +41,14 @@ export async function deploy(context: IActionContext, target?: vscode.Uri | Site
         siteConfig = await target.root.client.getSiteConfig();
     }
 
+    if (ext.azureAccountTreeItem.trialAppNode) {
+        target = await getDeployNodeWithTrialApp(context, target);
```
nturinski, line 45: I think you should only be doing this if you don't have a target already.

src/commands/deploy/deploy.ts
```diff
@@ -39,6 +41,14 @@ export async function deploy(context: IActionContext, target?: vscode.Uri | Site
         siteConfig = await target.root.client.getSiteConfig();
     }
```
nturinski, line 42: I don't think you should be calling await getDeployNodeWithTrialApp just because they have a running app.  It should only be if they deployed if there is no target.

So you can probably do something like:

```
if (target instanceof SiteTreeItem) {
  webAppSource = WebA[[Sprice/tree'
  siteConfig = await target.root.client.getSiteConfig();
} else if (!target) {
  if (ext.azureAccountTreeItem.trialAppNode) {
      target = await getDeployNodeWithTrialApp(context)
  }
}

File: src/commands/deploy/deployTrialApp.ts [hunk 5]
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    const title: string = localize('deploying', 'Deploying to "{0}"... Check [output window](command:{1}) for status.', trialAppTreeItem.client.fullName, `${ext.prefix}.showOutputChannel`);
+    await window.withProgress({ location: ProgressLocation.Notification, title }, async () => {
+        await localGitDeploy(trialAppTreeItem.client, { fsPath: path, branch: 'RELEASE', commit: commit }, context);
+        return showDeployCompletedMessage(trialAppTreeItem);
+    });
+}
+
+export async function getDeployNodeWithTrialApp(context: IActionContext, target?: vscode.Uri | TrialAppTreeItem | SiteTreeItem): Promise<vscode.Uri | SiteTreeItem | TrialAppTreeItem | undefined> {
```
nturinski, line 40: Rather than having its own function, I think it makes sense to add this logic to the `getDeployNode`.  You should be able to change the showTreeItemPicker in there to allow you to select the trial app depending on the user's settings/logged in status.

src/commands/deploy/deployTrialApp.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (!target) {
+        const trialApp: TrialAppTreeItem | undefined = ext.azureAccountTreeItem.trialAppNode;
+        if (trialApp) {
+            if (ext.azureAccountTreeItem.isLoggedIn) {
```
nturinski, line 44: Logic seems to be flipped here.

src/commands/deploy/deployTrialApp.ts
alexweininger: The reason I didn't add that logic to `getDeployNode` is that for trial apps we need to get the deploy node before we get the deploy fs path. 

And `getDeployNode` depends on the fs path. I think there might be a way to make this work but I believe that it would end up worse and less clear than it is now.

This problem spawns from the fact that normally we ask for the path first, but for trial apps we need to know the target first since they could be deploying to a trial app or a real app.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -54,4 +42,42 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
+                const ti: AzExtTreeItem[] = await this.createTreeItemsWithErrorHandling(
+                    [loginSession],
+                    'trialAppInvalid',
+                    async (source: string): Promise<AzExtTreeItem> => {
+                        return await TrialAppTreeItem.createTrialAppTreeItem(this, source);
+                    },
+                    (_source: unknown): string => {
+                        return 'Trial App';
+                    });
+                if (ti.length > 0) {
```
nturinski, line 74: I think you could just return ti.pop().  If there's an element, it'll be the trial app tree item.  If it's empty, then it should be just be undefined.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -54,4 +42,42 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
+                    const treeItem: AzExtTreeItem = ti[0];
+                    this.trialAppNode = treeItem instanceof TrialAppTreeItem ? treeItem : this.trialAppNode;
```
nturinski, line 76: I don't understand this logic.  Wouldn't you want this to be undefined it we couldn't create a treeItem?

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -54,4 +42,42 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
+        } else {
+            this.childTypeLabel = subscription;
+        }
+
+        return super.pickTreeItemImpl(expectedContextValues);
+    }
+
+    private async loadTrialAppNode(): Promise<AzExtTreeItem | undefined> {
+        if (this.trialAppNode) {
+            await this.trialAppNode.refresh();
```
nturinski, line 60: Why do you need to refresh it?

src/explorer/trialApp/TrialAppClient.ts
```diff
@@ -54,7 +54,21 @@ export class TrialAppClient implements IAppSettingsClient, IFilesClient {
     }
 
     public get defaultHostUrl(): string {
-        return `https://${this.metadata.url}`;
+        return this.metadata.url;
+    }
+    public get gitUrl(): string {
```
nturinski, line 59: Don't make me nit comment all of these Alex 😭

alexweininger: Ok so we always put spaces between getters? Because I did that on purpose (I hate it). But I thought there are places we don't put spaces between getters. 

Since it seems like we DO want spaces between getters I will stop doing that (thank goodness).

nturinski: Oh, really?  Okay, I mean that's my preference too, so since I'm the one reviewing, let's put spaces 😂

src/commands/deploy/showDeployCompletedMessage.ts
```diff
@@ -7,13 +7,13 @@ import { MessageItem, window } from "vscode";
...
 import { SiteTreeItem } from "../../explorer/SiteTreeItem";
+import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
 import { ext } from "../../extensionVariables";
 import { localize } from "../../localize";
-import { browseWebsite } from '../browseWebsite';
 import { startStreamingLogs } from '../logstream/startStreamingLogs';
 
-export function showDeployCompletedMessage(node: SiteTreeItem): void {
-    const message: string = localize('deployCompleted', 'Deployment to "{0}" completed.', node.root.client.fullName);
+export function showDeployCompletedMessage(node: SiteTreeItem | TrialAppTreeItem): void {
```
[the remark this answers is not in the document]
alexweininger: This would be a large change. If it's worth it to you then I will do it.

src/commands/deploy/deploy.ts
```diff
@@ -28,7 +30,8 @@ import { showDeployCompletedMessage } from './showDeployCompletedMessage';
 
 const postDeployCancelTokens: Map<string, vscode.CancellationTokenSource> = new Map();
 
-export async function deploy(context: IActionContext, target?: vscode.Uri | SiteTreeItem, _multiTargets?: (vscode.Uri | SiteTreeItem)[], isTargetNewWebApp: boolean = false): Promise<void> {
+// tslint:disable-next-line: cyclomatic-complexity
```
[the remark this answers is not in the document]
alexweininger: I'll fix it 😄

src/commands/deploy/showDeployCompletedMessage.ts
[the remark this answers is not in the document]
nturinski: Nah, it's fine, don't worry about it

src/commands/deploy/deployTrialApp.ts
```diff
@@ -0,0 +1,53 @@
...
+export async function deployTrialApp(context: IActionContext, trialAppTreeItem: TrialAppTreeItem): Promise<void> {
+    const workspaceFolders: WorkspaceFolder[] | undefined = workspace.workspaceFolders;
+
+    let path: string;
+    let workspaceFolder: WorkspaceFolder | undefined;
+    if (workspaceFolders?.length === 1) {
+        workspaceFolder = workspaceFolders[0];
+        path = workspaceFolder.uri.fsPath;
+    } else {
+        const selectFolderWithRepo: string = localize('selectFolderWithRepo', 'Select a folder containing a repository to deploy');
```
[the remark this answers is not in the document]
alexweininger: If a folder does not have a repo in it this is the current behavior:

image

src/commands/deploy/deploy.ts
```diff
@@ -51,6 +55,15 @@ export async function deploy(context: IActionContext, target?: vscode.Uri | Site
     await setPreDeployTaskForDotnet(deployContext);
     const { node, isNewWebApp }: IDeployNode = await getDeployNode(deployContext, target, isTargetNewWebApp);
 
+    if (node instanceof TrialAppTreeItem) {
+        const commit: boolean = deployContext.workspace.name === node.metadata.siteName;
```
nturinski, line 59: Since it's kind of a niche case, I think it's still worth separating this out as its own function, but leaving it in the deploy folder.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -54,4 +40,47 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
+            }
+        } else {
+            this.childTypeLabel = subscription;
+        }
+
+        return super.pickTreeItemImpl(expectedContextValues);
+    }
+
+    public async refreshImpl(): Promise<void> {
+        await this.trialAppNode?.refresh();
```
nturinski, line 63: You should probably call `loadTrialAppNode` here too.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -54,4 +40,47 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
         }
         return super.compareChildrenImpl(item1, item2);
     }
+
+    public async pickTreeItemImpl(expectedContextValues: (string | RegExp)[]): Promise<AzExtTreeItem | undefined> {
+        const subscription: string = localize('subscription', 'subscription');
+        const subscriptionOrTrialApp: string = localize('subscriptionOrTrialApp', 'subscription or trial app');
+        const trialAppOrSignIn: string = localize('trialAppOrSignIn', 'trial app, sign in, or create a free Azure account');
```
nturinski, line 47: Not a blocker, but I'm not sure about this last label.  Might be good to just say trial app.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -54,4 +40,47 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
+    }
+
+    public async refreshImpl(): Promise<void> {
+        await this.trialAppNode?.refresh();
+    }
+
+    private async loadTrialAppNode(): Promise<AzExtTreeItem | undefined> {
+        const loginSession: string | undefined = ext.context.globalState.get(TrialAppLoginSession);
+        if (!loginSession) {
+            return undefined;
```
nturinski, line 69: Should also reset `this.trialAppNode` to be undefined.

src/commands/deploy/getDeployNode.ts
```diff
@@ -6,23 +6,24 @@
...
-export async function getDeployNode(context: IDeployContext, target: Uri | string | SiteTreeItem | undefined, isTargetNewWebApp: boolean): Promise<IDeployNode> {
+export async function getDeployNode(context: IDeployContext, target: Uri | string | SiteTreeItem | TrialAppTreeItem | undefined, isTargetNewWebApp: boolean): Promise<IDeployNode> {
     const defaultWebAppId: string | undefined = getWorkspaceSetting(configurationSettings.defaultWebAppToDeploy, context.workspace.uri.fsPath);
-    let node: SiteTreeItem | undefined;
+    let node: SiteTreeItem | TrialAppTreeItem | undefined;
 
     // if the entry point for deploy was via "Deploy" after creating, it is also a new web app (so confirmDeployment would be false)
     let isNewWebApp: boolean = isTargetNewWebApp;
-    if (target instanceof SiteTreeItem) {
+    if (target instanceof SiteTreeItem || target instanceof TrialAppTreeItem) {
```
nturinski, line 26: nit: You can just use `SiteTreeItemBase`.  It's cleaner, but adds another import, so iunno

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -54,4 +40,47 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
         return super.compareChildrenImpl(item1, item2);
     }
+
+    public async pickTreeItemImpl(expectedContextValues: (string | RegExp)[]): Promise<AzExtTreeItem | undefined> {
+        const subscription: string = localize('subscription', 'subscription');
+        const subscriptionOrTrialApp: string = localize('subscriptionOrTrialApp', 'subscription or trial app');
+        const trialAppOrSignIn: string = localize('trialAppOrSignIn', 'trial app, sign in, or create a free Azure account');
+
+        if (expectedContextValues.includes(TrialAppTreeItem.contextValue) && this.trialAppNode) {
+            if (this.isLoggedIn) {
```
nturinski, line 50: nit: I prefer

`this.childTypeLabel = this.isLoggedIn ? subscriptionOrTrialApp : trialAppOrSignIn`

src/utils/workspace.ts
```diff
@@ -8,6 +8,19 @@ import * as vscode from 'vscode';
 import { IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { IDeployContext } from '../commands/deploy/IDeployContext';
 import { ext } from '../extensionVariables';
+import { localize } from '../localize';
+
+export async function selectWorkspaceFolder(placeHolder: string): Promise<string> {
```
nturinski, line 13: I think you can get rid of this now that you moved everything into getDeployNode.

PR #1604, opened 2020-06-17
Fix simple-git webpack errors with loader rule

alexweininger:
Add a rule to `webpack.config.js` to use `ts-loader`  to load the `simple-git` package.

This fixes the following error that was throwing during `npm run test`.

```
ERROR in ./node_modules/simple-git/src/lib/runners/promise.ts 71:28
Module parse failed: Unexpected token (71:28)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| const {gitInstanceFactory} = require('../../git-factory');
| 
> export function gitP(baseDir?: string): SimpleGit {
| 
|    let git: any;
```

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -27,6 +27,29 @@ let config = dev.getDefaultWebpackConfig({
         '../build/Release/bufferutil': 'commonjs ../build/Release/bufferutil',
         '../build/default/bufferutil': 'commonjs ../build/default/bufferutil',
     },
+    /**
+     *  The simple-git package was causing the following webpack error. This rule makes sure to use ts-loader to load the package.
+     *
+     * ERROR in ./node_modules/simple-git/src/lib/runners/promise.ts 71:28
+     *  Module parse failed: Unexpected token(71:28)
+     *  You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
+     *  | const { gitInstanceFactory } = require('../../git-factory');
+     *  |
+     *  > export function gitP(baseDir?: string): SimpleGit {
+     *  |
+     *  | let git: any;
+     */
+
+    loaderRules: [
+        {
+            test: /\.ts$/,
+            include: /node_modules[\\\/]simple-git/,
+            use: [{
+                // Note: the TS loader will transpile the .ts file directly during webpack (i.e., webpack is directly pulling the .ts files, not .js files from out/)
+                loader: require.resolve('ts-loader')
+            }]
+        }
+    ]
 });
 
 if (DEBUG_WEBPACK) {


webpack.config.js
```diff
@@ -27,6 +27,29 @@ let config = dev.getDefaultWebpackConfig({
...
+     *  Module parse failed: Unexpected token(71:28)
+     *  You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
+     *  | const { gitInstanceFactory } = require('../../git-factory');
+     *  |
+     *  > export function gitP(baseDir?: string): SimpleGit {
+     *  |
+     *  | let git: any;
+     */
+
+    loaderRules: [
```
EricJizbaMSFT, line 43: Good job pulling this out into a separate PR 👍  Webpack stuff is notoriously complicated so it's definitely worth handling on its own.

The only problem is - the issue doesn't repro on master, right? So this PR is fixing something that doesn't exist yet. I assume the problem starts happening with [this PR](https://github.com/microsoft/vscode-azureappservice/pull/1595)? If so - it would be better to isolate the root cause of the problem and include that in this PR. That way we can verify the solution with the problem.

I would also like to make sure we exhaust all other options before we start messing with webpack config. For example, I assume updating to v2.6 of simple-git is the root cause? If so:
- Do we know why a ts-loader is needed? It's highly uncommon for packages to ship a ts file instead of a js file
- 1.92 -> 2.6 was a breaking change. Are we sure we adjusted accordingly?
- Did we need to update to 2.6 in the first place? Or can we revert to the previous version?

alexweininger: I think I found a version of simple-git that has the features we need (reason we updated) but is before they changed their code to TypeScript. The version is here: https://www.npmjs.com/package/simple-git/v/1.132.0. I have tested it locally and this fixes our builds. Should we just use this version and not worry about the builds. Only issue I can see with this solution is that if we eventually need to update simple-git in the future we will have to go back and fix the build again.

alexweininger: Also as for when this problem started happening, it is associated with this [this PR](https://github.com/microsoft/vscode-azureappservice/pull/1595) since it uses the 0.62.0 release of appservice.

EricJizbaMSFT: I think I would vote for using an older version of simple-git instead of messing with webpack.

> Only issue I can see with this solution is that if we eventually need to update simple-git in the future we will have to go back and fix the build again.

Two points on this:
- AFAIK our usage of simple-git is pretty limited. Just a few calls to get status and do a few pushes. Might not even be too difficult to re-implement ourselves and abandon simple-git (using child processes and parsing the output)
- We can always file an issue on "simple-git". Feels pretty unorthodox how they're handling the typescript files and I'm guessing other people might run into the same problem we did

alexweininger: Ok, I will close this PR in favor of using an older version of simple-git.

nturinski: FWIW I agree with Eric here.  We'll have to change the webpack settings for _anything_ that consumes the appservice package, whether it uses any of the functionality or not.

PR #1605, opened 2020-06-18
Create and use findSubscriptionTreeItem util

### LINKED ISSUE #1602: An error occurs when executing "Open in Portal" action for "Application Settings" node
**OS:** Win 10/Mac
**Build Version:** 20200616.2

**Repro Steps:**
1. Expand one web app.
2. Right click the "Application Settings" node and select "Open in Portal".

**Expect:**
Open in portal successfully without any error.

**Actual:**
An error occurs.
image

alexweininger:
This added treeUtil is for gaining access to the SubscriptionTreeItem of any node in the tree.

Fixes #1602

### DIFF
--- a/src/commands/openInPortal.ts
+++ b/src/commands/openInPortal.ts
@@ -4,30 +4,31 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { DeploymentsTreeItem } from "vscode-azureappservice";
-import { IActionContext, openInPortal as uiOpenInPortal } from "vscode-azureextensionui";
+import { AzExtParentTreeItem, IActionContext, ISubscriptionContext, openInPortal as uiOpenInPortal } from "vscode-azureextensionui";
 import { DeploymentSlotsTreeItem } from "../explorer/DeploymentSlotsTreeItem";
-import { SiteTreeItem } from "../explorer/SiteTreeItem";
 import { WebAppTreeItem } from "../explorer/WebAppTreeItem";
 import { ext } from "../extensionVariables";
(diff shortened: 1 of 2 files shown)


alexweininger:
> Let's hold off on this one for now, but I was imaginging more having the util function, and then adding the root property to the tree items by doing something like
> 
> ```
> get root(): ISubscriptionContext {
>   return findSubscriptionTreeItem(this)
> ]
> ```
> 
> Or something like that. That way you wouldn't have to make changes across the entire repo to remove .root references

Would we add this to `AzExtTreeItem`? Or would we add it to each tree item?

EricJizbaMSFT:
> That way you wouldn't have to make changes across the entire repo to remove .root references

@nturinski I'm a little confused on this because he's only changed two files?

I was actually thinking you would modify the `openInPortal` function in the ui package. Very similar to what you're doing here, but it would help more people. It would allow passing in `types.ISubscriptionContext | types.AzExtTreeItem` instead of just `types.ISubscriptionContext`.

alexweininger:
@EricJizbaMSFT 

What about when we need to access the root before calling `uiOpenInPortal`?

https://github.com/microsoft/vscode-azureappservice/blob/b9add4e9929882467558203041d13db1344065f1/src/commands/openInPortal.ts#L27

EricJizbaMSFT:
I believe `node.parent.fullId` would be the same as `node.root.client.id` in that case

alexweininger:
Closing in favor of https://github.com/microsoft/vscode-azuretools/pull/752

PR #1614, opened 2020-06-23
Transfer trial apps to an Azure subscription command

alexweininger:
This command is to let users easily "transfer" their trial app to an Azure subscription. It prompts them to create a web app then it initiates a deploy which automatically targets their newly created web app.

We will also show this command as the only child of an expired trial app. (upcoming PR)


nturinski:
So today it'll ask the user what workspace they want to deploy, but I feel like the merit of this transfer command is that it takes what's currently on the trial app and deploys that to the new app.  I'm not sure what the value add of this is in the current implementation besides just exposure.

alexweininger:
> So today it'll ask the user what workspace they want to deploy, but I feel like the merit of this transfer command is that it takes what's currently on the trial app and deploys that to the new app. I'm not sure what the value add of this is in the current implementation besides just exposure.

I've updated the PR (see edited description).

src/commands/deploy/deploy.ts
```diff
@@ -54,7 +54,10 @@ export async function deploy(context: IActionContext, target?: vscode.Uri | Site
     const { node, isNewWebApp }: IDeployNode = await getDeployNode(deployContext, target, isTargetNewWebApp);
 
     if (node instanceof TrialAppTreeItem) {
-        return await deployTrialApp(deployContext, node);
+        await deployTrialApp(deployContext, node);
```
nturinski, line 57: Did you verify this works?  This seems like it's in the wrong order to me.  I figured deployTrialApp would be called last.

alexweininger: It works, but I'm now realizing that I should call deployTrialApp last so that it actually works the first time they deploy.

src/commands/deploy/getDeployNode.ts
```diff
@@ -41,7 +41,8 @@ export async function getDeployNode(context: IDeployContext, target: Uri | strin
         const newNodes: SiteTreeItem[] = [];
         const disposable: Disposable = ext.tree.onTreeItemCreate((newNode: SiteTreeItem) => { newNodes.push(newNode); });
         try {
-            node = await ext.tree.showTreeItemPicker<SiteTreeItem>([WebAppTreeItem.contextValue, TrialAppTreeItem.contextValue], context);
+            const contextValues: string[] = isTargetNewWebApp ? [WebAppTreeItem.contextValue] : [WebAppTreeItem.contextValue, TrialAppTreeItem.contextValue];
```
nturinski, line 44: Can you explain the reason behind this check?

alexweininger: I should have made a comment, but I think at the time of writing that check it made complete sense 😅 (as always). But I think this check is now obsolete since we are calling `createWebApp` before calling `deploy` in `transferTrialApp`.

I will remove it.

PR #1615, opened 2020-06-23
Add deployments to trial app children

alexweininger:
Add deployments node to trial app children, and ensure that all context menu items are supported for trial apps. Mainly `Open in Portal` must be removed from trial app children.

So the context menu for deployments and deployment items under trial apps now look like:

`DeploymentsTreeItem`:
image

`DeploymentTreeItem`:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -655,17 +655,17 @@
                 },
                 {
                     "command": "appService.Refresh",
-                    "when": "view == azureAppService && viewItem =~ /^deployments(C|Unc)onnected(|NoPortal)$/",
+                    "when": "view == azureAppService && viewItem =~ /^deployments(C|Unc)onnected$/",
                     "group": "2@1"
                 },
                 {
                     "command": "appService.viewDeploymentLogs",
-                    "when": "view == azureAppService && viewItem =~ /deployment(|noportal)//",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -655,17 +655,17 @@
                 },
                 {
                     "command": "appService.Refresh",
-                    "when": "view == azureAppService && viewItem =~ /^deployments(C|Unc)onnected$/",
+                    "when": "view == azureAppService && viewItem =~ /^deployments(C|Unc)onnected(|NoPortal)$/",
```
EricJizbaMSFT, line 658: Curious if Nathan agrees, but the "NoPortal" thing in the context value feels a bit overly complex to me. I think I'd rather let it be shown on trial apps and simply throw an error in that case

alexweininger: Another alternative: Trial apps don't have the deployments node

nturinski: I think I agree with Eric.  Does redeploy work for trial apps?  I think the logs could be useful though.

alexweininger: Redeploy works for trial apps.

EricJizbaMSFT: Yeah I feel like it's worth it for the deployment logs

nturinski: Let's just throw an error then

PR #1616, opened 2020-06-24
Add create trial app node

alexweininger:
Shows up when a user doesn't have a trial app AND is not signed in.

Looks like this:

image

### DIFF
--- a/src/explorer/AzureAccountTreeItem.ts
+++ b/src/explorer/AzureAccountTreeItem.ts
@@ -32,7 +32,7 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
         } else {
             // Must check children[0] instanceof GenericTreeItem because if Azure account is still loading then isLoggedIn is false.
             if (!this.isLoggedIn && children[0] instanceof GenericTreeItem) {
-                children.push(new GenericTreeItem(this, { label: 'Create free trial app...', contextValue: 'createTrialApp', commandId: `${ext.prefix}.CreateTrialApp`, iconPath: getIconPath('WebApp') }));
+                children.push(new GenericTreeItem(this, { label: 'Create free trial app...', contextValue: 'azureCommand', commandId: `${ext.prefix}.CreateTrialApp`, iconPath: getIconPath('WebApp'), includeInTreeItemPicker: true }));
             }
         }
         return children;
--- a/src/explorer/AzureAccountTreeItem.ts
+++ b/src/explorer/AzureAccountTreeItem.ts
@@ -31,8 +31,8 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
             children.push(ti);
         } else {
             // Must check children[0] instanceof GenericTreeItem because if Azure account is still loading then isLoggedIn is false.
-            if (!this.isLoggedIn && children[0] instanceof GenericTreeItem) {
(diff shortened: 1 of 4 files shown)


src/explorer/AzureAccountTreeItem.ts
```diff
@@ -25,8 +26,14 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
     public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
         const ti: AzExtTreeItem | undefined = this.trialAppNode ?? await this.loadTrialAppNode();
         const children: AzExtTreeItem[] = await super.loadMoreChildrenImpl(clearCache, context);
+
         if (ti) {
             children.push(ti);
+        } else {
+            // Must check children[0] instanceof GenericTreeItem because if Azure account is still loading then isLoggedIn is false.
+            if (!this.isLoggedIn && children[0] instanceof GenericTreeItem) {
```
nturinski, line 34: I think you can move this logic into `loadTrialAppNode`  I think you could for this inside the  `(!loginSession)` condition.  I'm not sure how necessary the `children[0]` check is.  Did you run into where it would say "Loading..." and create new trial app?

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -25,8 +26,14 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
+                children.push(new GenericTreeItem(this, { label: 'Create free trial app...', contextValue: 'azureCommand', commandId: `${ext.prefix}.CreateTrialApp`, iconPath: getIconPath('WebApp'), includeInTreeItemPicker: true }));
```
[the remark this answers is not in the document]
nturinski: Including it in the treePicker has some weird behaviors.  I think because it doesn't return anything, if I click on it from the command palette, it opens up the page to create, but doesn't close the quickpick or anything.

src/explorer/AzureAccountTreeItem.ts
alexweininger: Ok after some investigating:
If I move the logic into `loadTrialAppNode` then I have to check `children[0]` (why I moved it out of `loadTrialAppNode` in the first place. **But** now I have found that if I leave the logic out of `loadTrialAppNode` then I don't have to check `children[0]`.

I moved it out since we already had access to `children` outside `loadTrialAppNode`.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -25,8 +26,14 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
     public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
         const ti: AzExtTreeItem | undefined = this.trialAppNode ?? await this.loadTrialAppNode();
         const children: AzExtTreeItem[] = await super.loadMoreChildrenImpl(clearCache, context);
+
         if (ti) {
             children.push(ti);
+        } else {
+            // Must check children[0] instanceof GenericTreeItem because if Azure account is still loading then isLoggedIn is false.
```
nturinski, line 33: Should remove this comment though

alexweininger: 🤦‍♂️😆

PR #1618, opened 2020-06-24
Hide certain trial app settings

alexweininger:
Hide most of the app settings that come one trial apps that aren't useful to trial app users.

```
        'SCM_BUILD_ARGS',
        'SCM_COMMAND_IDLE_TIMEOUT',
        'SCM_LOGSTREAM_TIMEOUT',
        'SCM_TRACE_LEVEL',
        'SCM_USE_LIBGIT2SHARP_REPOSITORY',
        'SITE_BASH_GIT_URL',
        'SITE_GIT_URL',
        'SITE_SITEKEY',
        'WEBSITE_AUTH_ENABLED',
        'WEBSITE_NODE_DEFAULT_VERSION',
        'WEBSITE_SITE_NAME'
```

### DIFF
--- a/src/explorer/trialApp/TrialAppTreeItem.ts
+++ b/src/explorer/trialApp/TrialAppTreeItem.ts
@@ -14,7 +14,7 @@ import { SiteTreeItemBase } from '../SiteTreeItemBase';
 import { ITrialAppMetadata } from './ITrialAppMetadata';
 import { TrialAppClient } from './TrialAppClient';
 
-const filteredSettings: string[] = [
+const settingsToHide: string[] = [
     'SCM_BUILD_ARGS',
     'SCM_COMMAND_IDLE_TIMEOUT',
     'SCM_LOGSTREAM_TIMEOUT',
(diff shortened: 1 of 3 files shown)


src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -16,6 +16,19 @@ import { TrialAppClient } from './TrialAppClient';
 
 export class TrialAppTreeItem extends SiteTreeItemBase implements ISiteTreeItem {
     public static contextValue: string = 'trialApp';
+    private static _settingsToHide: string[] = [
```
alexweininger, line 19: I wasn't sure where to put this. Is this good or should I put this somewhere more useful?

nturinski: I think it's fine in here, but I'm not sure if I'd attach it to the tree.  Maybe just as a constant in the file.  Also, I'd call it `filteredSettings`.  No real reason why, just feels like filter is the appropriate verb here to me.

EricJizbaMSFT: I like "settingsToHide" better. "filteredSettings" means the opposite to me, aka that would the the only ones we _show_, not _hide_. For example, that's how "azure.resourceFilter" works

alexweininger: I agree with Eric, I like "settingsToHide" because it removes the ambiguity of "filter". I'm going to change it to "settingsToHide" here and in the other PR.

PR #1620, opened 2020-06-25
Trial app Didact tutorial

alexweininger:
* Update the VS Code enginer and VS Code types to `1.42.0`. I had to do this to access the `ThemeIcon` constructor [here](https://github.com/microsoft/vscode-azureappservice/compare/alex/didact?expand=1#diff-8f8e73048154aea9f1c0b5f48d1aeffbR36).
* The trial app tutorial is a [didact](https://github.com/redhat-developer/vscode-didact) file which is invoked [here](https://github.com/microsoft/vscode-azureappservice/compare/alex/didact?expand=1#diff-9ea984351018242cbecfabb383efed0eR48).
* If a user does not have the Didact extension installed, I use a timer to wait until they install the extension just like we do in [installCosmosDBExtension.ts](https://github.com/microsoft/vscode-azureappservice/blob/f6591d861d0aeb2471a2bac1ee3898b33dc051d2/src/commands/installCosmosDBExtension.ts)

Note: The "Transfer your app to an Azure subscription" will be added after #1614 is merged.


resources/TrialApp.didact.md
```diff
@@ -0,0 +1,55 @@
...
+<img src="./trialAppTutorial/rightClickCloneScreenshot.png" style="width: 256px;">
+
+4. Choose a folder to clone the Trial App source code to.
+5. When the repository has finished cloning a notification like the one below will appear. Open the folder in VS Code or add it to the workspace.
+
+<img src="./trialAppTutorial/cloneNotificationScreenshot.png" style="width: 256px;">
+
+### Step 2: Make some changes to your app
+
+Make some changes to your trial app. Try changing some text in `template.html`.
```
nturinski, line 25: Might be overboard, but maybe we can add a screenshot of one of your changes?  I'm not sure what level of experience we're expecting from the users...

resources/TrialApp.didact.md
```diff
@@ -0,0 +1,55 @@
...
+
+[Click here](didact://?commandId=workbench.view.explorer) to view the project files in the Explorer.
+
+### Step 3: Deploy your changes
+
+1. Select the blue up arrow icon to deploy your code to your Trial App.
```
nturinski, line 31: Personally, I think we should have the deploy scenario be one-click.  We can do that by leveraging the same settings that we use to set up one-click deploy for Azure users.

resources/TrialApp.didact.md
```diff
@@ -0,0 +1,55 @@
...
+
+<img src="./trialAppTutorial/selectTrialAppDeployScreenshot.png" style="width: 512px;">
+
+You will see the following notification once your deployment is complete.
+
+<img src="./trialAppTutorial/deployCompleteScreenshot.png" style="width: 512px;">
+
+### Step 5: View your changes live
+
+Now that you've deployed changes to your Trial App. You can browse your website to view the changes live.
```
nturinski, line 49: I think you should direct users to do this in the step above when the notification pops up.

resources/TrialApp.didact.md
```diff
@@ -0,0 +1,55 @@
...
+
+<img src="./trialAppTutorial/browseWebsiteRedScreenshot.png" style="width: 256px;">
+
+---
```
nturinski, line 53: I assume we'll add the transferring stuff into the tutorial when that's actually in the repo?

nturinski:
I'm not really an expert on the tutorial stuff myself either, so I'm hoping @fiveisprime will spruce it up with some PM magic for us

resources/TrialApp.didact.md
alexweininger: Ya
> Note: The "Transfer your app to an Azure subscription" will be added after #1614 is merged.

src/commands/trialApp/showTutorial.ts
```diff
@@ -0,0 +1,49 @@
...
+        });
+    }
+
+    if (extensions.getExtension(extensionId)) {
+        showDidact();
+    } else {
+        const commandToRun: string = 'extension.open';
+        commands.executeCommand(commandToRun, extensionId);
+
+        while (Date.now() < maxTime) {
```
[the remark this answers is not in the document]
alexweininger: > I'm not really an expert on the tutorial stuff myself either, so I'm hoping @fiveisprime will spruce it up with some PM magic for us

Yeah I have some changes and polished to make with the tutorial. But I wanted to get this PR out Asap in case there were any code concerns.

alexweininger:
> I'm not really an expert on the tutorial stuff myself either, so I'm hoping @fiveisprime will spruce it up with some PM magic for us

Yeah I expect to do a lot of polishing to the tutorial. I wanted to get this PR out asap in case there were any code concerns.

nturinski:
To be honest, I'm not sure if it's worth updating our engines for the book icon.  I might be wrong here, but even though users _should_ update VS Code, we have ~6,000 users who are between 1.40 and 1.42.  That being said, since the icon doesn't add much (and we can just download our own icon), not sure we should even risk having some fall-off because they have to update VS Code.

🤷‍♀️ But I dunno, I'd be curious to know what others think about this.

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -95,6 +98,13 @@ export class TrialAppTreeItem extends SiteTreeItemBase implements ISiteTreeItem
     public isAncestorOfImpl?(contextValue: string | RegExp): boolean {
         return contextValue === TrialAppTreeItem.contextValue;
     }
+
+    public compareChildrenImpl(item1: AzExtTreeItem, item2: AzExtTreeItem): number {
```
nturinski, line 102: I'd feel more comfortable if you checked item1 as well.  The `super.compareChildrenImpl` sorts by alphabetical order, so if item1 is the `GenericTreeItem`, it'll most likely be sorted underneath whatever other node you have since it starts with 's'.

alexweininger:
Regarding updating the VS Code engine, I didn't know the side effects of updating. And now I really really don't think it's worth possible loosing users just for an icon. I removed the icon for now, and downgraded back to the 1.40.0 engine.

resources/TrialApp.didact.md
alexweininger: I added examples instead. So the user could copy and paste if they want.

resources/TrialApp.didact.md
fiveisprime: Agreed. I like the idea of it being consistent, but if you can make it a one-click experience, go for it.

alexweininger: It's now a one click experience for our happy path users 😄

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -95,6 +98,16 @@ export class TrialAppTreeItem extends SiteTreeItemBase implements ISiteTreeItem
...
+        if (item1 instanceof GenericTreeItem) {
+            return -1;
+        }
+        if (item2 instanceof GenericTreeItem) {
+            return 1; // tutorial node at top
```
nturinski, line 107: nit: Can you move the comment up above line 103 so it's more obvious both checks are for the tutorial node?

src/utils/installExtension.ts
```diff
@@ -0,0 +1,29 @@
...
+import { commands, extensions } from 'vscode';
+import { delay } from '../utils/delay';
+
+/**
+ * Returns `true` as soon as the extension is installed, or if the the extension is already installed. Returns `false` if the
+ * extension isn't installed within `timeoutInSeconds` seconds.
+ * @param extensionId id of the extension to install
+ * @param timeoutInSeconds Maximum time to wait for the user to install the extension. Defaults to 60 seconds.
+ */
+export async function installExtension(extensionId: string, timeoutInSeconds: number = 60): Promise<boolean> {
```
nturinski, line 15: Beautiful 😭

PR #1622, opened 2020-06-26
saveDeployDefaults before deploying Trial app

alexweininger:
Extracted these changes out of #1614 

This makes it so when deploying a trial app, the user doesn't have to pick the trial app node as the deploy target.

### DIFF
--- a/src/commands/deploy/getDeployNode.ts
+++ b/src/commands/deploy/getDeployNode.ts
@@ -26,7 +26,7 @@ export async function getDeployNode(context: IDeployContext, target: Uri | strin
     let isNewWebApp: boolean = isTargetNewWebApp;
     if (target instanceof SiteTreeItemBase) {
         node = target;
-    } else if (defaultWebAppId && defaultWebAppId !== none && !(ext.azureAccountTreeItem.trialAppNode && ext.azureAccountTreeItem.isLoggedIn)) {
+    } else if (defaultWebAppId && defaultWebAppId !== none) {
         node = await ext.tree.findTreeItem(defaultWebAppId, context); // resolves to undefined if app can't be found
         if (node) {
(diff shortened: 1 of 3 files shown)


alexweininger:
I'm wondering if we should also check if they are logged in before we call `saveDeployDefaults`. If they are logged in we might not want to save the trial app as default. Or maybe if they are logged in we prompt like we normally do.

nturinski:
I agree with that.  We should probably only save the defaultNodeId if they're not logged in.  Otherwise when they deploy, they'll be asked if they want to deploy to this app, which is the normal user experience.

P.S. Good initiative extracting it out into a function :)

src/commands/deploy/getDeployNode.ts
```diff
@@ -26,7 +26,7 @@ export async function getDeployNode(context: IDeployContext, target: Uri | strin
     let isNewWebApp: boolean = isTargetNewWebApp;
     if (target instanceof SiteTreeItemBase) {
         node = target;
-    } else if (defaultWebAppId && defaultWebAppId !== none) {
+    } else if (defaultWebAppId && defaultWebAppId !== none && !(ext.azureAccountTreeItem.trialAppNode && ext.azureAccountTreeItem.isLoggedIn)) {
```
nturinski, line 29: I don't think this logic is quite right.  In the case that a user has a a default web app and a trial app, we should be using their setting.

I think I know why you're doing this and I think we can mitigate that by deleting the defaultWebAppId setting when we transfer the subscription IF it's the trial app.

alexweininger: Ok, so to delete their setting we need to know their workspace. And the one place during transfer that we know their workspace is after we call getDeployFsPath within deploy. So somehow I need to know if they are transferring a trial app from within the deploy function.

PR #1625, opened 2020-06-27
Don't open extension if already installed

alexweininger:


src/utils/installExtension.ts
```diff
@@ -23,6 +21,8 @@ export async function installExtension(extensionId: string, timeoutInSeconds: nu
         if (extensions.getExtension(extensionId)) {
             return true;
         }
+        const commandToRun: string = 'extension.open';
+        commands.executeCommand(commandToRun, extensionId);
```
EricJizbaMSFT, line 25: Doesn't this re-open the page every 5 seconds? That sounds annoying...

nturinski: Yeah, I going to guess this isn't intentional, but this would re-open the page until they installed it every 5 seconds for a minute...

alexweininger: Yeah that wasn't intentional... I'm fixing it so that it only opens the page once.

PR #1630, opened 2020-06-29
Expired trial apps

alexweininger:
When a Trial app is expired the only child is a "Transfer to Subscription..." node.

image

And the only actions in the right click context menu are:
image

### DIFF
--- a/src/explorer/trialApp/TrialAppTreeItem.ts
+++ b/src/explorer/trialApp/TrialAppTreeItem.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AppSettingsTreeItem, DeploymentsTreeItem, LogFilesTreeItem, SiteFilesTreeItem } from 'vscode-azureappservice';
+import { ext } from 'vscode-azureappservice/out/src/extensionVariables';
 import { AzExtTreeItem, GenericTreeItem, IActionContext } from 'vscode-azureextensionui';
 import { localize } from '../../localize';
 import { openUrl } from '../../utils/openUrl';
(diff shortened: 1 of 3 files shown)


src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -96,6 +99,9 @@ export class TrialAppTreeItem extends SiteTreeItemBase implements ISiteTreeItem
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtTreeItem[]> {
+        if (isNaN(this.minutesLeft)) {
```
nturinski, line 102: I think we should just have a property on the trialAppTreeItem that is a boolean called `isExpired`.  You'll have to update it in refres, but then you don't have to have that isNaN check everywhere.

alexweininger: I added it to `TrialAppClient`. I think that's the best place for it. I need to wait about 30 minutes to test it...

nturinski, line 102: Missed this one

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AppSettingsTreeItem, DeploymentsTreeItem, LogFilesTreeItem, SiteFilesTreeItem } from 'vscode-azureappservice';
+import { ext } from 'vscode-azureappservice/out/src/extensionVariables';
```
nturinski, line 7: Wrong ext was auto imported here.

src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -96,6 +99,9 @@ export class TrialAppTreeItem extends SiteTreeItemBase implements ISiteTreeItem
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, _context: IActionContext): Promise<AzExtTreeItem[]> {
+        if (this.client.isExpired) {
+            return [new GenericTreeItem(this, { label: 'Transfer to Subscription...', commandId: `${ext.prefix}.TransferToSubscription`, contextValue: 'transferToSubscription' })];
```
nturinski, line 103: Could you localize the label string please? 🙏

PR #1631, opened 2020-06-29
Delay for 15 seconds when deploying trial app

alexweininger:
I'm not happy that delaying 5 seconds or 10 seconds wasn't enough. But I really don't want users to hit "Browse" and not see the changes they just deployed.

### DIFF
--- a/src/commands/deploy/deployTrialApp.ts
+++ b/src/commands/deploy/deployTrialApp.ts
@@ -18,7 +18,7 @@ export async function deployTrialApp(deployContext: IDeployContext, node: TrialA
         const title: string = localize('deploying', 'Deploying to "{0}"... Check [output window](command:{1}) for status.', node.client.fullName, `${ext.prefix}.showOutputChannel`);
         return await window.withProgress({ location: ProgressLocation.Notification, title }, async () => {
             await localGitDeploy(node.client, { fsPath: deployContext.workspace.uri.fsPath, branch: 'RELEASE', commit: commit }, deployContext);
-            // Delay for 15 seconds to make sure changes are reflected immediatley after deploy 'finishes'
+            // Delay for 15 seconds to make sure changes are reflected immediately after deploy 'finishes'
             await delay(15000);
         });
     });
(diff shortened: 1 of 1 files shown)


nturinski:
How did you choose 15 seconds, by the way?

alexweininger:
> How did you choose 15 seconds, by the way?

@nturinski 

I tested it at 5 and 10 and the changes we're still not showing up every time. And at 15 seconds they showed up every time.

EricJizbaMSFT:
> Namely, this delay needs to happen with every deploy, not just the first.

This is pretty concerning. Why is that the case?

alexweininger:
> This is pretty concerning. Why is that the case?

@EricJizbaMSFT 

This issue only happens with Trial apps. When we deploy a Trial app with `localGitDeploy`, `tryGetLatestDeployment` within `waitForDeploymentToComplete` shows that the deployment has completed long before the changes are reflected on the Trial app site. This is not the case for real apps using `localGitDeploy`.

EricJizbaMSFT:
Yeah but why is that the case? We never liked having a 10 second delay for normal apps and I'm pretty sure we had a decent amount of discussion with the platform team before giving up and adding the delay in a _small number of cases_. A 15 second delay for every single deploy is pretty bad

alexweininger:
I agree, a 15 second delay for every deploy is terrible. I haven't been able to come to any conclusions from my own investigating. And I'm not sure what direction to take in terms of what to investigate next. From our end Kudu just says the deployment has completed. The deployment logs match up with when we receive the deployment completed notification.

So to me, it seems that Kudu is reporting that the deployment is complete way before the changes are reflected on the Trial app site. 

I'm not sure what (if anything) we can do to investigate this further, other than to ask the owners of TryAppService.

nturinski:
> A 15 second delay for every single deploy is pretty bad

I agree it's pretty bad, but I suspect that most people using the trial app will only deploy with this method once.  And even if they do multiple deploys, according to Alex, the time ends up being about the same, it's just that Kudu reports the deployment is complete much sooner than it normally does for typical web app deployment, so I think the experience ends up being roughly the same.

We can investigate it further, but considering it's likely a platform issue for an experience that a user should only have to do once, I think it's more valuable for Alex to finish up the experience and if he has time, he can come back and investigate this further.

EricJizbaMSFT:
Why would people only deploy once? Feels like the most successful scenario for us is they deploy multiple times, leverage the full free hour, realize how great it is, and then convert to a real user. If they only deploy once, I doubt they would convert to a real user.

Deploying is the most important scenario. I think it's totally worth it to prioritize that scenario instead of other advanced features that new users probably won't know/care about yet. I would at least reach out to the TryAppService team. You can also test with local git deploy on a normal app and see if it has similar issues.

PR #1635, opened 2020-06-30
Add trialApp property to telemetry on deploy

alexweininger:


src/commands/deploy/deployTrialApp.ts
```diff
@@ -12,6 +12,7 @@ import { IDeployContext } from './IDeployContext';
 import { showDeployCompletedMessage } from './showDeployCompletedMessage';
 
 export async function deployTrialApp(deployContext: IDeployContext, node: TrialAppTreeItem): Promise<void> {
+    deployContext.telemetry.properties.trialApp = 'true';
```
nturinski, line 15: So `deployContext` isn't the context that is registered for the deploy event telemetery.  This is why we pass both contexts to `confirmDeploymentPrompt` like so https://github.com/microsoft/vscode-azureappservice/blob/master/src/commands/deploy/deploy.ts#L94

nturinski: Never mind, I'm wrong.  I've been informed that because deployContext copies context, it'll end up copying the references to telemetry and errorHandling.  So you can set the telemetry on deployContext and it'll affect the original telemetry object 👍

PR #1636, opened 2020-06-30
Throw error on Trial app editScmType

alexweininger:
Trial apps do not support changing the ScmType.

### DIFF
--- a/src/commands/deployments/editScmType.ts
+++ b/src/commands/deployments/editScmType.ts
@@ -7,18 +7,18 @@ import * as appservice from "vscode-azureappservice";
 import { DeploymentsTreeItem } from "vscode-azureappservice";
 import { IActionContext } from "vscode-azureextensionui";
 import { ScmType } from "../../constants";
+import { OperationNotSupportedError } from '../../errors';
 import { SiteTreeItem } from "../../explorer/SiteTreeItem";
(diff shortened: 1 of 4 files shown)


src/commands/deployments/editScmType.ts
```diff
@@ -8,13 +8,18 @@ import { DeploymentsTreeItem } from "vscode-azureappservice";
...
 import { WebAppTreeItem } from "../../explorer/WebAppTreeItem";
 import { ext } from "../../extensionVariables";
+import { localize } from '../../localize';
 
 export async function editScmType(context: IActionContext, node?: SiteTreeItem | DeploymentsTreeItem, newScmType?: ScmType, showToast?: boolean): Promise<void> {
     if (!node) {
         node = <SiteTreeItem>await ext.tree.showTreeItemPicker(WebAppTreeItem.contextValue, context);
     } else if (node instanceof DeploymentsTreeItem) {
+        if (node.parent instanceof TrialAppTreeItem) {
+            throw Error(localize('notSupported', 'This operation is not supported.'));
```
nturinski, line 21: Think it makes sense to just add this string to the constants since it gets used three times in the repo.

EricJizbaMSFT: Would suggest following the "errors.ts" pattern instead of using "constants.ts". For example: https://github.com/microsoft/vscode-azuretools/blob/master/ui/src/errors.ts

PR #1637, opened 2020-06-30
Add telemetry parameter to TryAppService url

alexweininger:
This parameter will help identify where traffic is coming from for the TryAppService site. This information is for the Try App Service team, not our own telemetry.

### DIFF
--- a/src/commands/trialApp/createTrialApp.ts
+++ b/src/commands/trialApp/createTrialApp.ts
@@ -7,5 +7,5 @@ import { IActionContext } from 'vscode-azureextensionui';
 import { openUrl } from '../../utils/openUrl';
 
 export async function createTrialApp(_context: IActionContext): Promise<void> {
-    await openUrl('https://code.visualstudio.com/tryappservice/');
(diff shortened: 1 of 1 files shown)


src/commands/trialApp/createTrialApp.ts
```diff
@@ -7,5 +7,5 @@ import { IActionContext } from 'vscode-azureextensionui';
 import { openUrl } from '../../utils/openUrl';
 
 export async function createTrialApp(_context: IActionContext): Promise<void> {
-    await openUrl('https://code.visualstudio.com/tryappservice/');
+    await openUrl('https://code.visualstudio.com/tryappservice/?utm_source=appservice-extension');
```
EricJizbaMSFT, line 10: We usually use the name in package.json to identify ourselves, aka "vscode-azureappservice"

alexweininger: Matt said to use the same value that we use [here](https://github.com/microsoft/vscode-azureappservice/blame/master/README.md#L21) in the README.

nturinski: Makes sense to just change the README unless there's a specific reason we have to use that string.

PR #1638, opened 2020-06-30
Prep for 0.18.0

alexweininger:
Pending with @fiveisprime on any changes he might want to make to the README.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.18.0 - 2020-07-10
 ### Added
-- Added support for Free App Service Trial Apps! Visit [here](https://code.visualstudio.com/tryappservice/?utm_source=appservice-extension) to Create and Import your own Free NodeJS Trial App.
+- Added support for Free App Service Trial Apps! Visit [here](https://code.visualstudio.com/tryappservice/?utm_source=appservice-extension) to Create and Import your own Free Node.js Trial App.
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log
 
+## 0.18.0 - 2020-07-10
+### Added
+- Added support for Free App Service Trial Apps! Visit [here](https://code.visualstudio.com/tryappservice/) to Create and Import your own Free NodeJS Trial App.
```
nturinski, line 5: I assume Matt would want the telemetry endpoint thing here as well

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azureappservice",
     "displayName": "Azure App Service",
     "description": "An Azure App Service management extension for Visual Studio Code.",
-    "version": "0.17.1-alpha",
+    "version": "0.18.0",
```
nturinski, line 5: Hold off on bumping the version.  Since we'll be testing, we don't our telemetry to be populated with this version number until we actually release so we can track any errors that customers might have once we release

EricJizbaMSFT: You also want to hold off on the NOTICE.html because our dependencies change pretty often and it needs to reflect the dependencies we actually release with.

Honestly the "Prep for release" PR is pretty trivial and we usually do it a few days before release, not a week and a half before release

CHANGELOG.md
fiveisprime: Yep, it's added! Also, please change "NodeJS" to "Node.js" and I'm good.

PR #1648, opened 2020-07-01
Better handle expired Trial apps

### LINKED ISSUE #1645: The state of Trial App will change to "Invalid" after it expired about four mins
**OS:** Win10
**Build Version:** 20200701.1

**Repro Steps:**
1. Create a new Trial App and wait for it to expire.
2. Refresh the Trial App after several mins.

**Expect:**
The state of the Trial App has always been expired.

**Actual:**
The state of Trial App from Expired change to Invalid.
image

alexweininger:
Some edge cases that I had to handle here:

1. Importing an expired trial app
    a. If we are still able to get the trial app metadata (the timeLeft property is undefined in this case).
    b. If the request to get metadata returns an error.
2. Reloading VS Code with an expired trial app. This is why I had to add additional information to the context.globalState

Fixes #1645

### DIFF
--- a/src/commands/trialApp/importTrialApp.ts
+++ b/src/commands/trialApp/importTrialApp.ts
@@ -5,7 +5,8 @@
 
 import { commands, ProgressLocation, window } from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
-import { TrialAppLoginSession } from '../../constants';
+import { TrialAppContext } from '../../constants';
+import { ITrialAppContext } from '../../explorer/trialApp/ITrialAppContext';
 import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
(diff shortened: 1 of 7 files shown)


src/commands/trialApp/importTrialApp.ts
```diff
@@ -15,7 +16,22 @@ export async function importTrialApp(_context: IActionContext, loginSession: str
     await window.withProgress({ location: ProgressLocation.Notification, cancellable: false }, async p => {
         p.report({ message: localize('importingTrialApp', 'Importing trial app...') });
         ext.azureAccountTreeItem.trialAppNode = await TrialAppTreeItem.createTrialAppTreeItem(ext.azureAccountTreeItem, loginSession);
-        ext.context.globalState.update(TrialAppLoginSession, loginSession);
+        const trialAppNode = ext.azureAccountTreeItem.trialAppNode;
+
+        if (trialAppNode.client.metadata.timeLeft === undefined) {
```
nturinski, line 21: Please leave a comment explaining why we need this check

src/commands/trialApp/importTrialApp.ts
```diff
@@ -15,7 +16,22 @@ export async function importTrialApp(_context: IActionContext, loginSession: str
...
+            throw new Error(localize('trialAppExpired', 'Trial app is expired. Could not import.'));
+        }
+
+        const expirationDate: Date = new Date();
```
nturinski, line 25: Use Date.now()  which returns the current time in ms.  Then just add the timeLeft * 1000

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -67,16 +69,19 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
             this.trialAppNode = undefined;
             return undefined;
         }
 
         const ti: AzExtTreeItem[] = await this.createTreeItemsWithErrorHandling(
-            [loginSession],
+            [trialAppContext.loginSession],
             'trialAppInvalid',
             async (source: string): Promise<AzExtTreeItem> => {
+                if (new Date(trialAppContext.expirationDate) < new Date()) {
```
nturinski, line 82: I'd use a ternary just so it's clear there's 2 different types of trees that we can return here

PR #1649, opened 2020-07-01
Add modal on removing trial app

### LINKED ISSUE #1641: No confirmation dialog pops up when removing Trial App
**OS:** Win10
**Build Version:** 20200701.1

**Repro Steps:**
1. Right click the Trial App.
2. Select "Remove" action.
3. Check the result.

**Expect:**
A confirmation dialog pops up, such as the dialog when deleting a web app.
image

**Actual:**
Remove the Trial App directly without confirmation dialog.

alexweininger:
Fixes #1641

### DIFF
--- a/src/commands/trialApp/removeTrialApp.ts
+++ b/src/commands/trialApp/removeTrialApp.ts
@@ -3,7 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { IActionContext } from 'vscode-azureextensionui';
+import { MessageItem, window } from 'vscode';
+import { DialogResponses, IActionContext } from 'vscode-azureextensionui';
 import { TrialAppLoginSession } from '../../constants';
 import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
 import { ext } from '../../extensionVariables';
@@ -13,7 +14,13 @@ export async function removeTrialApp(context: IActionContext, node?: TrialAppTre
         node = await ext.tree.showTreeItemPicker<TrialAppTreeItem>(TrialAppTreeItem.contextValue, context);
     }
 
-    ext.context.globalState.update(TrialAppLoginSession, undefined);
-    delete ext.azureAccountTreeItem.trialAppNode;
-    await ext.tree.refresh();
+    const message: string = `Are you sure you want to remove Trial app "${node.client.fullName}"?`;
+    const remove: MessageItem = { title: 'Remove' };
+    window.showInformationMessage(message, { modal: true }, remove, DialogResponses.cancel).then(async (response: MessageItem): Promise<void> => {
+        if (response === remove) {
+            ext.context.globalState.update(TrialAppLoginSession, undefined);
+            delete ext.azureAccountTreeItem.trialAppNode;
+            await ext.tree.refresh();
+        }
+    });
 }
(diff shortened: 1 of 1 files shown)


src/commands/trialApp/removeTrialApp.ts
```diff
@@ -13,7 +14,13 @@ export async function removeTrialApp(context: IActionContext, node?: TrialAppTre
         node = await ext.tree.showTreeItemPicker<TrialAppTreeItem>(TrialAppTreeItem.contextValue, context);
     }
 
-    ext.context.globalState.update(TrialAppLoginSession, undefined);
-    delete ext.azureAccountTreeItem.trialAppNode;
-    await ext.tree.refresh();
+    const message: string = `Are you sure you want to remove Trial app "${node.client.fullName}"?`;
+    const remove: MessageItem = { title: 'Remove' };
+    window.showInformationMessage(message, { modal: true }, remove, DialogResponses.cancel).then(async (response: MessageItem): Promise<void> => {
```
EricJizbaMSFT, line 19: We only use `.then` instead of `await` when it's not a core part of the command. For example, after deploying, we don't await the prompt to "view output"/"upload app settings"/etc. because it's not required and plenty of people ignore the notification. The reason it matters is because the telemetry doesn't get sent until the command finishes and we really want all deploy telemetry regardless of if they ignore that notification.

In this case, it's a modal dialog (aka they can't ignore it) and it's core to the command itself (aka we're fine waiting until they respond to send telemetry).

src/commands/trialApp/removeTrialApp.ts
```diff
@@ -13,7 +14,13 @@ export async function removeTrialApp(context: IActionContext, node?: TrialAppTre
...
+        if (response === remove) {
```
EricJizbaMSFT, line 20: You don't need to include `DialogResponses.cancel`. It's included by default these days

src/commands/trialApp/removeTrialApp.ts
```diff
@@ -13,7 +14,13 @@ export async function removeTrialApp(context: IActionContext, node?: TrialAppTre
         node = await ext.tree.showTreeItemPicker<TrialAppTreeItem>(TrialAppTreeItem.contextValue, context);
     }
 
-    ext.context.globalState.update(TrialAppLoginSession, undefined);
-    delete ext.azureAccountTreeItem.trialAppNode;
-    await ext.tree.refresh();
+    const message: string = `Are you sure you want to remove Trial app "${node.client.fullName}"?`;
+    const remove: MessageItem = { title: 'Remove' };
```
EricJizbaMSFT, line 18: I'd prefer `ext.ui.showWarningMessage` instead of `window.showInformationMessage`. It handles 'cancels' better and a warning feels more appropriate

src/commands/trialApp/removeTrialApp.ts
```diff
@@ -13,7 +14,13 @@ export async function removeTrialApp(context: IActionContext, node?: TrialAppTre
         node = await ext.tree.showTreeItemPicker<TrialAppTreeItem>(TrialAppTreeItem.contextValue, context);
     }
 
-    ext.context.globalState.update(TrialAppLoginSession, undefined);
-    delete ext.azureAccountTreeItem.trialAppNode;
-    await ext.tree.refresh();
+    const message: string = `Are you sure you want to remove Trial app "${node.client.fullName}"?`;
```
nturinski, line 17: nit: We should localize any new string going forward.  

Also, don't like the casing.  We should either do Trial App or trial app.  lol I just noticed you're pretty inconsistent in your tutorial about this but you have generally been using "trial apps" in messages.

src/commands/trialApp/removeTrialApp.ts
```diff
@@ -13,7 +14,13 @@ export async function removeTrialApp(context: IActionContext, node?: TrialAppTre
...
+    const remove: MessageItem = { title: 'Remove' };
+    const response: MessageItem | undefined = await ext.ui.showWarningMessage(message, { modal: true }, remove);
+
+    if (response === remove) {
```
nturinski, line 21: I think you technically don't have to this check here since if the user hits cancel, it'll throw a UserCancelledError  (since there is only one choice here, if it passes, we can assume the response we remove).

Means we don't need that const response either.

PR #1652, opened 2020-07-02
Add telemetry properties to trial app commands

alexweininger:
Examples of queries we can make from these properties:
- How many trial app users are deploying their own project vs. the template?
- How much time do users take after importing a trial app to deploy (if ever)?
- How many users are logged in when they deploy their trial app?

### DIFF
--- a/src/commands/logstream/startStreamingLogs.ts
+++ b/src/commands/logstream/startStreamingLogs.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as appservice from 'vscode-azureappservice';
-import { ISimplifiedSiteClient } from 'vscode-azureappservice';
 import { IActionContext } from 'vscode-azureextensionui';
 import { SiteTreeItem } from '../../explorer/SiteTreeItem';
 import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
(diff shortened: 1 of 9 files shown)


src/commands/logstream/startStreamingLogs.ts
```diff
@@ -25,6 +24,5 @@ export async function startStreamingLogs(context: IActionContext, node?: SiteTre
         }
     };
 
-    const client: ISimplifiedSiteClient = (node instanceof TrialAppTreeItem) ? node.client : node.root.client;
-    await appservice.startStreamingLogs(client, verifyLoggingEnabled, node.logStreamLabel);
+    await appservice.startStreamingLogs(node.client, verifyLoggingEnabled, node.logStreamLabel);
```
alexweininger, line 27: This change is not related to telemetry, just removing the extra line of code.

src/commands/trialApp/importTrialApp.ts
```diff
@@ -30,7 +30,7 @@ export async function importTrialApp(_context: IActionContext, loginSession: str
             expirationDate: expirationDate,
             loginSession: loginSession
         };
-
+        context.telemetry.properties.trialTimeRemaining = String(trialAppNode.metadata.timeLeft);
```
nturinski, line 33: I feel like you might as well just use the addTrialAppTelemetry here too even though it's obviously going to be a trialApp and they _should_ not be logged in

nturinski: Ah but whatever, it doesn't matter.  Could just be cluttering up the properties too

nturinski: And there's no node, never mind lol sorry for all the comments

PR #1664, opened 2020-07-06
Alex/fix remove trial app

### LINKED ISSUE #1660: Fail to remove an expired Trial App
**OS:** Win10/Mac
**Build Version:** 20200703.6

**Repro Steps:**
1. Right click the expired Trial App.
2. Select "Remove" action.
3. Check the result.

**Expect:**
Success to remove Trial App without any error.

**Actual:**
Fail to remove an expired Trial App.
image

alexweininger:
Fixes #1660

### DIFF
--- a/src/commands/trialApp/removeTrialApp.ts
+++ b/src/commands/trialApp/removeTrialApp.ts
@@ -7,7 +7,6 @@ import { MessageItem } from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
 import { TrialAppContext } from '../../constants';
 import { ExpiredTrialAppTreeItem } from '../../explorer/trialApp/ExpiredTrialAppTreeItem';
-import { ITrialAppContext } from '../../explorer/trialApp/ITrialAppContext';
 import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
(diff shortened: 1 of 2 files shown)


src/commands/trialApp/removeTrialApp.ts
```diff
@@ -6,21 +6,32 @@
...
 import { addTrialAppTelemetry } from './addTrialAppTelemetry';
 
-export async function removeTrialApp(context: IActionContext, node?: TrialAppTreeItem): Promise<void> {
+export async function removeTrialApp(context: IActionContext, node?: TrialAppTreeItem | ExpiredTrialAppTreeItem): Promise<void> {
     if (!node) {
         node = await ext.tree.showTreeItemPicker<TrialAppTreeItem>(TrialAppTreeItem.contextValue, context);
     }
 
-    const message: string = localize('removeTrialApp', 'Are you sure you want to remove trial app "{0}"?', node.client.fullName);
+    const trialAppContext: ITrialAppContext | undefined = ext.context.globalState.get(TrialAppContext);
```
nturinski, line 21: Rather than getting the context here, I think it makes sense to just use the label instead.  You have to have a context to create an ExpiredTrialApp anyway, so might as well remove this check.

nturinski: I mean changing `node.client.fullName` to just `node.label`.  I think that should be fine for this case since the label is the name.

alexweininger:
Closing in favor of changes in #1679

PR #1673, opened 2020-07-06
Make pictures in tutorial larger

### LINKED ISSUE #1665: Trial App screenshots very small
The text in the screenshot seems to be a lot smaller than the normal text in the markdown. Makes it hard to read, and zooming doesn't help very much

Screen Shot 2020-07-06 at 11 19 38 AM

alexweininger:
Fixes #1665


EricJizbaMSFT:
Well can I get a screenshot or something?

resources/TrialApp.didact.md
```diff
@@ -8,17 +8,17 @@ This tutorial will show you how to make and deploy changes to your Azure App Ser
 
 1. Locate your imported trial app in the Azure view.
 
-<img src="./trialAppTutorial/viewTrialAppScreenshot.png" style="width: 256px;">
+<img src="./trialAppTutorial/viewTrialAppScreenshot.png" style="width: 384px;">
```
EricJizbaMSFT, line 11: Usually we don't specify the width. Do you have to?

For example:
```
![Deploy button](resources/deploy-button.png)
```
https://raw.githubusercontent.com/microsoft/vscode-azureappservice/master/README.md

alexweininger:
image
image

resources/TrialApp.didact.md
alexweininger: To make the pictures larger and a consistent size I think I have to. If I don't then some are very very large and others are smaller etc.

nturinski:
Oh, I thought this https://github.com/microsoft/vscode-azureappservice/blob/6a5bd0389f8f98262d9bc6d22d60668fd354cc8e/resources/TrialApp.didact.md would be how it looked in VS Code, but it looks quite a bit different.  Does markdown not normally support the width property?

alexweininger:
The Didact view is different than the markdown view on GitHub. I made sure the text in the images is of similar size to the text in the tutorial.

resources/TrialApp.didact.md
EricJizbaMSFT: Ah - yeah I know what you're talking about. I think the trick is to change the size of the picture itself instead of specifying the style in the markdown. No need to ship a "very very large" picture in the vsix. Not sure how to do it on Windows, but on a Mac there's an "Adjust size" tool that makes it pretty easy

alexweininger: Hmm. Doing that on Windows and it seems to reduce the quality substantially. I think I'd rather let the HTML + CSS rendering do the resize and have the image look nice. But maybe I'm not resizing correctly.

image

EricJizbaMSFT: Yowza. Can you point me to the original image you just used and the exact pixel sizes you did? I can try

alexweininger: https://github.com/microsoft/vscode-azureappservice/blob/master/resources/trialAppTutorial/viewTrialAppScreenshot.png

I resized the width to be 256 px, while maintaining the aspect ratio. This is the same resize that I have in the tutorial.

EricJizbaMSFT:
Oh while you're messing with images - can you double check if there are any extra ones in "resources/trialAppTutorial"? Seems like there's more images than used in the markdown

PR #1677, opened 2020-07-06
Notification to explain installing Didact on showTutorial

### LINKED ISSUE #1668: [Suggestion] Add a notification explaining that Didact must be installed for tutorial to work
Clicking the "Show tutorial" tree item opened the Didact extension page directly. It may be good to throw up a notification explaining that it's required to continue. Felt a lil weird

alexweininger:
Fixes #1668 

image

### DIFF
--- a/src/commands/installCosmosDBExtension.ts
+++ b/src/commands/installCosmosDBExtension.ts
@@ -3,17 +3,20 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { extensions } from 'vscode';
 import { AzureTreeItem, IActionContext } from 'vscode-azureextensionui';
 import { CosmosDBTreeItem } from '../explorer/CosmosDBTreeItem';
 import { installExtension } from '../utils/installExtension';
 
 export async function installCosmosDBExtension(context: IActionContext, treeItem: AzureTreeItem): Promise<void> {
     const extensionId: string = 'ms-azuretools.vscode-cosmosdb';
-    if (await installExtension(extensionId)) {
-        if (treeItem.parent) {
-            await treeItem.parent.refresh();
-            if ((<CosmosDBTreeItem>treeItem.parent).cosmosDBExtension) {
-                context.telemetry.properties.installedCosmos = 'true';
+    if (!extensions.getExtension(extensionId)) {
+        if (await installExtension(extensionId)) {
+            if (treeItem.parent) {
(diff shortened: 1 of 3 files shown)


src/commands/trialApp/showTutorial.ts
```diff
@@ -15,6 +15,9 @@ export async function showTutorial(context: IActionContext): Promise<void> {
 
     if (trialAppNode) {
         const extensionId: string = 'redhat.vscode-didact';
+        if (!extensions.getExtension(extensionId)) {
+            window.showInformationMessage(localize('mustInstallDidact', 'The Didact extension must be installed to view the trial app tutorial.'));
```
nturinski, line 19: I prefer using this kind of error dialog:
https://github.com/microsoft/vscode-azurestaticwebapps/blob/master/src/commands/createHttpFunction.ts#L78-L80

src/commands/trialApp/showTutorial.ts
```diff
@@ -15,6 +15,9 @@ export async function showTutorial(context: IActionContext): Promise<void> {
 
     if (trialAppNode) {
         const extensionId: string = 'redhat.vscode-didact';
+        if (!extensions.getExtension(extensionId)) {
```
nturinski, line 18: Since we're checking for getExtension here now, I think we don't actually need the check in `installExtension`.  I think `installExtension` should only be called if we need to install anyway.  The getExtension check inside the while loop should stay, but the one before `executeCommand(extension.open)` can go.

src/commands/installCosmosDBExtension.ts
```diff
@@ -3,17 +3,20 @@
...
 import { installExtension } from '../utils/installExtension';
 
 export async function installCosmosDBExtension(context: IActionContext, treeItem: AzureTreeItem): Promise<void> {
     const extensionId: string = 'ms-azuretools.vscode-cosmosdb';
-    if (await installExtension(extensionId)) {
-        if (treeItem.parent) {
-            await treeItem.parent.refresh();
-            if ((<CosmosDBTreeItem>treeItem.parent).cosmosDBExtension) {
-                context.telemetry.properties.installedCosmos = 'true';
+    if (!extensions.getExtension(extensionId)) {
```
nturinski, line 13: You don't actually need to do this here since the treeItem doesn't appear unless the extension isn't installed: https://github.com/microsoft/vscode-azureappservice/blob/master/src/explorer/CosmosDBTreeItem.ts#L50

PR #1679, opened 2020-07-07
Add trial app commands to command palette

### LINKED ISSUE #1642: [Suggestion] Add the commands for Trial App in command palette
**OS:** Win10
**Build Version:** 20200701.1

**Repro Steps:**
1. Right click one Trial App and check the context menu.
image
2. F1 to check the command list.

**Actual:**
There is no command for Trial App.

**Suggestion:**
Add the commands for Trial App in command palette.

### LINKED ISSUE #1660: Fail to remove an expired Trial App

alexweininger:
- Commands only show up when a trial app is in the tree
- Clone command does not show up if the trial has expired

Fixes #1642, Fixes #1660

### DIFF
--- a/src/commands/trialApp/removeTrialApp.ts
+++ b/src/commands/trialApp/removeTrialApp.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { commands, MessageItem } from 'vscode';
+import { MessageItem } from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
 import { TrialAppContext } from '../../constants';
 import { ITrialAppContext } from '../../explorer/trialApp/ITrialAppContext';
@@ -30,7 +30,6 @@ export async function removeTrialApp(context: IActionContext, node?: TrialAppTre
         addTrialAppTelemetry(context, node);
     }
     ext.context.globalState.update(TrialAppContext, undefined);
-    await commands.executeCommand('setContext', 'hasTrialApp', false);
     delete ext.azureAccountTreeItem.trialAppNode;
     await ext.tree.refresh();
 }
--- a/src/commands/trialApp/removeTrialApp.ts
+++ b/src/commands/trialApp/removeTrialApp.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { MessageItem } from 'vscode';
+import { commands, MessageItem } from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
 import { TrialAppContext } from '../../constants';
 import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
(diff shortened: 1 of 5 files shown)


src/commands/trialApp/removeTrialApp.ts
```diff
@@ -3,26 +3,34 @@
...
+        node = ext.azureAccountTreeItem.trialAppNode;
     }
 
-    const message: string = localize('removeTrialApp', 'Are you sure you want to remove trial app "{0}"?', node.client.fullName);
+    const trialAppContext: ITrialAppContext | undefined = ext.context.globalState.get(TrialAppContext);
+    if (!trialAppContext) {
+        throw new Error(localize('noTrialContext', 'Cannot get trial app context.'));
+    }
+
+    const message: string = localize('removeTrialApp', 'Are you sure you want to remove trial app "{0}"?', trialAppContext.name);
```
nturinski, line 25: Could you go back and make those changes we discussed in the other PR?

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -77,13 +78,16 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
 
         if (trialAppContext.expirationDate < Date.now()) {
             this.trialAppNode = undefined;
+            await commands.executeCommand('setContext', 'trialAppExpired', true);
+            await commands.executeCommand('setContext', 'hasTrialApp', true);
```
nturinski, line 82: Are these contexts persistent? Like after you close out the editor?

alexweininger: No. However, I accounted for that by setting the values here in AzureAccountTreeItem.

src/commands/trialApp/removeTrialApp.ts
alexweininger: Since we are now able to call removeTrialApp from the command palette, we no longer have access to the trial app node. So I think the best way is to grab the name from here instead of using a tree item picker (since there is always only 1 trial app)

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -69,6 +70,17 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
         await this.trialAppNode?.refresh();
     }
 
+    private async setContext(treeItem: AzExtTreeItem | undefined): Promise<void> {
```
alexweininger, line 73: I know this is a bit different than what we discussed, so here are was my thought process:

1. All `setContext` calls are done in this function, to remove any confusion and mistakes when setting it in multiple places.
2. Since we are calling `setContext` inside of `loadMoreChildrenImpl`, the command palette will always (assuming the logic inside `setContext` is correct) be in sync with what is shown in the tree.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -69,6 +70,17 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
...
+        if (!treeItem) {
+            await commands.executeCommand('setContext', 'hasTrialApp', false);
+            return;
+        }
+        await commands.executeCommand('setContext', 'trialAppExpired', treeItem instanceof ExpiredTrialAppTreeItem);
+        if (treeItem instanceof TrialAppTreeItem || treeItem instanceof ExpiredTrialAppTreeItem) {
+            await commands.executeCommand('setContext', 'hasTrialApp', true);
+        }
+    }
+
```
alexweininger, line 83: I know this is a bit different than what we discussed, so here was my thought process:

1. All `setContext` calls are done in this function, to remove any confusion and mistakes when setting it in multiple places.
2. Since we are calling `setContext` inside of `loadMoreChildrenImpl`, the command palette will always (assuming the logic inside `setContext` is correct) be in sync with what is shown in the tree.

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -68,6 +70,17 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
         await this.trialAppNode?.refresh();
     }
 
+    private async setContext(treeItem: AzExtTreeItem | undefined): Promise<void> {
+        if (!treeItem) {
+            await commands.executeCommand('setContext', 'hasTrialApp', false);
+            return;
+        }
+        await commands.executeCommand('setContext', 'trialAppExpired', treeItem instanceof ExpiredTrialAppTreeItem);
+        if (treeItem instanceof TrialAppTreeItem || treeItem instanceof ExpiredTrialAppTreeItem) {
```
nturinski, line 79: Do you need this check?  I assume if there's no treeItem, then the treeItem has to be either the TrialApp or ExpiredTrial tree item types.

alexweininger: It could be an invalid tree item if there was an error.

PR #1688, opened 2020-07-09
Implement shouldWriteFile and getResourceName

### LINKED ISSUE #1610: There is no output log and notification when updating the file
**OS:** Mac
**Build Version:** 20200618.1
**.NET Core:** 2.1

**Repro Steps:**

1. Expand one web app --> Files --> Open one file.
2. Make some change for this file and save it.
3. Check the output log.

**Expect:**
There is output log and notification.

**Actual:**
There is no output log and notification.

alexweininger:
Relies on https://github.com/microsoft/vscode-azuretools/pull/764

Fixes #1610

### DIFF
--- a/src/AppServiceFileSystem.ts
+++ b/src/AppServiceFileSystem.ts
@@ -20,6 +20,10 @@ export class AppServiceFileSystem extends AzExtTreeFileSystem<FileTreeItem> {
         return node.label;
     }
 
+    public getResourceName(node: FileTreeItem): string {
+        return node.client.fullName;
+    }
+
     public async statImpl(_context: IActionContext, _node: FileTreeItem): Promise<FileStat> {
         // this is not implemented for Azure App Services
         return { type: FileType.File, ctime: 0, mtime: 0, size: 0 };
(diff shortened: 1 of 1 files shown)


src/AppServiceFileSystem.ts
```diff
@@ -32,17 +36,6 @@ export class AppServiceFileSystem extends AzExtTreeFileSystem<FileTreeItem> {
...
-            if (result === DialogResponses.alwaysUpload) {
-                await updateGlobalSetting(showSavePromptKey, false);
-            } else if (result === DialogResponses.dontUpload) {
-                throw new UserCancelledError();
-            }
-        }
-
         let etag: string = nonNullValue(this._etags.get(node.fullId), 'etag');
         try {
             await putFile(node.client, content.toString(), node.path, etag);
```
EricJizbaMSFT, line 41: I believe the issue could be solved with just two lines of code...by adding the logs around this call to `putFile`.

I don't think all users of `AzExtTreeFileSystem` want logs, and if they do they might want a different message. We could work around those considerations if we really cared about sharing code, but IMO the "logs" feature doesn't feel important enough to justify the added complexity

alexweininger: Ok, yeah I debated if I should just do that. Should have asked if we wanted to share the logs. 👍

Issue #455, opened 2018-06-04
Keep output window hidden, use long running notifications


fiveisprime:
Keep the output window hidden when deploying or deleting (those are the commands I can think of that open the output window). Still write info to output and just keep it out of the way, instead use the long running notification.

On failed deployment, prompt to check the output window and open to the correct channel if selected.

nturinski:
@fiveisprime I think the answer is yes, but do we want to show the output window when users Start Streaming?

fiveisprime:
Super late considering you just merged this, but... yeah. haha

PR #1689, opened 2020-07-09
Output logs when opening and uploading files

### LINKED ISSUE #1610: There is no output log and notification when updating the file

alexweininger:
Fixes #1610

### DIFF
--- a/src/AppServiceFileSystem.ts
+++ b/src/AppServiceFileSystem.ts
@@ -63,6 +63,5 @@ export class AppServiceFileSystem extends AzExtTreeFileSystem<FileTreeItem> {
 
     public appendLineToOutput(value: string, options?: { resourceName?: string, date?: Date }): void {
         ext.outputChannel.appendLog(value, options);
-        ext.outputChannel.show(true);
     }
 }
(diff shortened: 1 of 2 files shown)


src/AppServiceFileSystem.ts
```diff
@@ -58,4 +60,9 @@ export class AppServiceFileSystem extends AzExtTreeFileSystem<FileTreeItem> {
         this._etags.set(node.fullId, etag);
         await node.refresh();
     }
+
+    public appendLineToOutput(value: string, options?: { resourceName?: string, date?: Date }): void {
+        ext.outputChannel.appendLog(value, options);
+        ext.outputChannel.show(true);
```
EricJizbaMSFT, line 66: We [generally avoid](https://github.com/microsoft/vscode-azureappservice/issues/455) showing the output channel because it can get annoying

src/AppServiceFileSystem.ts
```diff
@@ -58,4 +60,8 @@ export class AppServiceFileSystem extends AzExtTreeFileSystem<FileTreeItem> {
         this._etags.set(node.fullId, etag);
         await node.refresh();
     }
+
+    public appendLineToOutput(value: string, options?: { resourceName?: string, date?: Date }): void {
+        ext.outputChannel.appendLog(value, options);
```
EricJizbaMSFT, line 65: Remove `appendLineToOutput` because it doesn't do anything and just use `ext.outputChannel.appendLog` directly

PR #1698, opened 2020-07-20
Fix removing cosmos connection

### LINKED ISSUE #1697: An error occurs when removing the Cosmos DB connection
**OS:** Win 10
**Build Version:** 20200716.4

**Repro Steps:**
1. Expand one Web App --> Click "Connections" node --> Click "Azure Databases" --> Install Azure Databases extension --> Click "Add Cosmos DB Connection…" and select one Mongo DB database to connect.
2. Right click the connection and select "Remove Connection...".

**Expect:**
Succeed to remove the connection without any error.

**Actual:**
An error occurs when removing the connection.
image

alexweininger:
Fixes #1697

### DIFF
--- a/src/explorer/CosmosDBConnection.ts
+++ b/src/explorer/CosmosDBConnection.ts
@@ -3,14 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { IAppSettingsClient, ISiteTreeRoot } from 'vscode-azureappservice';
-import { AzureTreeItem, DialogResponses } from 'vscode-azureextensionui';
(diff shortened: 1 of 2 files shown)


src/explorer/CosmosDBConnection.ts
```diff
@@ -47,15 +47,16 @@ export class CosmosDBConnection extends AzureTreeItem<ISiteTreeRoot> {
     }
 
     public async deleteTreeItemImpl(): Promise<void> {
-        const appSettings = await this.root.client.listApplicationSettings();
+        const appSettingsClient: IAppSettingsClient = this.parent.parent.client;
```
EricJizbaMSFT, line 50: Should probably make `CosmosDBConnection` extend `AzExtTreeItem` so that we can be sure it doesn't try to use `root` anywhere else

PR #1702, opened 2020-07-21
Reorder loadMoreChildrenImpl calls

### LINKED ISSUE #1600: Sign-in/Create account tree items never load if a trial app exists
### Repro steps:

1. Sign out of Azure
2. Import trial app
3. Reload window

### Expected behavior: 
Sign-in and Create account tree items load, along with the imported trial app.

### Actual behavior:
The `Loading...` node stays there forever.

image

#### Note: Refreshing the tree using the refresh button makes everything load properly.

Possibly related to https://github.com/microsoft/vscode-azuretools/issues/741

alexweininger:
Fixes #1600 for me. However, I'd like some more testing because it's such a weird fix.

This fix works because I moved the `await super.loadMoreChildrenImpl(clearCache, context)` call so that's it's after all other async calls in the function.

### DIFF
--- a/src/explorer/AzureAccountTreeItem.ts
+++ b/src/explorer/AzureAccountTreeItem.ts
@@ -29,6 +29,8 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
     public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
         const ti: AzExtTreeItem | undefined = this.trialAppNode ?? await this.loadTrialAppNode();
         await this.setContext(ti);
+
+        // Must be called after all other async calls to prevent getting stuck on the "Loading..." tree item.
         const children: AzExtTreeItem[] = await super.loadMoreChildrenImpl(clearCache, context);
 
         if (ti) {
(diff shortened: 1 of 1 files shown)


nturinski:
I think we should investigate the root cause more.  We're gonna run out of places to move that line of code 😬

src/explorer/AzureAccountTreeItem.ts
```diff
@@ -28,9 +28,11 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase {
 
     public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
         const ti: AzExtTreeItem | undefined = this.trialAppNode ?? await this.loadTrialAppNode();
-        const children: AzExtTreeItem[] = await super.loadMoreChildrenImpl(clearCache, context);
         await this.setContext(ti);
 
+        // Must be called after all other async calls to prevent getting stuck on the "Loading..." tree item.
```
EricJizbaMSFT, line 33: Lol a bit overconfident IMO. I would add a comment like this:

```
// Putting this after all other async calls to hopefully fix https://github.com/microsoft/vscode-azureappservice/issues/1600
```

PR #1734, opened 2020-08-14
Updates to connections - Postgres support to Azure Databases connections

neelip:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1738


src/vscode-cosmos.api.d.ts
```diff
@@ -54,7 +59,7 @@ export interface DatabaseTreeItem extends DatabaseAccountTreeItem {
 
 export type CosmosDBResourceType = 'DatabaseAccount' | 'Database';
 
-export type CosmosDBApiType = 'Mongo' | 'SQL' | 'Graph' | 'Table';
+export type CosmosDBApiType = 'Mongo' | 'SQL' | 'Graph' | 'Table' | 'Postgres';
```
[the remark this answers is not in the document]
neelip: I guess one option is detect the version and throw an error for any older versions, have the user upgrade to new version. Another option could be to include support for older versions in the code; which possible requires checks at several places of the code. 

But, do we need to support older versions tho given we are trying to rebrand?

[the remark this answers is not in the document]
EricJizbaMSFT: Ideally users who have previously connected a mongo database would still see that database in the tree (even if the UI is slightly different). I think the "rebrand" is more for new users who previously didn't know that we supported postgres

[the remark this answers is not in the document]
neelip: This is the current status of version difference: 
Older App Service + Latest Databases
- Can see all postgres accounts when trying to add connections on App Service, both attached and azure
- However, no appsettings support and no creation of a treeitem in case user does add a postgres connection, it will be added as a MongoURL appsetting and treeitem would not be created

Latest App Service + Older Databases
- Cannot see any postgres accounts when trying to add connections on App Service

The first scenario could be adjusted if needed, but I don't see any harm it keeping it that for this release.

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -159,10 +171,48 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+                const portKey = prefix + this._pgPortSuffix;
+                const host: string | undefined = appSettings[hostKey];
+                const dbName: string | undefined = appSettings[dbNameKey];
+                const username: string | undefined = appSettings[userKey];
+                const password: string | undefined = appSettings[passKey];
+                const port: string | undefined = appSettings[portKey];
+
+                if (host) {
+                    const keys: string[] = [hostKey];
+                    keys.push(userKey, passKey, portKey, dbNameKey);
```
EricJizbaMSFT, line 197: I think this was better how you had it before. We only need to add to keys if the value is defined

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -159,10 +171,48 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+                    let connectionString: string = connectionStringPrefix;
+                    if (username && password) {
+                        connectionString += `${username}:${password}@`;
+                    }
+                    connectionString += `${host}:${port ? port : portDefault}`;
+                    if (dbName) {
+                        connectionString += `/${dbName}`;
```
EricJizbaMSFT, line 204: Technically this should do all the same encoding stuff you're doing in the cosmos repo with encodeUriComponent. However, there's another option...

The connectionString is only used as a part of `TreeItemQuery` in `findTreeItem`. And the only reason we did it that way is because Mongo is the first one we added support for and it saves the app setting as an entire connection string. Since postgres splits up the connection string into multiple app settings, I think it makes total sense to add another option on the query like this:
```
export interface TreeItemQuery {
    /**
     * An account or database connection string
     */
    connectionString?: string;

    postgresData?: {
        hostName?: string;
        // etc.
    }
}
```

That way we don't have to construct the connection string here (and deal with all the encoding issues, etc.) just to have it deconstructed in our own code in the cosmos repo

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -159,10 +171,48 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         return result;
     }
 
+    private detectPostgresConnections(appSettings: { [propertyName: string]: string }): IDetectedConnection[] {
```
nturinski, line 174: This code is pretty similar to `detectDocDBConnetions`.  I wonder if we could do something to share some more code.

I was thinking maybe you could make a helper function that accepts the prefix and suffixes as parameters, and then it can iterate over these arrays.  For instance, with Postgres, you could pass `[_pgHostSuffix,  _pgDbNameSuffix, _pgUserSuffix, _pgPassSuffix, _pgPortSuffix]` and then build all of the appSettings out that you need?

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -215,6 +265,32 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         return new Map([[appSettingKey, database.connectionString]]);
     }
 
+    private async promptForPostgresAppSettings(appSettingsDict: StringDictionary, database: DatabaseTreeItem): Promise<Map<string, string>> {
```
nturinski, line 268: Kind of the same idea here.  It's pretty much the same code as `promptForDocDBAppSettings`

src/explorer/CosmosDBTreeItem.ts
neelip: I agree with you on adding encoding, however I would prefer to keep it with connectionString for this release as the changes here would require quite a few changes in the api code in Databases as well.

src/explorer/CosmosDBTreeItem.ts
neelip: I like this idea, given detectPostgresConnections is too long.

src/explorer/CosmosDBTreeItem.ts
EricJizbaMSFT: Yeah docDbData could do the same thing, but it's been using connectionString for a while, so I don't mind keeping it that way.

On the other hand, neither PR for the postgres connections api (in this repo or Cosmos) has been merged, so it feels a bit premature to delay this kind of change. The suggestion is meant to make your code _less_ complex (aka easier to review/test) because you don't have to encode/decode/parse the connection string on both sides. I also err on the side of putting the work in up-front when dealing with an API between two extensions because of all the versioning complexity involved.

src/explorer/CosmosDBTreeItem.ts
neelip: I tried to simplify the prefix and suffix creation part however it was just creating more redundant code. Even if we can the list of final keys (prefix + suffix) we still need to store them as individual variables in order to access the values from `appSettings` to create the connection string.

src/explorer/CosmosDBTreeItem.ts
neelip: I was able to reduce the promptForAppSettings functions, it is available in the latest commit.

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -132,6 +132,16 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
                 await createdDatabase.cosmosExtensionItem.reveal();
             }
         });
+        if (createdDatabase.cosmosExtensionItem.azureData && createdDatabase.cosmosExtensionItem.connectionString.startsWith('postgres://')) {
+            const openInPortal: vscode.MessageItem = { title: 'Open in Portal' };
```
nturinski, line 136: Please localize the string

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -132,6 +132,16 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+            const manageFirewallMessage: string = localize('manageFirewallRulesMsg', `Manage Firewal Rules for the new Database connection "${createdDatabase.label}".`);
```
nturinski, line 137: When you localize strings, you can't use template strings like this.  Instead, it should be like this:

` const manageFirewallMessage: string = localize('manageFirewallRulesMsg', 'Manage Firewal Rules for the new Database connection "{0}".', createdDatabase.label);`

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -145,97 +168,185 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         throw new Error('You must have the "Azure Databases" extension installed to perform this operation.');
     }
 
-    private detectMongoConnections(appSettings: { [propertyName: string]: string }): IDetectedConnection[] {
-        const result: IDetectedConnection[] = [];
+    private detectConnections(appSettings: { [propertyName: string]: string }): IDetectedConnection[] {
```
EricJizbaMSFT, line 171: > I think it might be worth putting this into its own function. Any type of DB we support from here on out is going to just have to keep getting concatenated and it's getting kind of ugly as it is.

I don't think this is what @nturinski had in mind? Before, it was a single line (even if it was a bit ugly), but now it's a 20-line method. I would prefer if we reverted it back to what it was, unless of course @nturinski can describe a better alternative that addresses his concern. For example, @nturinski - did you just want her to move this single line into it's own method (without refactoring)?:
```
const connections: IDetectedConnection[] = this.detectMongoConnections(appSettings).concat(this.detectDocDBConnections(appSettings)).concat(this.detectPostgresConnections(appSettings));
```

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -119,6 +132,16 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
                 await createdDatabase.cosmosExtensionItem.reveal();
             }
         });
+        if (createdDatabase.cosmosExtensionItem.azureData && createdDatabase.cosmosExtensionItem.connectionString.startsWith('postgres://')) {
```
EricJizbaMSFT, line 135: I don't like having multiple prompts at once. How about you add a "Manage Firewall Rules" button to the above prompt?

As an aside - you see how adding this new code (aka the firewall stuff) caused Nathan and I to add more comments to the PR? This is actually one pitfall to avoid that can keep PRs open for longer. It's best to focus on existing comments so that a longstanding PR can get merged, and work on new enhancements separately after that's done

src/explorer/CosmosDBConnection.ts
```diff
@@ -43,7 +43,10 @@ export class CosmosDBConnection extends AzExtTreeItem {
     }
 
     public get iconPath(): IThemedIconPath {
-        return getThemedIconPath('Database');
+        if (this.cosmosExtensionItem.connectionString.startsWith('postgres://')) {
```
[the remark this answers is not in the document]
neelip: postgresData will be undefined for azure subscriptions with Databases implementation right now, working on a fix for this.

src/explorer/CosmosDBTreeItem.ts
nturinski: Yeah, I was thinking that it could just be its own function called getAllDetectedConnections or something to that effect.  I just think it's kind of ugly and confusing as to what it's doing, so a descriptive name could have been helpful.

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -202,44 +290,53 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
         return result;
     }
 
-    private async promptForMongoAppSettings(appSettingsDict: StringDictionary, database: DatabaseTreeItem): Promise<Map<string, string>> {
+    private async promptForAppSettings(appSettingsDict: StringDictionary, accountAppSettings: Map<string | undefined, string | undefined>, suffixes: string[] | undefined, errorMsg: string, defaultPrefixString: string): Promise<Map<string, string>> {
         const prompt: string = 'Enter new connection setting key';
```
EricJizbaMSFT, line 294: This message needs to be adjusted based on whether or not there's suffixes, but it feels like overkill to pass in the messages. How about you do this:
```typescript
const prompt: string = suffixes ? localize('enterPrefix', 'Enter new connection setting prefix') : localize('enterKey', 'Enter new connection setting key');
```

And do something similar for the error message below, instead of passing it in.

src/constants.ts
```diff
@@ -35,3 +35,7 @@ export const envFileName: string = '.env';
 export const detectorTimestampFormat: string = 'YYYY-MM-DDTHH:mm';
 
 export const TrialAppContext: string = 'TrialAppContext';
+
+export const databasesPostgresPrefix = 'postgres://';
+
+export enum defaultPrefix { AZURE_COSMOS = 'AZURE_COSMOS', POSTGRES = 'POSTGRES', MONGO_URL = 'MONGO_URL' }
```
EricJizbaMSFT, line 41: There are only used once, so an enum feels excessive IMO

src/explorer/CosmosDBTreeItem.ts
```diff
@@ -110,24 +141,31 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
...
+        if (createdDatabase.cosmosExtensionItem.azureData && createdDatabase.cosmosExtensionItem.postgresData) {
+            buttons.push(manageFirewallRules);
+        }
+        vscode.window.showInformationMessage(message, ...buttons).then(async (result: vscode.MessageItem | undefined) => {
             if (result === revealDatabase) {
                 await createdDatabase.cosmosExtensionItem.reveal();
+            } else if (result === manageFirewallRules) {
+                const resourceID: string = 'connectionSecurity';
+                const accountId: string | undefined = createdDatabase.cosmosExtensionItem.azureData?.accountId;
+                await openInPortal(this.root, `${accountId}` + `/${resourceID}`);
```
EricJizbaMSFT, line 158:
```suggestion
                const accountId: string | undefined = createdDatabase.cosmosExtensionItem.azureData?.accountId;
                await openInPortal(this.root, `${accountId}/connectionSecurity`);
```

EricJizbaMSFT (approved):

PR #1743, opened 2020-08-26
Exclude dev sections from marketplace page

EricJizbaMSFT:
Same as https://github.com/microsoft/vscode-azurefunctions/pull/2339

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 # Azure App Service for Visual Studio Code (Preview)
 
+<!-- region exclude-from-marketplace -->
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -67,6 +71,8 @@ Connecting to log stream...
 
 * Local Git deployment may fail with large commits
 
+<!-- region exclude-from-marketplace -->
```
nturinski, line 74: How come we want to remove this from the marketplace btw?

EricJizbaMSFT: Just didn't feel like this info was relevant for people viewing our marketplace page. The typical user that installs our extension doesn't care about how to debug our extension or how to sign a CLA. And for the few users that actually want to contribute, there's a "Repository" button to take them to the source code where they can see this info

Screen Shot 2020-08-26 at 1 44 17 PM

PR #1758, opened 2020-09-08
Update test case for application setting

v-wuzhai:


test/nightly/azureResourceWebAppActions.test.ts
```diff
@@ -102,18 +102,22 @@ suite('Web App actions', async function (this: Mocha.Suite): Promise<void> {
     });
 
     test(`Add and delete settings for ${WebsiteOS0} Web App`, async () => {
-        const appSettingKey: string = getRandomHexString();
-        const appSettingValue: string = getRandomHexString();
+        let appSettingKey: string;
+        const appSettingList: string[] = [getRandomHexString(10), `f ${getRandomHexString(10)}`, `${getRandomHexString(10)} `, ` ${getRandomHexString(10)}`];
```
nturinski, line 106: Did you mean to add a space at the end of one of this app settings?  Since we take the user's value and trim it, all the whitespace will be gone and this will cause the tests to fail, but this is by-design.

PR #1783, opened 2020-09-24
Add survey prompt on certain deployment failures

zooba:
Right now we want to prompt on certain deployment failures for Python apps, but it should be fairly easy to extend for more cases.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -9,7 +9,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import * as appservice from 'vscode-azureappservice';
 import { getDeployFsPath, getDeployNode, IDeployContext, IDeployPaths, showDeployConfirmation } from 'vscode-azureappservice';
-import { IActionContext } from 'vscode-azureextensionui';
+import { IActionContext, parseError } from 'vscode-azureextensionui';
 import * as constants from '../../constants';
 import { SiteTreeItem } from '../../explorer/SiteTreeItem';
 import { TrialAppTreeItem } from '../../explorer/trialApp/TrialAppTreeItem';
@@ -23,6 +23,7 @@ import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
 import { LinuxRuntimes } from '../createWebApp/LinuxRuntimes';
 import { runPostDeployTask } from '../postDeploy/runPostDeployTask';
 import { deployTrialApp } from './deployTrialApp';
+import { failureMoreInfoSurvey } from './failureMoreInfoSurvey';
 import { enableScmDoBuildDuringDeploy, promptScmDoBuildDeploy } from './promptScmDoBuildDeploy';
 import { promptToSaveDeployDefaults, saveDeployDefaults } from './promptToSaveDeployDefaults';
 import { setPreDeployTaskForDotnet } from './setPreDeployTaskForDotnet';
@@ -107,7 +108,14 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri | 
     }
 
     await node.runWithTemporaryDescription("Deploying...", async () => {
-        await appservice.deploy(nonNullValue(node).root.client, <string>deployPath, context);
+        try {
+            await appservice.deploy(nonNullValue(node).root.client, <string>deployPath, context);
+        } catch (error) {
+            if (failureMoreInfoSurvey(parseError(error), nonNullValue(siteConfig))) {
+                actionContext.errorHandling.suppressDisplay = true;
+            }
+            throw error;
+        }
     });
 
     const tokenSource: vscode.CancellationTokenSource = new vscode.CancellationTokenSource();
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -111,7 +111,8 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri | 
         try {
             await appservice.deploy(nonNullValue(node).root.client, <string>deployPath, context);
         } catch (error) {
-            if (failureMoreInfoSurvey(parseError(error), nonNullValue(siteConfig))) {
+            if (!actionContext.errorHandling.suppressDisplay
+                && failureMoreInfoSurvey(parseError(error), nonNullValue(siteConfig))) {
                 actionContext.errorHandling.suppressDisplay = true;
             }
             throw error;
--- a/src/commands/deploy/failureMoreInfoSurvey.ts
+++ b/src/commands/deploy/failureMoreInfoSurvey.ts
@@ -1,0 +1,81 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { WebSiteManagementModels } from '@azure/arm-appservice';
+import { env, MessageItem, Uri, window } from 'vscode';
+import * as appservice from 'vscode-azureappservice';
+import { DialogResponses, IParsedError } from 'vscode-azureextensionui';
+import { localize } from "../../localize";
+
+const SURVEY_URL: string = 'https://aka.ms/AppServiceExtDeploymentFeedback';
+
+function shouldPromptForSurvey(error: IParsedError, siteConfig: WebSiteManagementModels.SiteConfigResource): string | null {
+    if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.toLowerCase().startsWith(appservice.LinuxRuntimes.python)) {
+        if (error.errorType === 'Error') {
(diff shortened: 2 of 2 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -118,3 +119,27 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
     // tslint:disable-next-line: no-floating-promises
     runPostDeployTask(node, correlationId, tokenSource);
 }
+
+async function runDeployTask(context: IDeployContext, client: appservice.SiteClient, siteConfig: WebSiteModels.SiteConfigResource, deployPath: string): Promise<void> {
```
EricJizbaMSFT, line 123: "task" has a special VS Code meaning, which doesn't apply here. We could find a new name, but seems easier to just move this code up into line 110 above. You already encapsulated most of the survey stuff into `shouldPromptForSurvey` and `failureMoreInfoSurvey`

src/commands/deploy/deploy.ts
```diff
@@ -118,3 +119,27 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
...
+        if (key) {
+            failureMoreInfoSurvey(ext.context, key);
+        }
+        throw error;
+    }
+}
+
+function shouldPromptForSurvey(error: IParsedError, siteConfig: WebSiteModels.SiteConfigResource): string | null {
+    if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.toLowerCase().startsWith(appservice.LinuxRuntimes.python)) {
+        if (error.errorType === 'Error' && error.message.toLowerCase().startsWith("the service is unavailable")) {
```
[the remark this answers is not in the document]
zooba: Follow up on the email thread for this one.

src/commands/deploy/failureMoreInfoSurvey.ts
```diff
@@ -0,0 +1,50 @@
...
+            }
+        };
+        const never = {
+            title: localize('neverAgain', "Don't Show Again"),
+            isSecondary: true,
+            run: async () => {
+                await globalState.update(NEVER_AGAIN_KEY, true);
+            }
+        };
+        const button = await window.showInformationMessage(
```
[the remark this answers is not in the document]
zooba: Bring this to the email thread too. This is already limited to "people who have hit a known but undiagnosed deployment issue", so we definitely don't want a random sample.

I understand the survey is going to ask about the project that's being deployed so we can figure out if there is anything in common, so perhaps limiting it to once-per-site-per-week will do?

src/commands/deploy/failureMoreInfoSurvey.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        const take = {
+            title: localize("provideFeedback", "Provide feedback"),
+            run: async () => {
+                env.openExternal(Uri.parse(`${SURVEY_URL}?k=${encodeURIComponent(failureKey)}&m=${encodeURIComponent(env.machineId)}`));
+            }
+        };
+        const never = {
+            title: localize('neverAgain', "Don't Show Again"),
+            isSecondary: true,
```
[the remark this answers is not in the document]
zooba: No idea, I copied it from another survey. I'll try and dig into it.

src/commands/deploy/failureMoreInfoSurvey.ts
```diff
@@ -0,0 +1,50 @@
...
+        if (env.language !== 'en' && !env.language.startsWith('en-')) {
+            return;
+        }
+
+        const skip = globalState.get(NEVER_AGAIN_KEY, false);
+        if (skip !== false) {
+            return;
+        }
+
+        const take = {
```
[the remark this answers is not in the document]
zooba: Got any examples I can copy? I'm not a TypeScript developer, just pretending to be one since the team is overburdened ;)

src/commands/deploy/deploy.ts
zooba: First time I tried to just keep it in the same function I hit a linting limit on complexity, but it seems to fit under the limit now.

src/commands/deploy/failureMoreInfoSurvey.ts
```diff
@@ -0,0 +1,47 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+'use strict';
```
EricJizbaMSFT, line 6: You can remove this. It should be controlled at the repo-level, not the file-level

src/commands/deploy/failureMoreInfoSurvey.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { env, ExtensionContext, Uri, window } from 'vscode';
+import { localize } from "../../localize";
+
+const SURVEY_URL = 'https://aka.ms/AppServiceExtDeploymentFeedback';
+
+const NEVER_AGAIN_KEY = 'failureMoreInfoSurvey/neverAgain';
```
EricJizbaMSFT, line 13: The typedef stuff is pretty trivial. For example here, all you need to do here is:
```
const NEVER_AGAIN_KEY: string = 'failureMoreInfoSurvey/neverAgain';
```

If you merge with master, the lint rule is enabled now. https://github.com/microsoft/vscode-azureappservice/pull/1788

The only weird type is your "take"/"never" buttons with "run". Normally we would do it like [this](https://github.com/microsoft/vscode-azureappservice/blob/b4cec5cc748e23833e191645ea295e3cc0d7f28a/src/commands/deploy/promptScmDoBuildDeploy.ts#L22) or [this](https://github.com/microsoft/vscode-azureappservice/blob/b4cec5cc748e23833e191645ea295e3cc0d7f28a/src/commands/deploy/promptToSaveDeployDefaults.ts#L20), where the button (of type `MessageItem`) is separate from what happens after you click it

src/commands/deploy/deploy.ts
```diff
@@ -106,7 +107,14 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
     }
 
     await node.runWithTemporaryDescription("Deploying...", async () => {
-        await appservice.deploy(nonNullValue(node).root.client, <string>deployPath, context);
+        try {
+            await appservice.deploy(nonNullValue(node).root.client, <string>deployPath, context);
+        } catch (error) {
+            if (failureMoreInfoSurvey(parseError(error), nonNullValue(siteConfig))) {
+                actionContext.errorHandling.suppressDisplay = true;
```
zooba, line 114: This doesn't seem to work, but I can't tell if I've just got mismatched packages. I can step through one of the `callWithTelemetryAndErrorHandling` calls that deployment is wrapped in, but apparently there's another one that ignores the context and displays the popup anyway?

FWIW, my deployment is broken because I have `azure-mgmt` in my requirements.txt (which is not installable), in case someone wants to attempt a repro.

EricJizbaMSFT: The other popup is coming from here: https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/deploy/waitForDeploymentToComplete.ts#L82

The problem is that it's _also_ doing custom error handling (So that we can add a "View Output" button), meaning it actually pops up the message before you even catch the error here. My gut instinct is that `shouldPromptForSurvey` should return false if `errorHandling.suppressDisplay` is already set to `true`. That specific error case usually has helpful logs in the output, i.e. a pip failure related to `azure-mgmt`, so I think directing them to the output is more important than prompting for the survey

src/commands/deploy/failureMoreInfoSurvey.ts
```diff
@@ -0,0 +1,81 @@
...
+import { WebSiteManagementModels } from '@azure/arm-appservice';
+import { env, MessageItem, Uri, window } from 'vscode';
+import * as appservice from 'vscode-azureappservice';
+import { DialogResponses, IParsedError } from 'vscode-azureextensionui';
+import { localize } from "../../localize";
+
+const SURVEY_URL: string = 'https://aka.ms/AppServiceExtDeploymentFeedback';
+
+function shouldPromptForSurvey(error: IParsedError, siteConfig: WebSiteManagementModels.SiteConfigResource): string | null {
+    if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.toLowerCase().startsWith(appservice.LinuxRuntimes.python)) {
```
EricJizbaMSFT, line 15: Fyi, I [moved](https://github.com/microsoft/vscode-azureappservice/pull/1824) `LinuxRuntimes` from the `appservice` package to a file in this repo:

https://github.com/microsoft/vscode-azureappservice/blob/main/src/commands/createWebApp/LinuxRuntimes.ts

src/commands/deploy/deploy.ts
zooba: Yes, you're right. That message is always going to win, so that's where this whole change should have been made.

I'll post on the thread.

EricJizbaMSFT (approved):
🎉

Issue #904, opened 2019-03-12
Switch over to azure-sdk-for-js


EricJizbaMSFT:
https://github.com/Azure/azure-sdk-for-js/blob/master/documentation/Migration.md

Ladiesoftheweb:
Switched to azure sdk for js

PR #1816, opened 2020-10-21
Update to new azure-sdk-for-js

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/904


src/commands/createWebApp/setPostPromptDefaults.ts
```diff
@@ -78,12 +78,7 @@ export async function setPostPromptDefaults(wizardContext: IAppServiceWizardCont
     }
 }
 
-export async function getAppServicePlan(wizardContext: IAppServiceWizardContext, rgName: string, newPlanName: string): Promise<AppServicePlan | null> {
-    const client: WebSiteManagementClient = createAzureClient(wizardContext, WebSiteManagementClient);
-    return await client.appServicePlans.get(rgName, newPlanName);
-}
-
-function checkPlanForPerformanceDrop(asp: AppServicePlan): boolean {
+function checkPlanForPerformanceDrop(asp: WebSiteManagementModels.AppServicePlan): boolean {
```
nturinski, line 81: Just to make sure, does` client.appServicePlans.list` now also return the asp's `numberOfSites` property?

src/commands/postDeploy/validateWebSite.ts
```diff
@@ -54,14 +47,10 @@ export async function validateWebSite(deploymentCorrelationId: string, siteTreeI
...
-                request.resolveWithFullResponse = true;
-                const response = await requestUtils.sendRequest<IncomingMessage>(request);
-                currentStatusCode = response.statusCode;
+                const response: HttpOperationResponse = await client.sendRequest({ method: 'GET', url });
+                currentStatusCode = response.status;
             } catch (error) {
-                // tslint:disable-next-line:strict-boolean-expressions
-                const response = (<WebError>error).response || {};
-                currentStatusCode = isNumber(response.statusCode) ? response.statusCode : 0;
+                currentStatusCode = error instanceof RestError ? error.statusCode : 0;
```
nturinski, line 53: The behavior seems a little different here.  Did we only want to record RestError status codes?

src/explorer/trialApp/TrialAppClient.ts
```diff
@@ -65,34 +63,35 @@ export class TrialAppClient implements ISimplifiedSiteClient {
...
+        });
     }
 
-    public async listApplicationSettings(): Promise<StringDictionary> {
+    public async listApplicationSettings(): Promise<WebSiteManagementModels.StringDictionary> {
         const kuduClient: KuduClient = await this.getKuduClient();
-        const settings: StringDictionary = {};
-        settings.properties = <{ [name: string]: string }>await kuduClient.settings.getAll();
+        const settings: WebSiteManagementModels.StringDictionary = {};
+        settings.properties = <{ [name: string]: string }>(await kuduClient.settings.getAll()).body;
```
nturinski, line 89: Why did the response type here change?

nturinski (approved):

src/commands/postDeploy/validateWebSite.ts
EricJizbaMSFT: I think the difference you're noticing just comes down to "request-promise" vs. "@azure/ms-rest-js". If there is a statusCode in the error, it'll be a `RestError` - that's just how "@azure/ms-rest-js" works. I validated a few sites with successful status codes (2xx) and a few with failed status codes (4xx) and we were working the same

src/commands/createWebApp/setPostPromptDefaults.ts
EricJizbaMSFT: Discussed offline

PR #1824, opened 2020-10-26
Copy SiteRuntimeStep over from appservice package

EricJizbaMSFT:
Related to https://github.com/microsoft/vscode-azuretools/pull/815, I want to move this code into the app service extension since it's not used by Functions. All of the code added in this PR is purely copied without functional changes, so it should be a quick review.

This will make my life easier as I work on .NET 5 support


nturinski:
Not required, but should probably bump the appservice package here too.

nturinski (approved):

Issue #301, opened 2018-02-08
Properties tool tip for apps


nturinski:
Leverage new VS Code "inline action" to have a tool tip that will display more properties of the web app

PR #1829, opened 2020-10-28
Add "View Properties" for web apps

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/301

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
         "onCommand:appService.RemoveCosmosDBConnection",
         "onCommand:appService.RevealConnection",
         "onCommand:appService.RevealConnectionInAppSettings",
+        "onCommand:appService.ViewProperties",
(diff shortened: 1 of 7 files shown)


nturinski (commented):
It feels a little confusing to me to have both the siteClient and site exist on SiteTreeItem.  Couldn't we just call `node.root.client.getSiteConfig()` like you're doing today?  It seems like that always gets called (I don't see any checks to see if it is cached or not) so I don't see the reason to add site to SiteTreeItem.

EricJizbaMSFT:
`site` and `siteConfig` are not the same thing. Calling `getSiteConfig` will not get me the whole `site` object

nturinski:
Oh, I see, you're actually passing the entire site into the openJson.  Okay then.

nturinski (approved):

PR #1835, opened 2020-11-03
Use new api to display runtime stacks

EricJizbaMSFT:
Some benefits to the new API (which the portal is already using):
1. The list of runtimes is more accurate (for example, the old API listed weird things like "State Site 1.0")
1. We no longer have to hard-code which runtime defaults to which OS (the API tells us). On a related note, we now prompt for runtime _before_ OS, and only prompt for OS if the selected runtime supports both.
1. For Java, we only list "Java 8" and "Java 11" in the main prompt, and have a separate prompt for the server stack

|Old|New|
|---|---|
|Screen Shot 2020-11-03 at 10 20 52 AM|Screen Shot 2020-11-03 at 10 22 25 AM|
||Screen Shot 2020-11-03 at 10 22 46 AM|

Other improvements in this PR:
1. Added a decent amount of logic in `WebAppCreateStep.getSiteConfig` related to runtime versions on Windows plans. In the past we never did any of this and users would've had to manually edit their site to get certain runtimes to work.
1. Updated the "Create Web App and deploy" tests to include windows plans in addition to linux. Also updated the test to verify it's actually running the version we selected
1.  Added a "showHiddenStacks" setting to test stacks that are in the works. This is helpful because ".NET 5" support is still hidden at the moment


src/commands/createWebApp/WebAppCreateStep.ts
```diff
@@ -51,7 +55,51 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
...
+            newSiteConfig.linuxFxVersion = stack.stack.value === 'java' ?
+                getJavaLinuxRuntime(stack.majorVersion.value, nonNullProp(context, 'newSiteJavaStack').minorVersion) :
+                nonNullProp(stack.minorVersion.stackSettings, 'linuxRuntimeSettings').runtimeVersion;
+        } else {
+            const runtimeVersion: string = nonNullProp(stack.minorVersion.stackSettings, 'windowsRuntimeSettings').runtimeVersion;
+            switch (stack.stack.value) {
+                case 'dotnet':
+                    if (!/core/i.test(stack.minorVersion.displayText)) {
+                        newSiteConfig.netFrameworkVersion = runtimeVersion;
+                    }
```
wwlorey, line 74: What is this check doing?

src/commands/createWebApp/setPrePromptDefaults.ts
```diff
@@ -22,17 +21,13 @@ export async function setPrePromptDefaults(wizardContext: IWebAppWizardContext &
         const fsPath: string = workspaceForRecommendation.uri.fsPath;
 
         if (await fse.pathExists(path.join(fsPath, 'package.json'))) {
-            wizardContext.recommendedSiteRuntime = [LinuxRuntimes.node];
-
+            wizardContext.recommendedSiteRuntime = ['node'];
```
wwlorey, line 24: Did you stop using `LinuxRuntimes` because these runtimes aren't necessarily for linux?

src/commands/createWebApp/stacks/getStackPicks.ts
```diff
@@ -0,0 +1,76 @@
...
+    for (const stack of stacks) {
+        for (const majorVersion of stack.majorVersions) {
+            const minorVersions: (AppStackMinorVersion<WebAppRuntimes & JavaContainers>)[] = majorVersion.minorVersions
+                // Filter out versions that have major, minor, _and_ patch specified (Aka we only want Node.js 14, not Node.js 14.1.1)
+                .filter(mv => !/\..*\./.test(mv.value))
+                // Filter out versions relative to the specified javaVersion
+                .filter(mv => {
+                    return (!javaVersion && (mv.stackSettings.linuxRuntimeSettings || mv.stackSettings.windowsRuntimeSettings)) ||
+                        (javaVersion && (getJavaLinuxRuntime(javaVersion, mv) || mv.stackSettings.windowsContainerSettings));
+                });
```
wwlorey, line 29: I found this difficult to understand but maybe that's because I'm not super familiar with these stacks 😅

wwlorey (approved):

src/commands/createWebApp/setPrePromptDefaults.ts
EricJizbaMSFT: Mostly because I wanted to use `WebAppStackValue` instead (from `WebAppStackModel.ts`). That type is copied from the source repo for the stacks API, so it'll be easier for us to stay in-sync. Keep in mind, it might look like a simple string, but it's not - we still get standard errors for things like misspelling, etc.:

Screen Shot 2020-11-05 at 1 26 04 PM

src/commands/createWebApp/WebAppCreateStep.ts
EricJizbaMSFT: I'm not sure how familiar you are with the differences between .NET Framework and .NET Core, but this property is only for .NET Framework and doesn't apply to .NET Core. I'm not a huge fan of using `displayText` to figure out which ones are .NET Core (for localization reasons, etc.), but I didn't find a better property in the stack object.

src/commands/createWebApp/stacks/getStackPicks.ts
EricJizbaMSFT: I found a slightly cleaner way to accomplish what I was doing here

PR #1838, opened 2020-11-05
Add a test case for deploying .net core project to windows web app

v-wuzhai:


EricJizbaMSFT:
This is already covered by https://github.com/microsoft/vscode-azureappservice/pull/1835

PR #1847, opened 2020-11-10
Add 'Early Access' label to .NET 5

EricJizbaMSFT:
First, I had to add support for the "isEarlyAccess" flag itself in our UI
Second, I had to hard-code the flag for .NET 5 because it's not set yet (related to https://github.com/Azure/azure-functions-ux/pull/5543)

### DIFF
--- a/src/commands/createWebApp/stacks/getStackPicks.ts
+++ b/src/commands/createWebApp/stacks/getStackPicks.ts
@@ -24,9 +24,10 @@ export async function getStackPicks(context: IWebAppWizardContext, javaVersion?:
 
             for (const minorVersion of minorVersions) {
                 let description: string | undefined;
-                if ((!javaVersion && (minorVersion.stackSettings.linuxRuntimeSettings?.isPreview || minorVersion.stackSettings.windowsRuntimeSettings?.isPreview)) ||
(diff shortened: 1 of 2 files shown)


src/commands/createWebApp/stacks/getStackPicks.ts
```diff
@@ -24,9 +24,10 @@ export async function getStackPicks(context: IWebAppWizardContext, javaVersion?:
 
             for (const minorVersion of minorVersions) {
                 let description: string | undefined;
-                if ((!javaVersion && (minorVersion.stackSettings.linuxRuntimeSettings?.isPreview || minorVersion.stackSettings.windowsRuntimeSettings?.isPreview)) ||
-                    (javaVersion && (minorVersion.stackSettings.linuxContainerSettings?.isPreview || minorVersion.stackSettings.windowsContainerSettings?.isPreview))) {
+                if (isFlagSet(minorVersion.stackSettings, 'isPreview')) {
```
nturinski, line 27: Should we still confirm it's not a Java version, whether that be in this conditional or in the function on line 45?

I recognize that JavaContainers seem to have LinuxJavaContainerSettings and windows settings, but this isn't the behavior that we had previously.

nturinski (approved):

src/commands/createWebApp/stacks/getStackPicks.ts
EricJizbaMSFT: Discussed offline

PR #1866, opened 2020-11-24
Run "Create and Deploy" tests in parallel

ejizba:
Depends on https://github.com/microsoft/vscode-azuretools/pull/820 and https://github.com/microsoft/vscode-azuretools/pull/821

Got it from ~11 minutes down to ~2 minutes on my local machine.


nturinski (approved):
Nifty

Issue #1880, opened 2020-12-04
Note: tryappservice is deprecated


rzhao271:
I deprecated https://code.visualstudio.com/tryappservice/ because it is outdated.
It now redirects to https://code.visualstudio.com/docs/azure/deployment.

The link in the README that says "try Azure for free" might need updating

nturinski:
Does this mean Try App Service is completely deprecated?  These two webpages didn't convey the same information.

rzhao271:
Yep, completely deprecated.
I think @fiveisprime suggested just removing any links that point to that URL.

ejizba:
We're still confused. Are you saying the _documentation_ is completely deprecated? Or the feature itself is completely deprecated? If it's just the documentation - can you explain what was outdated about it?

rzhao271:
The feature itself is completely deprecated. There's no owner for it and there were many security vulnerabilities

alexweininger:
😭

nturinski:
😭🌹

ejizba:
![salute](https://media3.giphy.com/media/fnKd6rCHaZoGdzLjjA/giphy.gif)

PR #1947, opened 2021-02-01
Remove trial app functionality

ejizba:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/1880 😢

Trial apps haven't worked for months, so the only stuff I see in telemetry is "createTrialApp" which just takes people to [this unhelpful link](https://code.visualstudio.com/docs/azure/deployment). Point is, users just can't physically hit this code anymore so I think the best thing to do is just rip it out.


src/explorer/trialApp/TrialAppTreeItem.ts
```diff
@@ -1,152 +0,0 @@
...
-    private readonly _appSettingsTreeItem: TrialAppApplicationSettingsTreeItem;
-    private readonly _siteFilesNode: SiteFilesTreeItem;
-    private readonly _tutorialNode: GenericTreeItem;
-
-    private constructor(parent: AzureAccountTreeItem, client: TrialAppClient) {
-        super(parent);
-        this.client = client;
-        this._appSettingsTreeItem = new TrialAppApplicationSettingsTreeItem(this, this.client, false, settingsToHide);
-        this._siteFilesNode = new SiteFilesTreeItem(this, this.client, false);
-        this._tutorialNode = new GenericTreeItem(this, { label: 'Show tutorial', commandId: 'appService.ShowTutorial', contextValue: 'showTutorial', iconPath: getThemedIconPath('book') });
```
nturinski, line 49: Can also remove the book.svg from the resources

src/commands/logstream/enableFileLogging.ts
```diff
@@ -25,7 +24,7 @@ export async function enableFileLogging(context: IEnableFileLoggingContext, node
         node = <SiteTreeItem>node.parent;
     }
 
-    const siteNode: SiteTreeItem | TrialAppTreeItem = node;
+    const siteNode: SiteTreeItem = node;
```
nturinski, line 27: nit: This and fullName are kind of pointless `consts` now

nturinski (approved):
You could get rid of the `errors.ts`  file as well if you changed `disconnectRepo` since it was only using the `OperationNotSupportedError ` in case the parent was a trial app.

src/commands/logstream/enableFileLogging.ts
ejizba: `siteNode` still helps
Screen Shot 2021-02-02 at 1 30 37 PM

src/commands/registerCommands.ts
```diff
@@ -62,32 +56,27 @@ export function registerCommands(): void {
...
-    registerCommand('appService.CloneTrialApp', cloneTrialApp);
     registerCommand('appService.ConfigureDeploymentSource', editScmType);
     registerCommand('appService.connectToGitHub', connectToGitHub);
     registerCommand('appService.CreateSlot', createSlot);
-    registerCommand('appService.CreateTrialApp', createTrialApp);
     registerCommand('appService.CreateWebApp', createWebApp);
     registerCommand('appService.CreateWebAppAdvanced', createWebAppAdvanced);
     registerCommand('appService.Delete', deleteWebApp);
     registerCommand('appService.DeploymentScript', generateDeploymentScript);
     registerCommand('appService.DisconnectRepo', disconnectRepo);
```
ejizba, line 66: > You could get rid of the errors.ts file as well if you changed disconnectRepo since it was only using the OperationNotSupportedError in case the parent was a trial app.

I didn't think of this, but now I feel like I'd rather leave it as is. I don't want to get rid of `node.parent instanceof SiteTreeItem` in `disconnectRepo` because I'd have to cast `parent` to `SiteTreeItem` and I don't like casting as much. Also, `errors.ts` is a common enough pattern across all our repos that I don't mind keeping it even if it's not super useful.

PR #1950, opened 2021-02-04
Make all strings localizable

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/38

Very possible I missed something, but I tried to be as thorough as I could.  Did some light spot checks to make sure that UI strings still appeared as they should.

Note that not _all_ strings were made localizable, just ones that users would see (and made sense to localize them).

### DIFF
--- a/src/commands/appSettings/downloadAppSettings.ts
+++ b/src/commands/appSettings/downloadAppSettings.ts
@@ -29,7 +29,7 @@ export async function downloadAppSettings(context: IActionContext, node?: AppSet
     const envVarUri: vscode.Uri = vscode.Uri.file(envVarPath);
 
     await node.runWithTemporaryDescription(context, localize('downloading', 'Downloading...'), async () => {
-        ext.outputChannel.appendLog(localize('downloading', 'Downloading settings from "{0}"...', client.fullName));
+        ext.outputChannel.appendLog(localize('downloadingSettings', 'Downloading settings from "{0}"...', client.fullName));
         const localEnvVariables: DotenvParseOutput = await getLocalEnvironmentVariables(envVarPath, true /* allowOverwrite */);
         const remoteEnvVariables: WebSiteManagementModels.StringDictionary = await client.listApplicationSettings();
         if (remoteEnvVariables.properties) {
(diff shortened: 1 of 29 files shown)


src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -23,12 +24,12 @@ export async function downloadAppSettings(context: IActionContext, node?: AppSet
...
 
-    const message: string = 'Select the destination file for your downloaded settings.';
+    const message: string = localize('selectDest', 'Select the destination file for your downloaded settings.');
     const envVarPath: string = await workspaceUtil.selectWorkspaceFile(message, () => envFileName);
     const envVarUri: vscode.Uri = vscode.Uri.file(envVarPath);
 
-    await node.runWithTemporaryDescription(context, 'Downloading...', async () => {
-        ext.outputChannel.appendLog(`Downloading settings from "${client.fullName}"...`);
+    await node.runWithTemporaryDescription(context, localize('downloading', 'Downloading...'), async () => {
+        ext.outputChannel.appendLog(localize('downloading', 'Downloading settings from "{0}"...', client.fullName));
```
ejizba, line 32: Use a different name from the one above ("downloading")

src/explorer/CosmosDBConnection.ts
```diff
@@ -53,8 +54,8 @@ export class CosmosDBConnection extends AzExtTreeItem {
         const appSettingsClient: IAppSettingsClient = this.parent.parent.client;
         const appSettings: WebSiteManagementModels.StringDictionary = await appSettingsClient.listApplicationSettings();
         if (appSettings.properties) {
-            const warning: string = `Are you sure you want to remove connection "${this.label}"? This will delete the following application settings: ${this.appSettingKeys.map((s) => `"${s}"`).join(', ')}.`;
-            await context.ui.showWarningMessage(warning, { modal: true }, DialogResponses.deleteResponse);
+            const warning: string = localize('removeConnection', 'Are you sure you want to remove connection "{0}"? This will delete the following application settings:', this.label);
+            await context.ui.showWarningMessage(`${warning} ${this.appSettingKeys.map((s) => `"${s}"`).join(', ')}`, { modal: true }, DialogResponses.deleteResponse);
```
ejizba, line 58: How come you aren't including this kind of stuff in localize? I would say we should

src/utils/nonNull.ts
```diff
@@ -21,7 +22,7 @@ export function nonNullValue<T>(value: T | undefined, propertyNameOrMessage?: st
     if (isNullOrUndefined(value)) {
         throw new Error(
             // tslint:disable-next-line:prefer-template
-            'Internal error: Expected value to be neither null nor undefined'
+            localize('notNullUndefined', 'Internal error: Expected value to be neither null nor undefined')
```
ejizba, line 25: I don't want to localize internal errors - this is for our telemetry, not for showing the user.

src/explorer/CosmosDBConnection.ts
nturinski: Well, it should just be the keys from the appSettings.  I guess the their position in the sentence might be something that needs to be localized?

ejizba: Yes, exactly

ejizba (approved):

PR #1955, opened 2021-02-09
Switch to ESLint

wwlorey:


test/parseDetectorTables.test.ts
```diff
@@ -8,18 +8,21 @@ import { ColumnName, detectorTable, findTableByName, getValuesByColumnName, vali
 import { createTestContext, ITestContext } from './global.test';
 
 suite('Detector Dataset Parser', () => {
-    test('Find table by table name', async () => {
+    /* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
ejizba, line 11: Can you do a test-specific eslint file just like we had a test-specifc tslint.json? I would disable these "any" rules for the whole test folder, just like it used to be.

package.json
```diff
@@ -812,8 +812,8 @@
         "cleanReadme": "gulp cleanReadme",
         "compile": "tsc -watch",
         "package": "vsce package --githubBranch main",
-        "lint": "eslint -c .eslintrc.js --ext .ts .",
-        "lint-fix": "eslint -c .eslintrc.js --ext .ts . --fix",
+        "lint": "eslint --ext .ts .",
+        "lint-fix": "eslint --ext .ts . --fix",
```
wwlorey, line 816: For ESLint to use the `.eslintrc.js` in `test/` I had to remove the config file option here

test/.eslintrc.js
```diff
@@ -0,0 +1,43 @@
+module.exports = {
```
ejizba, line 1: I'm pretty sure there's a way you can remove the duplicate code and get this based off the root `.eslintrc.js`

This doc talks about using "extends" and also about the fact that these are literally javascript files, so you might be able to work some magic related to that: https://eslint.org/docs/developer-guide/shareable-configs

ejizba (approved):
🎉

PR #1962, opened 2021-02-11
Localize package.json

nturinski:
Localize the commands and settings in package.json


package.nls.json
```diff
@@ -1,4 +1,67 @@
 {
+    "appService.description": "An Azure App Service management extension for Visual Studio Code.",
```
ejizba, line 2: Can you sort all of these please

ejizba (approved):

package.nls.json
nturinski: Just alphabetically?

ejizba: Yup

PR #1983, opened 2021-02-26
Add experimentation service

ejizba:
AFAIK, this is all we need to do an A/A test. cc @fiveisprime @bwateratmsft to confirm

Depends on https://github.com/microsoft/vscode-azuretools/pull/855


fiveisprime:
Yeah, this is pretty much what I did when I tested in SWA to verify the data was coming through.

bwateratmsft:
I assume you've set up an A/A test in the experimentation portal, targeting `ms-azuretools.vscode-azureappservice`, right?

nturinski (approved):

PR #1987, opened 2021-03-02
Handle mismatched sku tiers for basic create

ejizba:
Related to https://github.com/microsoft/vscode-azureappservice/pull/1984

### DIFF
--- a/src/commands/createWebApp/setPostPromptDefaults.ts
+++ b/src/commands/createWebApp/setPostPromptDefaults.ts
@@ -17,6 +17,7 @@ import { IWebAppWizardContext } from './IWebAppWizardContext';
 import { AzConfig, AzConfigProperty, readAzConfig } from "./readAzConfig";
 
 const maxNumberOfSites: number = 3;
+const freeTier = 'free';
 
 export async function setPostPromptDefaults(wizardContext: IWebAppWizardContext, siteNameStep: SiteNameStep): Promise<void> {
     // Reading az config should always happen after prompting because it can cause a few seconds delay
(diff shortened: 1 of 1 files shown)


src/commands/createWebApp/setPostPromptDefaults.ts
```diff
@@ -27,21 +28,27 @@ export async function setPostPromptDefaults(wizardContext: IWebAppWizardContext,
...
     const defaultPlanName: string = defaultName;
 
     const client: WebSiteManagementClient = await createWebSiteClient(wizardContext);
     const asp: WebSiteManagementModels.AppServicePlan | undefined = await tryGetAppServicePlan(client, defaultGroupName, defaultPlanName);
-    if (asp && checkPlanForPerformanceDrop(asp)) {
+    const hasPerfDrop = checkPlanForPerformanceDrop(asp);
+    if (asp && (hasPerfDrop || !matchesTier(asp, newSkuTier))) {
         // Subscriptions can only have 1 free tier Linux plan so show a warning if there are too many apps on the plan
-        if (wizardContext.newSiteOS === WebsiteOS.linux) {
+        if (wizardContext.newSiteOS === WebsiteOS.linux && newSkuTier === freeTier && hasPerfDrop) {
```
nturinski, line 46: Shouldn't the `newSkuTier === basicTier` as well?  Though I'm not sure why you need to check `newSkuTier === freeTier` here because `hasPerfDrop` should have covered that already.

nturinski (approved):
I'm pretty sure that this should work.   relatively low to me.  Since the asp names are kind of bizarre, there is probably only us and the CLI that uses this name.  The tier not matching seems like something that has a relatively low chance of happening in the first place.  But assuming the logic was correct in the first place, the check you added for matching tiers should handle any mismatching sku tiers.

src/commands/createWebApp/setPostPromptDefaults.ts
ejizba: See the comment right above this line (which I'm pretty sure you wrote?):
> // Subscriptions can only have 1 free tier Linux plan so show a warning if there are too many apps on the plan

This is a unique case because linux free only supports 1 plan per subscription. Users can have multiple linux basic plans, so we want to go to the else case for linux basic.

PR #2002, opened 2021-03-17
Cover more location and pricing tier in test cases

v-wuzhai:


test/nightly/getRotatingValue.ts
```diff
@@ -0,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+let locationCount: number = -1;
+const locationItem: string[] = ['Australia East', 'Australia Southeast', 'Brazil South', 'Canada Central', 'Central US', 'East Asia', 'East US', 'East US 2', 'France Central', 'Japan East', 'Japan West', 'Korea Central', 'Korea South', 'North Europe', 'South Central US', 'Southeast Asia', 'UK South', 'West Europe', 'West US', 'West US 2'];
```
ejizba, line 7: This still needs to be closer to what's in https://github.com/microsoft/vscode-azurefunctions/blob/main/test/nightly/getRotatingValue.ts

For example, call it `getRotationLocation` instead of `getLocation` and use the same logic with `getStartingIndex` instead of the `new Date()` stuff

test/nightly/getRotatingValue.ts
```diff
@@ -0,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+let locationCount: number = getStartingIndex();
+const locations: string[] = ['Australia East', 'Australia Southeast', 'Brazil South', 'Canada Central', 'Central US', 'East Asia', 'East US', 'East US 2', 'France Central', 'Japan East', 'Japan West', 'Korea Central', 'Korea South', 'North Europe', 'South Central US', 'Southeast Asia', 'UK South', 'West Europe', 'West US', 'West US 2'];
```
ejizba, line 7: I would shorten this list a bit - we want a big enough list to know we work in multiple regions, but we don't need to test every region. Sometimes lesser-used regions might get a feature later than others or sometimes a region could just have an outage. The platform team owns that and it's on them to test every region, not us.

With a smaller list, it's easier for us to debug issues if there's a failure - aka it's easier for us to know if the issue is on our side or the platform team's side.

test/nightly/getRotatingValue.ts
```diff
@@ -0,0 +1,31 @@
...
+const pricingTiers: string[] = ['P1v2', 'P2v2', 'P3v2', 'B1', 'B2', 'B3', 'S1', 'S2', 'S3'];
+export function getRotatingPricingTier(): string {
+    pricingTierCount += 1;
+    return pricingTiers[pricingTierCount % pricingTiers.length];
+}
+
+/**
+ * Adds a little more spice to the rotation
+ */
+function getStartingIndex(): number {
```
ejizba, line 23: If you need to do `0,5,10` I feel like that's a sign the lists above are too long. We need to strike a balance between testing every possible combination vs making the tests reliable/easy to debug. Right now I think it's too skewed towards the former. Please stick with `0,1,2` and make the lists a bit shorter

test/nightly/getRotatingValue.ts
v-wuzhai: @ejizba Do you have any suggestions for the locations we need to cover? Or we can have the same choice as https://github.com/microsoft/vscode-azurefunctions/blob/8525f5e03ac943a79f5b1918f33d9f0ef3aeb519/test/nightly/getRotatingValue.ts#L14??

ejizba: Yeah the same locations would be fine.

ejizba (approved):

PR #2009, opened 2021-03-31
"Upload Local Settings" added to "postDeploy" prompt

### LINKED ISSUE #1494: Add deployment of application settings when deploying the app
**Scenario:**

1. Create a new Python/Flask web app with some app settings in a `.env` file
1. Deploy this web app to Azure using the app service extension
1. View the web page

*Expected*

The app should load

*Actual*

The web app will give an application error

**Explanation**

The reason for the application error is because the app is deployed, but the settings in the `.env` file are not automatically deployed, they have to be deployed manually, then wait for the web app to pick them up.

This is not obvious, and has caused a lot of confusion especially with beginners using VS Code to build Flask apps at events like student hackathons. At hackathons there is often a push to use AI services that rely on API keys that get set in `.env` files, so this problem gets hit a lot.

**Suggested enhancement**

It would be good to have an option when deploying to upload any app settings that do not already exist on the app, with a note to say if they are not uploaded then they must be set by hand.

neelip:
Fixes #1494


src/commands/deploy/showDeployCompletedMessage.ts
```diff
@@ -9,25 +9,51 @@ import { AppServiceDialogResponses } from "../../constants";
...
-        await callWithTelemetryAndErrorHandling('postDeploy', async (context: IActionContext) => {
-            context.telemetry.properties.dialogResult = result?.title;
-            if (result === AppServiceDialogResponses.viewOutput) {
-                ext.outputChannel.show();
-            } else if (result === browseWebsiteBtn) {
-                await node.browse();
-            } else if (result === streamLogs) {
-                await startStreamingLogs(context, node);
-            }
+    const filesByFileExt = await workspaceUtil.findFilesByFileExtension(undefined, "env");
```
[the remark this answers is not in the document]
nturinski: Yeah, we don't have logic to find the .env file, we just assume the `envFileName` to be `.env`.  We have logic in `uploadAppSettings.ts` to let users select which .env file they want, but by default it just shows the `{workspaceFolder}/.env`.

Based off of this https://gist.github.com/ericelliott/4152984 we may be able to consider .env and extension since they're using production.env, dev.env, staging.env as examples.

neelip:
The latest changes for the "Upload Local Settings": 
- This button replaces "View Output" button 
- It is always displayed with a deploy prompt (along with the other 2 buttons)
- It always opens the "selectWorkspaceFile" prompt not depending on whether a workspace contains ".env" file
   (similar behavior to "Upload Local Settings" in the context menu)

@fiveisprime Hey Matt, could you please review & confirm this design for the upcoming release? 

- [ ] Review & Confirm Design

src/commands/deploy/showDeployCompletedMessage.ts
```diff
@@ -9,16 +9,21 @@ import { AppServiceDialogResponses } from "../../constants";
 import { SiteTreeItem } from "../../explorer/SiteTreeItem";
 import { ext } from "../../extensionVariables";
 import { localize } from "../../localize";
+import { uploadAppSettings } from "../appSettings/uploadAppSettings";
 import { startStreamingLogs } from '../logstream/startStreamingLogs';
 
 export function showDeployCompletedMessage(node: SiteTreeItem): void {
-    const message: string = localize('deployCompleted', 'Deployment to "{0}" completed.', node.client.fullName);
+    const message: string = localize('deployCompleted', 'Deployment to "{0}" completed. ([View Output](command:{1}))', node.client.fullName, ext.prefix + '.showOutputChannel');
```
ejizba, line 16: How come this was added? I don't see it on the design comment above. It's basically a fourth button, and if we're not okay with a fourth button, I don't think we should be okay with a link here either.

src/commands/deploy/showDeployCompletedMessage.ts
```diff
@@ -9,16 +9,21 @@ import { AppServiceDialogResponses } from "../../constants";
...
     ext.outputChannel.appendLog(message);
     const browseWebsiteBtn: MessageItem = { title: localize('browseWebsite', 'Browse Website') };
     const streamLogs: MessageItem = { title: localize('streamLogs', 'Stream Logs') };
+    const uploadSettingsBtn: MessageItem = { title: localize('uploadMessage', 'Upload Local Settings') };
+    const buttons: MessageItem[] = [browseWebsiteBtn, streamLogs, uploadSettingsBtn];
+
+    let envPath: string;
```
ejizba, line 23: I don't see `envPath` used anymore.

src/commands/deploy/showDeployCompletedMessage.ts
neelip: As discussed offline, removing "View Output" completely from the post-deploy message. 
- "View Output" in this post-deploy scenario had low usage data and "View Output" is already available during deployment and with the deployment failed message.

src/commands/deploy/showDeployCompletedMessage.ts
```diff
@@ -9,16 +9,20 @@ import { AppServiceDialogResponses } from "../../constants";
...
 import { localize } from "../../localize";
+import { uploadAppSettings } from "../appSettings/uploadAppSettings";
 import { startStreamingLogs } from '../logstream/startStreamingLogs';
 
 export function showDeployCompletedMessage(node: SiteTreeItem): void {
     const message: string = localize('deployCompleted', 'Deployment to "{0}" completed.', node.client.fullName);
     ext.outputChannel.appendLog(message);
     const browseWebsiteBtn: MessageItem = { title: localize('browseWebsite', 'Browse Website') };
     const streamLogs: MessageItem = { title: localize('streamLogs', 'Stream Logs') };
+    const uploadSettingsBtn: MessageItem = { title: localize('uploadMessage', 'Upload Local Settings') };
```
ejizba, line 20: I think @fiveisprime said "Upload Settings". It's shorter and still makes sense

ejizba (approved):

PR #2019, opened 2021-04-15
Fix AuthorizationFailed Error with App Service - Basic Create

### LINKED ISSUE #1490: Create web app fails when access is restricted to plan
Does this occur consistently? Yes
Repro steps:
1. Sign into the azure extension
2. Try and deploy a new app service

Action: appService.CreateWebApp
Error type: AuthorizationFailed
Error Message: The client 'xxxxxx@xxx.com' with object id 'XXXXXXXXXXXXXXXX' does not have authorization to perform action 'Microsoft.Web/serverfarms/read' over scope '/subscriptions/XXXXXXX/resourceGroups/appsvc_linux_centralus/providers/Microsoft.Web/serverfarms/appsvc_linux_centralus' or the scope is invalid. If access was recently granted, please refresh your credentials.

Version: 0.16.5
OS: darwin
OS Release: 19.4.0
Product: Visual Studio Code
Product Version: 1.44.1
Language: en

<details>
<summary>Call Stack</summary>

```
extension.bundle.js:225:3585174
retryCallback extension.bundle.js:225:116429
retryCallback extension.bundle.js:157:31051
extension.bundle.js:157:32664
handleRedirect extension.bundle.js:171:7747
extension.bundle.js:157:29836
Request._callback extension.bundle.js:1:582086
Request.r._callback.r.callback.r.callback extension.bundle.js:201:4468
Request.emit events.js:203:13
Request.<anonymous> extension.bundle.js:201:17577
Request.emit events.js:208:15
IncomingMessage.<anonymous> extension.bundle.js:201:16367
Object.onceWrapper events.js:291:20
IncomingMessage.emit events.js:208:15
endReadableNT _stream_readable.js:1168:12
processTicksAndRejections task_queues.js:77:11
```

</details>

neelip:


neelip:
Current commit: 
1. Start with "Create New Web App" 
2. If we get this error
image
Display this prompt (draft): 
image

What do all of you think about this approach?
@fiveisprime How do you like to see the prompt and the button label?

src/commands/createWebApp/createWebApp.ts
```diff
@@ -16,14 +16,15 @@ export async function createWebApp(context: IActionContext & Partial<ICreateChil
         node = <AzureParentTreeItem>await ext.tree.showTreeItemPicker(SubscriptionTreeItem.contextValue, context);
     }
     let newSite: WebAppTreeItem;
+    const authErrorRegexp = new RegExp(`The client '.+' with object id '.+' does not have authorization to perform action 'Microsoft\.Web\/serverfarms\/read' over scope '.+' or the scope is invalid\. If access was recently granted, please refresh your credentials\.`);
```
neelip, line 19: Is this is a good approach?

I found this specific error to be present many times with the same wording. 
image
Since this is the case, would shortening the regex be a good idea?

neelip, line 19: Should we shorten this regex or is it better to use the full string?

src/commands/createWebApp/createWebApp.ts
```diff
@@ -3,24 +3,47 @@
...
+        }
+        return newSite;
+    } catch (error) {
+        if (!parseError(error).message.match(authErrorRegexp)) {
+            throw error;
+        } else {
+            const message = localize('notAuthorizedToCreateRGs', 'You subscription is not authorized to create resource groups. Please create using Advanced Create to select existing resource groups.')
+            const advancedCreate: MessageItem = { title: localize('createAdvanced', 'Create - Advanced') };
+            // don't wait
+            void window.showInformationMessage(message, advancedCreate).then(async (result: MessageItem | undefined) => {
```
ejizba, line 33: We should be able to come up with a solution that keeps the user in basic create. For example, we're generating the plan name ourselves, so maybe we just generate a new name that doesn't conflict with the existing plan they don't have access to.

I think your first priority should be to figure out how to recreate the exact error yourself. You might need someone else to give you partial access to some resources on their subscription. That should help you pinpoint where the actual error is occurring and how to fix it within the basic create wizard.

neelip: I have been testing this scenario using Sandbox accounts and narrowed down the issue to be happening at [`tryGetAppServicePlan(...)`](https://github.com/microsoft/vscode-azureappservice/blob/8b750a07d0f741982f475ac02ca3f98fff4954b4/src/commands/createWebApp/setPostPromptDefaults.ts#L42).

I haven't tested scenarios other than Sandbox. I will check if those account scenarios lead to the same breakpoint.

src/commands/createWebApp/createWebApp.ts
```diff
@@ -3,24 +3,47 @@
...
+        newSite = <WebAppTreeItem>await node.createChild(context);
+        if (!suppressCreatedWebAppMessage) {
+            showCreatedWebAppMessage(newSite);
+        }
+        return newSite;
+    } catch (error) {
+        if (!parseError(error).message.match(authErrorRegexp)) {
+            throw error;
+        } else {
+            const message = localize('notAuthorizedToCreateRGs', 'You subscription is not authorized to create resource groups. Please create using Advanced Create to select existing resource groups.')
```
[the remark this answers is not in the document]
neelip: That's true, it is not the resources group but authorization failing with accessing the plan.

neelip:
Updated this with the following fix --> https://github.com/microsoft/vscode-azuretools/pull/888
This PR is independent from the above PR. We can close this PR if the above fix is more suitable.

ejizba:
@neelip It's unclear to me if this PR is still needed. If not, please close it in favor of the tools PR. If yes, please add repro steps and new expected behavior

src/commands/createWebApp/setPostPromptDefaults.ts
```diff
@@ -39,48 +39,74 @@ export async function setPostPromptDefaults(wizardContext: IWebAppWizardContext,
...
+                }
             }
+        } else {
+            wizardContext.newResourceGroupName = defaultGroupName;
+            wizardContext.newPlanName = defaultPlanName;
+        }
+    } catch (e) {
+        if (parseError(e).errorType === 'AuthorizationFailed') {
+            wizardContext.newResourceGroupName = defaultGroupName;
+            wizardContext.newPlanName = defaultPlanName;
```
neelip, line 89: These changes are needed to catch this error and set these values to some arbitrary values so that they are not undefined and continue with the rest of the code as is. 
Undefined values for `newResourceGroupName` and `newPlanName` is not expected with ResourceGroupCreateStep, which will throw an error.

src/commands/createWebApp/setPostPromptDefaults.ts
```diff
@@ -39,48 +39,74 @@ export async function setPostPromptDefaults(wizardContext: IWebAppWizardContext,
...
+            wizardContext.planSkuFamilyFilter = /^F$/i;
+            break;
+        case 'B':
+            wizardContext.planSkuFamilyFilter = /^B$/i;
+            break;
+        case 'PV2':
+            wizardContext.planSkuFamilyFilter = /^PV2$/i;
+            break;
+        default:
+            break;
```
neelip, line 109: This part is for when  the user doesn't have permission to create a App Service Plan and they are prompted to select from existing. 
By setting the `planSkuFamilyFilter`, only plans matching the user selected tier would be displayed for them.

neelip:
The changes in this PR are connected with https://github.com/microsoft/vscode-azuretools/pull/888

src/commands/createWebApp/setPostPromptDefaults.ts
neelip: Update: removing this change for this PR as it is not related to #1490.

neelip:
[Repro Steps](https://github.com/microsoft/vscode-azureappservice/issues/1490#issuecomment-840902910) for the AuthorizationFailed error with Basic Create

src/commands/createWebApp/setPostPromptDefaults.ts
```diff
@@ -37,49 +37,60 @@ export async function setPostPromptDefaults(wizardContext: IWebAppWizardContext,
...
+                    throw new Error(localize('noUniqnueNameAsp', 'Failed to generate unique name for app service plan. Use advanced creation to manually enter plan names.'));
+                }
             }
+        } else {
+            wizardContext.newResourceGroupName = defaultGroupName;
+            wizardContext.newPlanName = defaultPlanName;
+        }
+    } catch (error) {
+        if (parseError(error).errorType === 'AuthorizationFailed') {
+            // setting these values to some arbitrary values as ResourceGroupCreateStep & AppServicePlanCreateStep expects these to be not null
```
ejizba, line 87: Here's what I would say:
```suggestion
            // User has restricted permissions for something. We'll use the default values for now and let the  `ResourceGroupCreateStep` & `AppServicePlanCreateStep` handle the 'AuthorizationFailed' error down the line
```

ejizba (approved):

PR #2024, opened 2021-04-23
Update defaults with Create and Enable SCM_DO_BUILD... by default with Deploy

### LINKED ISSUE #1357: Update defaults when creating new apps

neelip:
Fixes #1357

### DIFF
--- a/src/commands/createWebApp/WebAppCreateStep.ts
+++ b/src/commands/createWebApp/WebAppCreateStep.ts
@@ -96,27 +96,12 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
                 default:
             }
         }
-        newSiteConfig.virtualApplications?.push({
-            virtualPath: '/',
-            physicalPath: 'site\\wwwroot',
-            preloadEnabled: true
-        })
-        // required property for setting "preloadEnabled" in VirtualApplications "true"
-        // "alwaysOn" cannot be true for "Free" tier
-        if (context?.plan?.sku?.tier !== 'Free') {
-            newSiteConfig.alwaysOn = true;
-        }
         return newSiteConfig;
     }
 
     private getAppSettings(context: IWebAppWizardContext): WebSiteManagementModels.NameValuePair[] {
         const appSettings: WebSiteManagementModels.NameValuePair[] = [];
         const disabled: string = 'disabled';
-        const trueString: string = 'true';
-        appSettings.push({
-            name: 'SCM_DO_BUILD_DURING_DEPLOYMENT',
-            value: trueString
-        });
 
         if (context.appInsightsComponent) {
             appSettings.push({
@@ -162,7 +147,7 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
             } else {
                 appSettings.push({
                     name: 'APPLICATIONINSIGHTSAGENT_EXTENSION_ENABLED',
-                    value: trueString
+                    value: 'true'
                 });
             }
         }
--- a/src/commands/createWebApp/WebAppCreateStep.ts
+++ b/src/commands/createWebApp/WebAppCreateStep.ts
@@ -12,7 +12,6 @@ import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
 import { createWebSiteClient } from '../../utils/azureClients';
 import { nonNullProp } from '../../utils/nonNull';
-import { updateWorkspaceSetting } from '../../vsCodeConfig/settings';
 import { FullJavaStack, FullWebAppStack, IWebAppWizardContext } from './IWebAppWizardContext';
 import { getJavaLinuxRuntime } from './stacks/getJavaLinuxRuntime';
 import { WindowsJavaContainerSettings } from './stacks/models/WebAppStackModel';
@@ -52,8 +51,6 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
             siteConfig: newSiteConfig,
             reserved: context.newSiteOS === WebsiteOS.linux  // The secret property - must be set to true to make it a Linux plan. Confirmed by the team who owns this API.
         });
-        // Sets "showBuildDuringDeployPrompt" to false --> SCM_DO_BUILD_DURING_DEPLOYMENT(set to true) is added as a appSetting by default for new web apps
-        await updateWorkspaceSetting(constants.configurationSettings.showBuildDuringDeployPrompt, false, undefined);
     }
 
     public shouldExecute(context: IWebAppWizardContext): boolean {
--- a/src/commands/createWebApp/WebAppCreateStep.ts
+++ b/src/commands/createWebApp/WebAppCreateStep.ts
@@ -6,7 +6,7 @@
 import { WebSiteManagementClient, WebSiteManagementModels } from '@azure/arm-appservice';
 import { Progress } from 'vscode';
 import { WebsiteOS } from 'vscode-azureappservice';
-import { AzureWizardExecuteStep } from 'vscode-azureextensionui';
+import { AzureWizardExecuteStep, LocationListStep } from 'vscode-azureextensionui';
 import * as constants from '../../constants';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
@@ -38,13 +38,14 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
 
         const siteName: string = nonNullProp(context, 'newSiteName');
         const rgName: string = nonNullProp(nonNullProp(context, 'resourceGroup'), 'name');
-        const locationName: string = nonNullProp(nonNullProp(context, 'location'), 'name');
+        const location = await LocationListStep.getLocation(context, constants.webProvider);
+        const locationName: string = nonNullProp(location, 'name');
 
         const newSiteConfig: WebSiteManagementModels.SiteConfig = this.getSiteConfig(context);
         const client: WebSiteManagementClient = await createWebSiteClient(context);
         context.site = await client.webApps.createOrUpdate(rgName, siteName, {
             name: context.newSiteName,
-            kind: context.newSiteKind,
+            kind: context.customLocation ? 'linux,kubernetes' : context.newSiteKind,
             location: locationName,
             serverFarmId: context.plan && context.plan.id,
             clientAffinityEnabled: true,
--- a/src/commands/createWebApp/WebAppCreateStep.ts
+++ b/src/commands/createWebApp/WebAppCreateStep.ts
@@ -96,20 +96,29 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
                 default:
             }
         }
-        newSiteConfig.httpLoggingEnabled = true;
         newSiteConfig.virtualApplications?.push({
             virtualPath: '/',
             physicalPath: 'site\\wwwroot',
             preloadEnabled: true
         })
-        newSiteConfig.logsDirectorySizeLimit = 100;
-        newSiteConfig.alwaysOn = true;
+        // required property for setting "preloadEnabled" in VirtualApplications "true"
+        // "alwaysOn" cannot be true for "Free" tier
+        console.log(context?.plan?.sku?.tier);
+        if (context?.plan?.sku?.tier !== 'Free') {
+            newSiteConfig.alwaysOn = true;
+        }
         return newSiteConfig;
     }
 
     private getAppSettings(context: IWebAppWizardContext): WebSiteManagementModels.NameValuePair[] {
         const appSettings: WebSiteManagementModels.NameValuePair[] = [];
         const disabled: string = 'disabled';
+        const trueString: string = 'true';
+
+        appSettings.push({
+            name: 'SCM_DO_BUILD_DURING_DEPLOYMENT',
+            value: trueString
+        });
 
         if (context.appInsightsComponent) {
             appSettings.push({
@@ -155,7 +164,7 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
             } else {
                 appSettings.push({
                     name: 'APPLICATIONINSIGHTSAGENT_EXTENSION_ENABLED',
-                    value: 'true'
+                    value: trueString
                 });
             }
         }
--- a/src/commands/createWebApp/WebAppCreateStep.ts
+++ b/src/commands/createWebApp/WebAppCreateStep.ts
@@ -96,6 +96,14 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
                 default:
             }
         }
+        newSiteConfig.httpLoggingEnabled = true;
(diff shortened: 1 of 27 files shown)


src/commands/createWebApp/WebAppCreateStep.ts
```diff
@@ -96,6 +96,14 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
...
             }
         }
+        newSiteConfig.httpLoggingEnabled = true;
+        newSiteConfig.virtualApplications?.push({
+            virtualPath: '/',
+            physicalPath: 'site\\wwwroot',
+            preloadEnabled: true
+        })
+        newSiteConfig.logsDirectorySizeLimit = 100;
+        newSiteConfig.alwaysOn = true;
```
neelip, line 106: "alwaysOn" needs to be true for setting "preloadEnabled" true in "virtualApplications".

src/commands/deploy/deploy.ts
```diff
@@ -66,9 +66,9 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
             const linuxFxVersion: string = siteConfig.linuxFxVersion.toLowerCase();
             if (linuxFxVersion.startsWith(LinuxRuntimes.node)) {
                 // if it is node or python, prompt the user (as we can break them)
-                await promptScmDoBuildDeploy(context, context.effectiveDeployFsPath, LinuxRuntimes.node);
+                await enableScmDoBuildDuringDeploy(context.effectiveDeployFsPath, LinuxRuntimes.node);
```
neelip, line 69: Removed the prompt for SCM_DO_BUILD_DEPLOY. SCM_DO_BUILD_DEPLOY will be enabled by default when an app is deployed.

fiveisprime: I'd like this to also set an app setting rather than create the .deployment file with the setting.

ejizba: The linked issue is only about updating the defaults for creating new apps, not changing existing apps.

Is there an open issue to change the behavior for deploying existing apps? If not, @fiveisprime can you create one and @neelip can we do that as a separate PR? I have some concerns with that and I think GitHub would be the best place to have that discussion.

ejizba (changes requested):

nturinski:
I believe we discussed that we only want to enable the setting on web app creation rather than on every deploy.  

If users already have a working project, we don't want to risk breaking their existing projects by changing an app setting when they deploy.

src/commands/deploy/deploy.ts
nturinski: I believe we discussed that we only want to enable the setting on web app creation rather than on every deploy.

If users already have a working project, we don't want to risk breaking their existing projects by changing an app setting when they deploy.

fiveisprime: Correct. We don't want to make that kind of change to existing apps on deployment, it doesn't feel right to me. I would be okay with prompting if it's not set and asking them if we should set it for them, but never automatically.

neelip: If we are planning to enable SCM_DO_BUILD_DEPLOY by default with `Create` in this PR without changing existing behavior we won't be able to replace ".deployment" file with appSetting for SCM_DO... @fiveisprime is that ok? 

I think we need to change the existing behavior in order to replace the ".deployment" file with a appSetting for SCM_DO..., which I agree needs a new PR.

src/commands/deploy/deploy.ts
```diff
@@ -70,7 +71,8 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
             } else if (linuxFxVersion.startsWith(LinuxRuntimes.python)) {
                 await promptScmDoBuildDeploy(context, context.effectiveDeployFsPath, LinuxRuntimes.python);
             }
-
+        } else {
+            await fse.writeFile(path.join(context.effectiveDeployFsPath, constants.deploymentFileName), constants.deploymentFile);
```
ejizba, line 75: As discussed, the only changes you're doing should be a part of "create", not "deploy". You should be able to leave ".deployment" behavior alone.

src/explorer/SiteTreeItem.ts
```diff
@@ -138,18 +138,23 @@ export abstract class SiteTreeItem extends SiteTreeItemBase implements ISiteTree
         await deleteSite(this.root.client);
     }
 
-    public async isHttpLogsEnabled(): Promise<boolean> {
+    public async isLogsEnabled(): Promise<boolean> {
         const logsConfig: WebSiteManagementModels.SiteLogsConfig = await this.root.client.getLogsConfig();
-        return !!(logsConfig.httpLogs && logsConfig.httpLogs.fileSystem && logsConfig.httpLogs.fileSystem.enabled);
+        return !!(logsConfig.httpLogs?.fileSystem?.enabled && logsConfig.applicationLogs?.fileSystem?.level === 'Verbose');
```
ejizba, line 143: I don't think we should be checking that it's "Verbose" here. As long as logging enabled, we should stream logs as-is and users can do whatever log level they want

src/commands/createWebApp/WebAppCreateStep.ts
```diff
@@ -96,12 +96,27 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
...
+        }
         return newSiteConfig;
     }
 
     private getAppSettings(context: IWebAppWizardContext): WebSiteManagementModels.NameValuePair[] {
         const appSettings: WebSiteManagementModels.NameValuePair[] = [];
         const disabled: string = 'disabled';
+        const trueString: string = 'true';
+        appSettings.push({
+            name: 'SCM_DO_BUILD_DURING_DEPLOYMENT',
```
ejizba, line 117: Please add a note in the changelog in the Unreleased/Changed section about this.

src/commands/createWebApp/WebAppCreateStep.ts
```diff
@@ -96,12 +96,27 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
...
         }
+        newSiteConfig.virtualApplications?.push({
+            virtualPath: '/',
+            physicalPath: 'site\\wwwroot',
+            preloadEnabled: true
+        })
+        // required property for setting "preloadEnabled" in VirtualApplications "true"
+        // "alwaysOn" cannot be true for "Free" tier
+        if (context?.plan?.sku?.tier !== 'Free') {
+            newSiteConfig.alwaysOn = true;
```
ejizba, line 107: Can I get more detail on "preloadEnabled" and "alwaysOn"? Without being super familiar with those settings, it sounds like it'd be a decent amount more expensive if my site is always running.

src/commands/deploy/deploy.ts
neelip: We won't be able to enable SCM_DO_BUILD_DEPLOY by default at Create as it involves writing a '.deployment' file in the workspace directory and we can't confirm if the directory open in the workspace is the project they want to deploy. 
This is only possible when the "Create Web App..." is run from the workspace folder menu.  
Is there a possibility that I might be missing?

ejizba: `SCM_DO_BUILD_DURING_DEPLOYMENT` as an app setting and `SCM_DO_BUILD_DURING_DEPLOYMENT` in the `.deployment` file do the same thing. You only need one, not both. You're already doing the app setting on create, right? So you should be good

src/commands/createWebApp/WebAppCreateStep.ts
neelip: From what I learned, both preloading and "alwaysOn" go together, and they are properties that enable faster loading times of the web app by keep the site/server alive even without usage. 
I am not sure about the specifics for Azure but I agree that it doesn't seem like a good idea to set them to `true` by default. 
Both "preloadEnabled" and "alwaysOn" seem to better fit frequent usage/production-level usage scenarios.

src/commands/deploy/deploy.ts
neelip: I don't know if we should set `SCM_DO_BUILD_DEPLOYMENT` to **true** at Create without deciding on the behavior for deployment. 
The problem comes up at below prompt if a user doesn't select 'Yes' but `SCM_DO_BUILD_DEPLOYMENT` is set **true**. 
image
In this case, build will happen even when the user selected 'No'. 
One option is to set `SCM_DO_BUILD_DEPLOYMENT` to **false** when user selects 'No' at this prompt. But this wouldn't necessarily solve anything as it is redundant behavior. 

Also, when I am testing this, I have noticed that `SCM_DO_BUILD_DEPLOYMENT` seems to only matter for Linux web apps and not for Windows. Windows deployments are successful even without the `SCM_DO_BUILD_DEPLOYMENT`.

ejizba: That prompt should _**not**_ be shown if the app they're deploying to already has `SCM_DO_BUILD_DEPLOYMENT` as an app setting. The point of that prompt is to encourage people to turn it on and in that case they already have it on.

src/commands/createWebApp/WebAppCreateStep.ts
neelip: Just talked to Matt offline, he is good with not setting these values.

src/commands/createWebApp/WebAppCreateStep.ts
```diff
@@ -49,6 +51,8 @@ export class WebAppCreateStep extends AzureWizardExecuteStep<IWebAppWizardContex
             siteConfig: newSiteConfig,
             reserved: context.newSiteOS === WebsiteOS.linux  // The secret property - must be set to true to make it a Linux plan. Confirmed by the team who owns this API.
         });
+        // Sets "showBuildDuringDeployPrompt" to false --> SCM_DO_BUILD_DURING_DEPLOYMENT(set to true) is added as a appSetting by default for new web apps
+        await updateWorkspaceSetting(constants.configurationSettings.showBuildDuringDeployPrompt, false, undefined);
```
ejizba, line 55: This doesn't really fix the root issue. Ideally our logic that prompts about `SCM_DO_BUILD_DURING_DEPLOYMENT` should check the app settings at that time. This doesn't handle people that create an app or change the app setting outside VS Code.

src/explorer/SubscriptionTreeItem.ts
```diff
@@ -125,6 +125,10 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         try {
             // enable HTTP logs by default
             await newSite.enableHttpLogs();
+            if (wizardContext.newSiteOS === WebsiteOS.windows) {
+                //enable Application logs by defaul (only for windows)
+                await newSite.enableApplicationLogs();
```
ejizba, line 130: Rather than make two methods, just add a flag to `enableHttpLogs` so that we only have to call `updateLogsConfig` (which actually sends the request) once

src/commands/deploy/deploy.ts
```diff
@@ -59,8 +59,11 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
     }
 
     const isZipDeploy: boolean = siteConfig.scmType !== constants.ScmType.LocalGit && siteConfig !== constants.ScmType.GitHub;
+    const appSettingsClient: IAppSettingsClient = node.client;
+    const remoteSettings: WebSiteManagementModels.StringDictionary = await appSettingsClient.listApplicationSettings();
+    const isBuildDuringDeploymentAppSetting: boolean = !!remoteSettings.properties && 'SCM_DO_BUILD_DURING_DEPLOYMENT' in remoteSettings.properties;
```
ejizba, line 64: How about "isBuildDuringDeploymentAppSetting" -> "hasScmDoBuildSetting"

"has" makes more sense than "is" and I think getting the first part of "SCM_DO_BUILD_DURING_DEPLOYMENT" in the name is more important than "DuringDeployment"

src/commands/deploy/deploy.ts
```diff
@@ -59,8 +59,11 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
     }
 
     const isZipDeploy: boolean = siteConfig.scmType !== constants.ScmType.LocalGit && siteConfig !== constants.ScmType.GitHub;
+    const appSettingsClient: IAppSettingsClient = node.client;
+    const remoteSettings: WebSiteManagementModels.StringDictionary = await appSettingsClient.listApplicationSettings();
```
ejizba, line 63:
```suggestion
    const remoteSettings: WebSiteManagementModels.StringDictionary = await node.client.listApplicationSettings();
```

src/commands/deploy/deploy.ts
```diff
@@ -59,8 +59,11 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
...
+    const isBuildDuringDeploymentAppSetting: boolean = !!remoteSettings.properties && 'SCM_DO_BUILD_DURING_DEPLOYMENT' in remoteSettings.properties;
     // only check enableScmDoBuildDuringDeploy if currentWorkspace matches the workspace being deployed as a user can "Browse" to a different project
-    if (getWorkspaceSetting<boolean>(constants.configurationSettings.showBuildDuringDeployPrompt, context.effectiveDeployFsPath)) {
+    if (!isBuildDuringDeploymentAppSetting && getWorkspaceSetting<boolean>(constants.configurationSettings.showBuildDuringDeployPrompt, context.effectiveDeployFsPath)) {
```
ejizba, line 66: Would be better if you checked the app setting stuff within this first "if" so that people who turned the VS Code setting off never even have to make the http call to get app settings

src/explorer/SiteTreeItem.ts
```diff
@@ -143,17 +143,22 @@ export abstract class SiteTreeItem extends SiteTreeItemBase implements ISiteTree
         return !!(logsConfig.httpLogs && logsConfig.httpLogs.fileSystem && logsConfig.httpLogs.fileSystem.enabled);
     }
 
-    public async enableHttpLogs(): Promise<void> {
-        const logsConfig: WebSiteManagementModels.SiteLogsConfig = {
-            httpLogs: {
+    public async enableLogs(enableAppLogs: boolean): Promise<void> {
+        const logsConfig: WebSiteManagementModels.SiteLogsConfig = {};
+        if (enableAppLogs) {
```
ejizba, line 148: I don't get why one call to `enableLogs` checks the OS and one just passes `false`. 

Would it make sense to get rid of the flag and just check `!this.root.client.isLinux`?

neelip: The instance with `false` is in the deployment path, I wasn't sure if we wanted to modify the behavior there. 
That's a good idea, it will shorten the code.

ejizba (approved):

PR #2026, opened 2021-04-28
Add Subpath to SaveDeployDefaults prompt

### LINKED ISSUE #2007: [Suggestion] It would be better to display the selected project name accurately in the dialog box
**OS:** All
**Build Version:** 20210327.1

**Repro Steps:**

1. Create a folder -> Move multiple projects into it.
2. Open the folder in VS Code -> Navigate to the File Explorer -> Right click a project -> Select "Deploy to Web App..." -> Select a subscription -> Select a web app -> Pop up a dialog to ask whether always deploy workspace to the special web app.
3. Check the dialog box.

**Actual:**
Show the workspace name instead of the selected project name in the dialog box.

**Suggestion:**
It would be better to display the selected project name accurately.
image

neelip:
Fixes #2007

### DIFF
--- a/src/commands/deploy/promptToSaveDeployDefaults.ts
+++ b/src/commands/deploy/promptToSaveDeployDefaults.ts
@@ -15,7 +15,7 @@ export async function promptToSaveDeployDefaults(context: IActionContext, node: 
     const defaultWebAppToDeploySetting: string | undefined = getWorkspaceSetting(constants.configurationSettings.defaultWebAppToDeploy, workspacePath);
     // only prompt if setting is unset
     if (!defaultWebAppToDeploySetting) {
-        const saveDeploymentConfig: string = localize('showDeploymentConfig', 'Always deploy the workspace "{0}" to "{1}"?', path.relative(workspacePath, deployPath) || path.basename(workspacePath), node.root.client.fullName);
+        const saveDeploymentConfig: string = localize('showDeploymentConfig', 'Always deploy the workspace "{0}" to "{1}"?', path.relative(workspacePath, deployPath) ? deployPath : path.basename(workspacePath), node.root.client.fullName);
         const dontShowAgain: MessageItem = { title: localize('dontShow', "Don't show again") };
         const result: MessageItem = await context.ui.showWarningMessage(saveDeploymentConfig, DialogResponses.yes, dontShowAgain, DialogResponses.skipForNow);
         if (result === DialogResponses.yes) {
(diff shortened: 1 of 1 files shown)


src/commands/deploy/promptToSaveDeployDefaults.ts
```diff
@@ -15,7 +15,7 @@ export async function promptToSaveDeployDefaults(context: IActionContext, node:
     const defaultWebAppToDeploySetting: string | undefined = getWorkspaceSetting(constants.configurationSettings.defaultWebAppToDeploy, workspacePath);
     // only prompt if setting is unset
     if (!defaultWebAppToDeploySetting) {
-        const saveDeploymentConfig: string = localize('showDeploymentConfig', 'Always deploy the workspace "{0}" to "{1}"?', path.basename(workspacePath), node.root.client.fullName);
+        const saveDeploymentConfig: string = localize('showDeploymentConfig', 'Always deploy the workspace "{0}" to "{1}"?', path.relative(workspacePath, deployPath) || path.basename(workspacePath), node.root.client.fullName);
```
neelip, line 18: Would "workspace" still work here? or should we use something more like "directory/folder"?

nturinski: I think workspace is probably still fine wordage, but did we want it to only show the subfolder name?  I'm not sure how I feel about it, but it seems like having the full path could be more useful, but I understand the concern if the subfolder was nested pretty deeply.

neelip: I agree, full path does give more info especially for nested folders. 

But I think it makes sense to fix this wording issue as in the case of deploying a nested folder, it get's confusing to display the parent folder for the prompt. 
Also, the `saveDeployDefaults()` behavior doesn't match the wording, this fix will match the wording with the function's behavior.

src/commands/deploy/promptToSaveDeployDefaults.ts
```diff
@@ -15,7 +15,7 @@ export async function promptToSaveDeployDefaults(context: IActionContext, node:
     const defaultWebAppToDeploySetting: string | undefined = getWorkspaceSetting(constants.configurationSettings.defaultWebAppToDeploy, workspacePath);
     // only prompt if setting is unset
     if (!defaultWebAppToDeploySetting) {
-        const saveDeploymentConfig: string = localize('showDeploymentConfig', 'Always deploy the workspace "{0}" to "{1}"?', path.relative(workspacePath, deployPath) || path.basename(workspacePath), node.root.client.fullName);
+        const saveDeploymentConfig: string = localize('showDeploymentConfig', 'Always deploy the workspace "{0}" to "{1}"?', path.relative(workspacePath, deployPath) ? deployPath : path.basename(workspacePath), node.root.client.fullName);
```
neelip, line 18: With the update, the prompt will display full path for nested folders and just the workspace name in the other case.

nturinski (approved):

PR #2027, opened 2021-04-28
Move 'Deploy' context menu item up

ejizba:
Same idea as https://github.com/microsoft/vscode-azurefunctions/pull/2785


nturinski (approved):
Not gonna lie, it slightly bothers me that they are now out of order in the `package.json` though 😬

PR #2064, opened 2021-05-20
Fix nightly tests

ejizba:
Related to https://github.com/microsoft/vscode-azuretools/pull/920, the sku prompt is a bit different now

### DIFF
--- a/test/nightly/getRotatingValue.ts
+++ b/test/nightly/getRotatingValue.ts
@@ -11,8 +11,8 @@ export function getRotatingLocation(): string {
 }
 
 let pricingTierCount: number = getStartingIndex();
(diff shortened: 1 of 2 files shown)


test/nightly/getRotatingValue.ts
```diff
@@ -11,8 +11,8 @@ export function getRotatingLocation(): string {
 }
 
 let pricingTierCount: number = getStartingIndex();
-const pricingTiers: string[] = ['P1v2', 'P2v2', 'P3v2', 'B1', 'B2', 'B3', 'S1', 'S2', 'S3'];
-export function getRotatingPricingTier(): string {
+const pricingTiers: (string | RegExp)[] = [/P1v2/, 'P2v2', 'P3v2', /B1/, 'B2', 'B3', 'S1', 'S2', 'S3'];
```
neelip, line 14: Why only 'P1v2' and 'B1' are treated as regex?

neelip (approved):

PR #2077, opened 2021-06-07
allow user to deploy a maven module (by selecting a maven folder/pom.xml) to Azure App Service directly

wangmingliang-ms:
A preDeployTask `mvn package...` will be added to settings if user choose to deploy a maven module folder/pom.xml. This preDeployTask will compile the source code of the module and package it in its distributable format(jar, war, ear, ...).
The first detected war/executable jar(depending on the runtime of the target Azure appservice) file under `$module_folder/target/` is deployed to Azure.


src/commands/deploy/setPreDeployTaskForMavenModule.ts
```diff
@@ -0,0 +1,63 @@
...
+import { TaskDefinition } from 'vscode';
+import { IDeployContext } from 'vscode-azureappservice';
+import * as constants from '../../constants';
+import { javaUtils } from "../../utils/javaUtils";
+import { getWorkspaceSetting, updateWorkspaceSetting } from '../../vsCodeConfig/settings';
+import * as tasks from '../../vsCodeConfig/tasks';
+
+export async function setPreDeployTaskForMavenModule(context: IDeployContext, module: { pom: string, artifactId: string, packaging: string }): Promise<void> {
+    const preDeployTaskKey: string = 'preDeployTask';
+    const workspaceFspath: string = context.workspaceFolder.uri.fsPath;
```
nturinski, line 16: I think you should pass in the effectiveDeployFsPath or originalDeployFsPath rather than using the workspaceFolder.uri.fsPath here.  I ran into a bug where it was using my first workspace folder rather than the one I deployed.

Look below:

image

wangmingliang-ms: **This seems a bug of the `getDeployFsPath.getContainingWorkspace`** ?! should I fix it?

I tried with the same project and deploying `user-service` to Azure webapp and found that this problem occurs only when all these 3 conditions meets:
* `user-service` is a submodule of `java-microservice` and
* both `user-service` and `java-microservice` are added as root folders of your workspace and 
* `java-microservice` is before `user-service` in workspace.

image
image
the `workspaceFolder` becomes correct after changing the order of the root folders in workspace(you can drag the root folders to change their order):
image

**the `context.workspaceFolder` is copied from `deployPaths: IDeployPaths` and originally generated by `getDeployFsPath.getContainingWorkspace`, which iterates the root folders and find the first one(this is exactly why the order matters) containing the selected folder as `workspaceFolder`**

wangmingliang-ms: image

wangmingliang-ms: image

wangmingliang-ms: image

src/commands/deploy/deploy.ts
```diff
@@ -54,7 +55,17 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
...
 
-    if (javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
+    let javaMavenModule = undefined;
+    if (siteConfig && javaUtils.isJavaRuntime(siteConfig)) {
+        fileExtensions = fileExtensions || await javaUtils.getJavaFileExtensions(siteConfig);
+        javaMavenModule = javaUtils.getMavenModule(context.effectiveDeployFsPath);
+        // get packaging type as artifact file extension for maven module.
+        const ext = javaMavenModule?.packaging ?? context.effectiveDeployFsPath.split('.').pop() ?? '';
+        if (!fileExtensions?.includes(ext)) { // show error and break if deploying a raw folder/or non-matching artifact to java runtime instance.
+            void vscode.window.showErrorMessage(localize('errorArtifactsNotMatchingJava', 'Only "{0}" file/maven module can be deployed to "{1}"', `${fileExtensions}`, siteConfig.name));
```
ejizba, line 65: We don't want to block users from deploying like this. Once they've added settings like "preDeployTask" and/or "deploySubpath", they should have full control and we will just do what they specified in the settings.

src/commands/deploy/setPreDeployTaskForMavenModule.ts
ejizba: No, I don't think you need to fix this bug in your PR. Do we have an issue to track it? I would guess it's an uncommon edge case to have two workspace folders where one is a sub folder of another, so I'm not super worried about it.

src/utils/javaUtils.ts
```diff
@@ -110,14 +107,75 @@ export namespace javaUtils {
...
+            const potentialMvnwPath: string = path.join(current, mvnw);
+            if (await fse.pathExists(potentialMvnwPath)) {
+                return potentialMvnwPath;
+            }
+            current = path.dirname(current);
+        }
+        return undefined;
+    }
+
+    export async function isExecutableJarFile(fsPath: string): Promise<boolean> {
```
ejizba, line 163: Please avoid using a utils file when possible. We really only want to put code in the utils folder if it's shared by several other files, but I think a decent amount of this code is only used in one place.

package.json
```diff
@@ -808,7 +808,9 @@
         "@azure/arm-resources": "^4.0.0",
         "@azure/ms-rest-js": "^2.2.1",
         "dotenv": "^6.2.0",
+        "fast-xml-parser": "^3.19.0",
```
ejizba, line 811: Please see if you can avoid adding dependencies for your PR, as each dependency added can negatively affect activation performance. I'm also asking this because I believe the dependencies are only used once and Java is not a very high usage scenario for us

ejizba (changes requested):

package.json
wangmingliang-ms: >  as each dependency added can negatively affect activation performance

the size of `fast-xml-parser` is small, it doesn't even have any dependency for production(only devDependencies)
https://www.npmjs.com/package/fast-xml-parser
as far as I know, only the first level functions/classes will be loaded (but none of them will be called) on `import`/`require`.
if we implement the same logic by ourselves, we also need to make some functions/classes, they also need to be loaded first. (the only overhead of using external module is it may load some/a few number of un-required methods for us) I don't think we need to consider this performance degradation.

> I'm also asking this because I believe the dependencies are only used once

`only used once` is not a good reason for us to reinvent the wheel, 
and the logic of converting xml into json is actually not that simple.

> Java is not a very high usage scenario for us

### **Java is First Class Citizen on Azure**
**`low usage` is the start point and a good reason for us to improve the UX for Java developers but not to give up them.**
Java developers matter for Microsoft and for Azure especially when **the number for java vscode developers increase rapidly.**

src/utils/javaUtils.ts
wangmingliang-ms: It's OK for me to move this method somewhere else. 

IMO:
we should judge if or not a method/function should be put in a utils file/class based on if the method/function itself is a util or not.
**in other words, a util method/function should be put into utils file/class regardless of it's shared or not for current moment.** in this way, future developers can easily find it when they need.

src/commands/deploy/deploy.ts
wangmingliang-ms: OK, I will change this part to allow proceeding and show a warning instead of error.

but I doubt if it's a good UX to allow proceeding when we know it will certainly fail at last.

src/commands/deploy/setPreDeployTaskForMavenModule.ts
wangmingliang-ms: issue is reported here:
https://github.com/microsoft/vscode-azureappservice/issues/2085

src/commands/deploy/deploy.ts
```diff
@@ -54,7 +55,17 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
     // if we already got siteConfig, don't waste time getting it again
     siteConfig = siteConfig ? siteConfig : await node.root.client.getSiteConfig();
 
-    if (javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
+    let javaMavenModule = undefined;
+    if (siteConfig && javaUtils.isJavaRuntime(siteConfig)) {
+        fileExtensions = fileExtensions || await javaUtils.getJavaFileExtensions(siteConfig);
```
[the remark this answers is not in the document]
wangmingliang-ms: ok, I will refactor this part based on your update.

src/commands/deploy/deploy.ts
```diff
@@ -54,7 +55,17 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
     // if we already got siteConfig, don't waste time getting it again
     siteConfig = siteConfig ? siteConfig : await node.root.client.getSiteConfig();
 
-    if (javaUtils.isJavaRuntime(siteConfig.linuxFxVersion)) {
+    let javaMavenModule = undefined;
+    if (siteConfig && javaUtils.isJavaRuntime(siteConfig)) {
```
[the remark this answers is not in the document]
wangmingliang-ms: fixed.

src/commands/deploy/deploy.ts
```diff
@@ -90,7 +101,16 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
     }
 
     // only respect the deploySubpath settings for zipdeploys
-    const deployPath: string = isZipDeploy ? context.effectiveDeployFsPath : context.originalDeployFsPath;
+    let deployPath: string = isZipDeploy ? context.effectiveDeployFsPath : context.originalDeployFsPath;
+    if (javaMavenModule) { // modify deployPath to the executable jar or war files generated by preDeployTask when deploying maven module
```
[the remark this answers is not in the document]
wangmingliang-ms: maven is some different: the name of the finally deployed file can not be determined before `preDeployTask` is finished.
when user choose a pom.xml(which represents a maven module) to deploy. the `deploySubpath` would be the relative path to the pom file, but the `deployPath` should be the path to the generated war/jar file by `preDeployTask`.

to follow your current pattern, I added a `runPreDeployTask`(similar to existing `setPreDeployConfig`/`runPostDeployTask`), and modify the `context.effectiveDeployFsPath` to point to the generated war/jar file after `preDeployTask` is run.
```ts
export async function runPreDeployTask(... ): Promise<void> {
    await appservice.runPreDeployTask(context, context.originalDeployFsPath, siteConfig.scmType);

    // only executable jar or war files generated by `preDeployTask`can be deployed to java runtime appservice, while
    // the `context.effectiveDeployFsPath` points the path of the pom file/module folder when deploying a maven module.
    // we need to modify `context.effectiveDeployFsPath` to point the executable jar or war files, of which the path
    // can only be determined after preDeployTask is run.
    if (javaUtils.isJavaRuntime(siteConfig)) {
        // ...
        context.effectiveDeployFsPath = executableArtifacts?.[0] ?? '';
        // ...
    }

```

src/commands/deploy/deploy.ts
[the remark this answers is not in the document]
wangmingliang-ms: fixed.

src/utils/javaUtils.ts
wangmingliang-ms: fixed to your pattern

wangmingliang-ms:
check https://github.com/microsoft/vscode-azureappservice/pull/2097 for the new implementation

PR #2084, opened 2021-06-17
Split up setPreDeployConfigForDotnet

ejizba:
so that other languages can benefit from some shared code and more easily follow the same pattern. The idea is that a new language would add a few lines of code to the end of `setPreDeployConfig` and otherwise put the rest of their code in a language-specific folder like I did for "dotnet". Also I chose "setPreDeployConfig" over "setPreDeployTask" because we should be setting "deploySubpath" at the same time.

This is purely moving code, no functional changes.


nturinski (approved):

src/commands/deploy/dotnet/tryGetCsprojFile.ts
```diff
@@ -0,0 +1,38 @@
...
+    const projectFiles: string[] = await checkFolderForCsproj(projectPath);
+    // it's a common pattern to have the .csproj file in a subfolder so check one level deeper
+    if (projectFiles.length === 0) {
+        const subfolders: string[] = await fse.readdir(projectPath);
+        await Promise.all(subfolders.map(async folder => {
+            const filePath: string = path.join(projectPath, folder);
+            // check its existence as this will check .vscode even if the project doesn't contain that folder
+            if (await fse.pathExists(filePath) && (await fse.stat(filePath)).isDirectory()) {
+                projectFiles.push(...await checkFolderForCsproj(filePath));
+                context.telemetry.properties.csprojInSubfolder = 'true';
```
wangmingliang-ms, line 20: I think 
you should check if `projectFiles` is non-empty before setting `csprojInSubfolder = 'true'`
and it would be better to move this line after 
```
await Promise.all(subfolders.map(async folder => {
  // ...
});
if(projectFiles.length>0){
  context.telemetry.properties.csprojInSubfolder = 'true';
}
```

src/commands/deploy/dotnet/tryGetCsprojFile.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export async function tryGetCsprojFile(context: IDeployContext, projectPath: string): Promise<string | undefined> {
+    const projectFiles: string[] = await checkFolderForCsproj(projectPath);
+    // it's a common pattern to have the .csproj file in a subfolder so check one level deeper
+    if (projectFiles.length === 0) {
+        const subfolders: string[] = await fse.readdir(projectPath);
+        await Promise.all(subfolders.map(async folder => {
+            const filePath: string = path.join(projectPath, folder);
+            // check its existence as this will check .vscode even if the project doesn't contain that folder
+            if (await fse.pathExists(filePath) && (await fse.stat(filePath)).isDirectory()) {
```
wangmingliang-ms, line 18: nit: you don't need to check `fse.pathExists(filePath)` again since it's read from `projectPath` at L14

PR #2087, opened 2021-06-23
Improve masking of potential user info

ejizba:
Built off of the work I did in the shared packages. See ui package [release v0.44.2](https://github.com/microsoft/vscode-azuretools/releases/tag/v0.44.2-ui) for more information

Also depends on https://github.com/microsoft/vscode-azuretools/pull/946

### DIFF
--- a/src/tree/CosmosDBTreeItem.ts
+++ b/src/tree/CosmosDBTreeItem.ts
@@ -24,6 +24,7 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
     public readonly parent!: SiteTreeItem;
     public readonly client: IAppSettingsClient;
     public cosmosDBExtension: vscode.Extension<AzureExtensionApiProvider | undefined> | undefined;
+    public suppressMaskLabel = true;
 
     private readonly _endpointSuffix: string = '_ENDPOINT';
     private readonly _keySuffix: string = '_MASTER_KEY';
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,8 +16,8 @@
                 "moment": "^2.24.0",
                 "open": "^8.0.4",
                 "portfinder": "^1.0.25",
-                "vscode-azureappservice": "^0.81.0",
-                "vscode-azureextensionui": "^0.44.0",
+                "vscode-azureappservice": "^0.81.1",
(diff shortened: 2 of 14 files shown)


src/tree/CosmosDBTreeItem.ts
```diff
@@ -24,6 +24,7 @@ export class CosmosDBTreeItem extends AzureParentTreeItem<ISiteTreeRoot> {
     public readonly parent!: SiteTreeItem;
     public readonly client: IAppSettingsClient;
     public cosmosDBExtension: vscode.Extension<AzureExtensionApiProvider | undefined> | undefined;
+    public suppressMaskLabel = true;
```
neelip, line 27: Doesn't the `client` properties above require to be added to `valuesToMask`? 
Something similar to...
`const valuesToMask = [ client.siteName, client.slotName, client.defaultHostName, client.resourceGroup,
            client.planName, client.planResourceGroup, client.kuduHostName, client.gitUrl,
            site.repositorySiteName, ...(site.hostNames || []), ...(site.enabledHostNames || [])
        ];`

neelip (approved):

src/tree/CosmosDBTreeItem.ts
ejizba: The app settings client? I don't think so. Since this tree item is a child of the SiteTreeItem, it will inherit all of the same valuesToMask. See here where I loop up through the tree based on `parent`: https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/tree/addTreeItemValuesToMask.ts#L17

If you have specific properties in mind that you think I should add, though, let me know

neelip: Never mind, I was thinking more of the appSettings keys and values. But I found that they are taken care of in the appservice package.

ejizba: I'm actually not sure the stuff in the appservice package will cover every case, but I also can't do the same thing I'm doing with the `SiteClient` example you mentioned. `SiteClient` already has a bunch of values we can mask, but `IAppSettingsClient` requires a separate async call to list the app settings. I may investigate app settings stuff more, but for now I think this PR is a big step in the right direction so I'll go ahead and merge as-is.

PR #2097, opened 2021-07-21
allow user to deploy a maven module (which has been added as a root folder) to Azure App Service directly

wangmingliang-ms:


src/commands/deploy/java/tryGetMavenModule.ts
```diff
@@ -0,0 +1,49 @@
...
+export async function tryGetMavenModule(context: IDeployContext, projectPath: string): Promise<MavenModule | undefined> {
+    const pomFile = path.join(projectPath, 'pom.xml');
+    if ((await fse.pathExists(projectPath)) && (await fse.lstat(projectPath)).isDirectory() && (await fse.pathExists(pomFile))) {
+        return getMavenModuleFromPom(pomFile);
+    }
+    return undefined;
+}
+
+async function getMavenModuleFromPom(pomFile: string): Promise<MavenModule | undefined> {
+    const pomContent = await fse.readFile(pomFile, 'utf8');
```
ejizba, line 31: Do we need to specify encoding? If you leave out `utf8` I think `readFile` will figure it out for us.

ejizba (approved):

src/commands/deploy/java/tryGetMavenModule.ts
wangmingliang-ms: the type of returned result is `Promise<String>` if `encoding` is specified, and `Promise<Buffer>` otherwise. and I want `string` here. 
```
// fs-extra/index.d.ts

export function readFile(file: string | Buffer | number, encoding: string): Promise<string>;
export function readFile(file: string | Buffer | number): Promise<Buffer>;
```
and it's same as `fs.readFile`
https://nodejs.org/api/fs.html#fs_fspromises_readfile_path_options
image

ejizba: Yeah we normally just do `.toString()` on the buffer

wangmingliang-ms: that also makes sense, I added another commit to use `Buffer.toString()` instead.

PR #2098, opened 2021-07-23
Support correlation id

ejizba:
Depends on https://github.com/microsoft/vscode-azuretools/pull/963 and https://github.com/microsoft/vscode-azuretools/pull/964

See those PRs for more info.

### DIFF
--- a/src/tree/SiteTreeItem.ts
+++ b/src/tree/SiteTreeItem.ts
@@ -4,8 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementModels } from '@azure/arm-appservice';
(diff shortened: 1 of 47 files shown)


src/tree/SiteTreeItem.ts
```diff
@@ -51,62 +48,52 @@ export abstract class SiteTreeItem extends SiteTreeItemBase implements ISiteTree
...
     public get defaultHostUrl(): string {
-        return this.root.client.defaultHostUrl;
+        return this.site.defaultHostUrl;
     }
 
     public get defaultHostName(): string {
-        return this.root.client.defaultHostName;
-    }
-
-    public async openInPortal(): Promise<void> {
```
neelip, line 66: why is openInPortal() removed?

neelip (approved):

src/tree/SiteTreeItem.ts
ejizba: This is an old pattern and it's not used anymore. The other place that used to do it was `AzureTreeItem`, which was deleted during the ui package correlation id PR: https://github.com/microsoft/vscode-azuretools/pull/963/files#diff-e2ec57a2fe686027c0075dfb05b4b485c69b829e08ced1aa205cf6135dcef18cL487

PR #2131, opened 2021-09-03
Fix sshToWeb issue with password input using Proposed API

### LINKED ISSUE #932: Use CustomExecution to run SSH Task instead of Terminal
When this is in stable https://github.com/Microsoft/vscode-docs/blob/vnext/release-notes/v1_33.md#custom-task-execution, should use it instead of `vscode.Terminal`

neelip:
Fixes #932

App Service is now approved and added to the vscode-proposedapi registry in vscode@1.60.0. This will allow us to use "onDidWriteTerminalData" property to fix this particular problem. 

In our meeting with @Tyriar, we have confirmed that the "onDidWriteTerminalData" is stable to use in the dev mode.


src/constants.ts
```diff
@@ -37,3 +37,5 @@ export const envFileName: string = '.env';
 export const detectorTimestampFormat: string = 'YYYY-MM-DDTHH:mm';
 
 export const webProvider: string = 'Microsoft.Web';
+
+export const sshURL: string = 'root@127.0.0.1';
```
alexweininger, line 41: Since this is only used in startSsh.ts you could declare it in startSsh.ts instead of exporting it here.

package.json
```diff
@@ -816,10 +816,12 @@
         "vscode-azureappservice": "^0.86.2",
         "vscode-azureextensionui": "^0.48.5",
         "vscode-azurekudu": "^0.2.0",
+        "vscode-dts": "^0.3.1",
```
alexweininger, line 819: If I understand what this is used for, I think this should be a dev dependency.

For reference I found another Microsoft extension include it as a dev dependency:
https://github.com/microsoft/vscode-simple-jupyter-notebook/blob/77cd8b0f1695e2228756f9ace1d8e2b4d4d21c6c/package.json#L122

alexweininger (commented):
Nice! I tested myself and connecting via SSH worked perfectly.

package.json
neelip: Yes, that makes more sense

Tyriar (changes requested):

Tyriar (approved):

alexweininger (commented):
Did we want to include the entire `vscode.proposed.d.ts` file or just the section we need?

Tyriar:
Related discussion: https://github.com/microsoft/vscode/issues/131165

Including just the section you need would be a good idea if that works.

alexweininger (approved):

PR #2139, opened 2021-09-23
Fix .NET6 nightly test project

v-wuzhai:


nturinski:
@v-wuzhai It seems like the .NET 6 tests are still broken even after these changes.

PR #2169, opened 2022-01-18
Update follow-redirects to 1.14.7

alexweininger:
Fixes CVE-2022-0155

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,7 @@
                 "webpack-cli": "^4.6.0"
             },
             "engines": {
-                "vscode": "^1.60.0"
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -49,7 +49,7 @@
                 "webpack-cli": "^4.6.0"
             },
             "engines": {
-                "vscode": "^1.60.0"
+                "vscode": "^1.63.0"
```
alexweininger, line 52: Was worried about this but it matches https://github.com/microsoft/vscode-azureappservice/blob/main/package.json#L10 so I think it's ok.

wwlorey (approved):

PR #2177, opened 2022-02-22
Add new proposed api

nturinski:
Based from release 7539cf687cab06110880cbff2532da3d6a5718ff.  Just adding the proposed API changes on top of it.


bwateratmsft (approved):
I wonder if we should establish a cleanup plan for removing these declaration files when they are no longer needed.

nturinski:
This one has been in proposed for like over 3 years, so not sure it's ever leaving 😬

bwateratmsft:
!remindme 10 years

:laugh:

PR #2178, opened 2022-02-22
Update appservice package to fix asp bug

nturinski:


bwateratmsft (approved):
I take it that it did indeed look like the same issue?

nturinski:
Yep, looks like it did it!

PR #2236, opened 2022-04-27
Fix connected deployments disconnect from repo command

### LINKED ISSUE #2228: An error occurs when executing "Disconnect from Repo..." action for "Deployment" node
**OS:** All
**Azure Resource Extension Build Version:** 20220421.32
**App Service Extension Build Version:** 20220421.6
**Regression:** Not a regression

**Repro Steps:**
1. Create a Linux web app.
2. Expand this web app -> "Deployments" node -> Click "Connect to a GitHub Repository..." to connect a repository.
3. Right click "Deployments" node and execute "Disconnect from Repo..." action.
4. Check whether succeeds to execute "Disconnect from Repo..." action or not.

**Expect:**
Succeed to execute "Disconnect from Repo..." action.

**Actual:**
An error occurs.
image

alexweininger:
Fixes #2228

### DIFF
--- a/src/tree/ResolvedWebAppResource.ts
+++ b/src/tree/ResolvedWebAppResource.ts
@@ -22,14 +22,14 @@ import { SiteTreeItem } from './SiteTreeItem';
 import { WebJobsNATreeItem, WebJobsTreeItem } from './WebJobsTreeItem';
 
 export function isResolvedWebAppResource(ti: unknown): ti is ResolvedWebAppResource {
-    return (ti as unknown as ResolvedWebAppResource).kind === ResolvedWebAppResource.kind;
+    return (ti as unknown as ResolvedWebAppResource).instance === ResolvedWebAppResource.instance;
 }
 
 export class ResolvedWebAppResource implements ResolvedAppResourceBase, ISiteTreeItem {
     public site: ParsedSite;
 
-    public static kind = 'resolvedWebApp';
-    public readonly kind = ResolvedWebAppResource.kind;
+    public static instance = 'resolvedWebApp';
+    public readonly instance = ResolvedWebAppResource.instance;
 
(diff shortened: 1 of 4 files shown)


src/tree/ResolvedWebAppResource.ts
```diff
@@ -21,9 +21,16 @@ import { NotAvailableTreeItem } from './NotAvailableTreeItem';
 import { SiteTreeItem } from './SiteTreeItem';
 import { WebJobsNATreeItem, WebJobsTreeItem } from './WebJobsTreeItem';
 
+export function isResolvedWebAppResource(ti: unknown): ti is ResolvedWebAppResource {
+    return (ti as unknown as ResolvedWebAppResource).kind === ResolvedWebAppResource.kind;
```
nturinski, line 25: Why check `kind`? I think that for App Service, the `kind` is always going to be blank so this will probably always be true.  If we're just making sure it's a `SiteTreeItem`, we should be able to check if the context matches with `azApp`

nturinski: Or we could check `type`?

alexweininger: Pardon the naming, but this is a _new_ `kind`.

https://github.com/microsoft/vscode-azureappservice/pull/2236/files#diff-33dbbf31b6cdf6477373c8a791073499ee49d90cd7abbd5cc44d4e2ad4c6af4fR31-R32

nturinski: Not sure if you saw my previous comment (because I did an edit) so bump.

alexweininger: Got any ideas for a better idea?

nturinski (approved):

PR #2352, opened 2023-01-17
Respond to AzExtTreeFileSystem changes

alexweininger:
Picking up changes made to `AzExtTreeFileSystem` in https://github.com/microsoft/vscode-azuretools/pull/1257


package.json
```diff
@@ -794,7 +794,7 @@
         "@azure/ms-rest-js": "^2.2.1",
         "@microsoft/vscode-azext-azureappservice": "^0.7.4",
         "@microsoft/vscode-azext-azureutils": "^0.3.4",
```
bwateratmsft, line 796: We should update the other two packages from `vscode-azuretools` at the same time. Otherwise, you'll have a mix of 0.3.7 and 0.4.0 of `@microsoft/vscode-azext-utils`.

src/commands/showFile.ts
```diff
@@ -4,17 +4,20 @@
...
     context.telemetry.eventVersion = 2;
 
     ext.fileSystem.appendLineToOutput(localize('opening', 'Opening "{0}"...', node.label), { resourceName: node.site.fullName });
     if (node.isReadOnly) {
         await node.openReadOnly(context);
     } else {
-        await ext.fileSystem.showTextDocument(node);
+        // ensure node.id is defined
+        node.id = node.fullId;
+        await ext.fileSystem.showTextDocument(node as FileTreeItem & { id: string });
```
bwateratmsft, line 21: Could export and reuse `FileSystemItem` for this `as ...`

package.json
alexweininger: I will hold off on updating app service since I don't want to incorporate the kudu changes yet.

bwateratmsft: Ok. Is that something we want to do before release?

alexweininger: Probably not for the v1.5 release. Unless it's super simple. I'll talk to Nathan about it.

bwateratmsft (approved):

PR #2354, opened 2023-01-25
Check `scmType` on an existing `siteConfig` comparison

MicroFish91:
I'm assuming this is what was originally meant to be checked


bwateratmsft (approved):

alexweininger:
Does this fix #2355?

MicroFish91:
That doesn't completely fix this specific issue unfortunately.

PR #2511, opened 2023-06-09
Add Service Connector

motm32:
Related to https://github.com/microsoft/vscode-azuretools/issues/1419

Also includes a utils update (needed for Service Connector) and some changes to fix errors that came with the update.


alexweininger:
> Also includes a utils update (needed for Service Connector) and some changes to fix errors that came with the update.

If it's not too much work, it would be great to just split those out into a new PR. We will be able to merge that sooner than the service connector changes.

motm32:
> If it's not too much work, it would be great to just split those out into a new PR. We will be able to merge that sooner than the service connector changes.

Will do!

src/commands/registerCommands.ts
```diff
@@ -56,9 +59,13 @@ export function registerCommands(): void {
     registerCommandWithTreeNodeUnwrapping('appService.appSettings.ToggleSlotSetting', toggleSlotSetting);
     registerCommandWithTreeNodeUnwrapping('appService.appSettings.Upload', uploadAppSettings);
     registerCommandWithTreeNodeUnwrapping('appService.Browse', browseWebsite);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    //@ts-ignore
```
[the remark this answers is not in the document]
bwateratmsft: Ah. Let's get https://github.com/microsoft/vscode-azuretools/pull/1492 fixed instead.

alexweininger:
Looks good to go, just need to release the package

motm32:
The previous commit includes hooking up the UI and some changes made in the tools package to make the UI work correctly. I am not totally sure if this is the best approach so any feedback would be great.

src/commands/serviceConnector/createServiceConnector.ts
```diff
@@ -3,25 +3,28 @@
...
 import { localize } from "../../localize";
 import { SiteTreeItem } from "../../tree/SiteTreeItem";
 import { createActivityContext } from "../../utils/activityUtils";
 import { pickWebApp } from "../../utils/pickWebApp";
 
-export async function createServiceConnector(context: IActionContext, item?: SiteTreeItem): Promise<void> {
+export async function createServiceConnector(context: IActionContext, item?: SiteTreeItem | ServiceConnectorGroupTreeItem): Promise<void> {
     item ??= await pickWebApp(context);
+    if (item instanceof SiteTreeItem) {
+        item = <ServiceConnectorGroupTreeItem>item.parent;
```
nturinski, line 16: This seems backwards to me.

If it was a SiteTreeItem, would its parent be the subscription node?
But the ServiceConnectorGroupTreeItem's parent would be the SiteTreeItem, right?

motm32: That is true it wasn't causing any problems for some reason since the type of item when using the command palette was not a SiteTreeItem so it was never calling that line. 

I'll flip it but with that implementation I don't thing resourceId will be needed anymore.

nturinski: Yeah, I believe that you're right. I know I just approved it, but we could probably disregard the `resourceId` added in this [PR](https://github.com/microsoft/vscode-azuretools/pull/1539#pullrequestreview-1559486372)

nturinski (approved):

package.nls.json
```diff
@@ -62,6 +64,7 @@
     "appService.showSavePrompt": "Show warning dialog on remote file uploading.",
     "appService.startStreamingLogs": "Start Streaming Logs",
     "appService.toggleAppSettingVisibility": "Toggle App Value Visibility",
+    "appService.validateServiceConnector": "Validate Service Connector...",
```
alexweininger, line 67: Since there are no additional prompts, the ellipses should be removed.

image

src/commands/serviceConnector/validateServiceConnector.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    if (item instanceof ServiceConnectorTreeItem) {
+        serviceConnectorName = item.label;
+        item = <SiteTreeItem>item.parent?.parent;
+    }
+
+    const activityContext = {
+        ...context,
+        ...await createActivityContext(),
+        activityTitle: localize('validate', `Validate {0}`, serviceConnectorName),
```
alexweininger, line 25: Last thing I promise

```suggestion
        activityTitle: localize('validate', `Validate connection "{0}"`, serviceConnectorName),
```

alexweininger (approved):

PR #2541, opened 2023-07-17
Fix create web app from workspace

alexweininger:


nturinski (approved):
Oof, these are some nasty workarounds. I'm okay with doing this, but we should have a cleaner, general fix for this.

I was thinking maybe we could create a sort of wrapper command. Like, `createWebAppInternal` and then `createWebAppWorkspace` and `createWebAppResources` or something that will know exactly what types of nodes it will have to handle, and then pass that into `createWebAppInternal`?

bwateratmsft (approved):

PR #2619, opened 2024-02-09
Create workspace based app insights

nturinski:
Leverages log analytics create step, which is required to migrate App Insights to the Workspace Based. Here is some [documentation](https://learn.microsoft.com/en-us/azure/azure-monitor/app/convert-classic-resource#migrate-your-resource) on that. This was already being done in Functions, so the step already existed.

It also uses the connection string property rather than the instrumentational key since that is the recommended path going forward.

Another reference: https://learn.microsoft.com/en-us/azure/azure-monitor/app/create-workspace-resource

How to test:
1. Right-click "App Service" node and select "Create Web App..."
2. Fill out the prompts
3. Check to see if a log analytics workspace is created and linked to app insights
4. Check in the portal for the app insight component and make sure it's workspace-based (it'll tell you that you need to migrate it)
5. Try it with advanced too


MicroFish91:
I'm running with advanced because there are no resources currently available to test on basic.

It seems like I'm never prompted about creating a log analytics workspace, but the output logs indicate it's being defaulted to an existing resource.  Is this intended behavior?

image

nturinski:
> (I'm running with advanced because there are no resources currently available to test on basic.)

What's the error? If you increase the sku, that will usually resolve the issue.

> It seems like I'm never prompted about creating a log analytics workspace, but the output logs indicate it's being defaulted to an existing resource. Is this intended behavior?

Yeah, there's nothing to really pick and as far as I know, no benefit to really having a bunch of workspaces. I'll confirm with the App Service team, but that is the intended behavior.  You must have had one created with a function app at some point.

MicroFish91:
> What's the error?  If you increase the sku, that will usually resolve the issue.

image

MicroFish91 (approved):
Checked Advanced and compared resources against the docs, so LGTM.  I'll double check Basic if it becomes available, but approving so as not to block you.

nturinski:
Are you using the free tier?

MicroFish91:
It reproduces on all the tiers offered via quick picks

PR #2621, opened 2024-02-09
Allow users to deploy zip files directly

nturinski:
Tested it with a zip file and it seems like it works perfectly fine.

How to test:
1. Clone a sample project (this one works fine https://github.com/johnpapa/node-hello)
2. Zip its contents (make sure that you don't zip the folder itself)
3. Right-click and deploy


alexweininger (dismissed):
We should add a test case for this

MicroFish91 (approved):

nturinski:
Can't actually test this right now due to tests being skipped, but I _think_ that this should work

alexweininger (approved):

PR #2648, opened 2024-06-21
Fix the nightly tests

MicroFish91:
WIP: Got everything hooked up, but some of the inputs don't match up with the quick picks in each implementation.  I assume we may have made some updates while the tests were disabled, and thus never updated the tests to reflect these changes.  I'll spend some time over the next week working on fixing all the tests. 🫡


MicroFish91:
Will reopen later with finalized split PRs

PR #2653, opened 2024-07-09
Enable nightly tests and first pass of fixes for `deploy.test.ts`

MicroFish91:
Depends on: https://github.com/microsoft/vscode-azuretools/pull/1756

You can disregard any changes to `azureResourceWebAppActions` as I will be posting all the fixes for that test file in the next PR.  The `suite.only` being used for now is also intentional and will be removed in the next PR as well.

Local:
image

Remote:
image
https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=9947761&view=results

Fixes for the following will come in follow-up PRs:
- zipfile
- .NET (needs further investigation)
- the latter half of deploy logic (needs further investigation)
- `azureResourceWebAppActions.test.ts`


nturinski:
I'm gonna hold off on approving until the appservice package gets released so we can observe what the long running tests do.

MicroFish91:
Confirmed everything is working, ready for final review :)

package.json
```diff
@@ -158,11 +153,6 @@
                 "title": "%appService.EnableFileLogging%",
                 "category": "Azure App Service"
             },
-            {
-                "command": "appService.InstallCosmosDBExtension",
```
nturinski: I don't think we want to include removing all the cosmos connection stuff in this PR as well.

src/commands/createWebApp/createWebApp.ts
```diff
@@ -30,7 +30,7 @@ function isSubscription(item?: AzExtParentTreeItem): boolean {
 
 export async function createWebApp(context: IActionContext & Partial<ICreateChildImplContext>, node?: SubscriptionTreeItemBase | undefined, _nodes?: (SubscriptionTreeItemBase | undefined)[], suppressCreatedWebAppMessage: boolean = false): Promise<SiteTreeItem> {
     if (!node || !isSubscription(node)) {
-        node = <SubscriptionTreeItemBase>await ext.rgApi.tree.showTreeItemPicker(SubscriptionTreeItemBase.contextValue, context);
+        node = <SubscriptionTreeItemBase>await ext.rgApi.tree.showTreeItemPicker(new RegExp(SubscriptionTreeItemBase.contextValue, 'i'), context);
```
nturinski, line 33: Can you explain why you had to change this to a regex?

test/global.test.ts
```diff
@@ -16,7 +17,10 @@ export const longRunningTestsEnabled: boolean = longRunningLocalTestsEnabled ||
 suiteSetup(async function (this: Mocha.Context): Promise<void> {
     this.timeout(120 * 1000);
     await vscode.commands.executeCommand('appService.Refresh'); // activate the extension before tests begin
+
     ext.outputChannel = new TestOutputChannel();
+    registerUIExtensionVariables(ext);
```
nturinski, line 22: This should have happened automatically when you activate (refresh) the extension. Why is this being invoked directly here?

test/nightly/deploy.test.ts
```diff
@@ -44,42 +41,45 @@ interface IParallelTest {
     callback(): Promise<void>;
 }
 
-suite('Create Web App and deploy', function (this: Mocha.Suite): void {
+suite.only('Create Web App and deploy', function (this: Mocha.Suite): void {
```
nturinski, line 44: I figure that this was for testing purposes, but you should remove the `.only` from the suite now.

test/nightly/deploy.test.ts
```diff
@@ -44,42 +41,45 @@ interface IParallelTest {
...
+        //     ]
+        // },
         {
             runtimePrefix: 'Python',
             workspaceFolder: 'python-docs-hello-world',
             versions: [
-                { version: '3.7', supportedAppOs: 'Linux' },
                 { version: '3.8', supportedAppOs: 'Linux' },
-                { version: '3.9', supportedAppOs: 'Linux' }
+                { version: '3.9', supportedAppOs: 'Linux' },
```
nturinski, line 79: Are any of these versions being deprecated? We probably don't need to test 5 different versions.

test/nightly/deploy.test.ts
```diff
@@ -120,41 +120,50 @@ suite('Create Web App and deploy', function (this: Mocha.Suite): void {
...
-    async function testCreateWebAppAndDeploy(os: string, promptForOs: boolean, runtime: string, workspacePath: string, expectedVersion: string, zipFile?: string): Promise<void> {
+    async function testCreateWebAppAndDeploy(os: string, promptForOs: boolean, runtime: string, _workspacePath: string, _expectedVersion: string, _zipFile?: string): Promise<void> {
         const resourceName: string = getRandomHexString();
-        const resourceGroupName = getRandomHexString();
-        resourceGroupsToDelete.push(resourceGroupName);
+        const resourceGroupName: string = azcodeResourcePrefix + getRandomHexString(6);
+        resourceGroupsToDelete.add(resourceGroupName);
 
-        const testInputs: (string | RegExp)[] = [resourceName, '$(plus) Create new resource group', resourceGroupName, runtime];
+        const testInputs: (string | RegExp)[] = [/Manual\sTest/i, resourceName, '$(plus) Create new resource group', resourceGroupName, runtime];
```
nturinski, line 128: With the service principal, there should only be one subscription that can be selected which results in an auto-selection. I think that this isn't being checked right now since you have the deploy tests commented out, but this probably won't work.

test/nightly/global.nightly.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import { type AzureSubscription } from '@microsoft/vscode-azureresources-api';
+import * as vscode from 'vscode';
+import { createSubscriptionContext, createWebSiteClient, ext, subscriptionExperience, type ISubscriptionContext } from '../../extension.bundle';
+import { longRunningTestsEnabled } from '../global.test';
+
+export let subscriptionContext: ISubscriptionContext;
+export let webSiteClient: WebSiteManagementClient;
+
+export const azcodeResourcePrefix: string = 'azc-app';
+export const resourceGroupsToDelete = new Set<string>();
```
nturinski, line 19: I don't really mind, but why did you decide to go with a set over just using an array? Feels like it just adds more overhead since you convert it into an array in the end anyway.

test/global.test.ts
MicroFish91: Refresh wasn't accomplishing that it seems and so was causing an error to be thrown.  I've added an activate block instead which we can use instead of registering these manually.

test/nightly/deploy.test.ts
MicroFish91: See original note, but I planned to keep this in until I add in the follow-up PR if that's okay.

test/nightly/deploy.test.ts
MicroFish91: I'm not sure about future deprecations, but these are all currently being offered in our create experience, so I figured I'd add them to match how it seems we were testing.  Are there any you'd prefer I remove?

test/nightly/deploy.test.ts
MicroFish91: > With the service principal, there should only be one subscription that can be selected which results in an auto-selection.

This is what I assumed as well, however, it seems that the auto-selection wasn't working even when there was only one subscription in the filter.  A better long term solution probably needs to be to investigate why the subscription isn't being automatically selected.

> I think that this isn't being checked right now since you have the deploy tests commented out, but this probably won't work.

I think it should be working-- I checked remotely and everything was passing.  And one of the test assertions is this:
```typescript
const createdApp: Site | undefined = await tryGetWebApp(webSiteClient, resourceGroupName, resourceName);
assert.ok(createdApp);
```

test/nightly/global.nightly.test.ts
MicroFish91: Probably a preference thing, but I don't like the possibility of having duplicate resource groups listed in the array when each one only needs one delete command issued.  

My initial impression was that it complicates the delete logic later because we need to then check to see if we've already deleted the resource group since the possibility of duplicates exist when we don't use a set.  Let me know if you prefer it the other way and I can change it back though :P

package.json
MicroFish91: Oh maybe I forgot to merge with main

src/commands/createWebApp/createWebApp.ts
MicroFish91: This line was hanging during tests causing fails, my hunch is that the context value of the sub has changed over time and may have new values being appended to it.  Using a regexp here allows us to still find the match without it being a direct 1:1.  I cannot confirm for sure without digging in and adding breakpoints into the picker logic, however, I think this is very likely the case seeing that this fixed the hanging issue.

MicroFish91: This was a workaround for an issue I was running into during parallel tests.  I thought I had found the solution, but upon closer investigation I believe a more robust solution is needed.  Will contact you to talk more about this offline.

test/nightly/deploy.test.ts
MicroFish91: Reinvestigated this and think we will need to talk more about this offline, see this [comment](https://github.com/microsoft/vscode-azureappservice/pull/2653/files#r1696108931).

test/global.test.ts
MicroFish91: Actually both ways still cause the error to reproduce:
image

Can talk more offline.

MicroFish91:
It's been a while, will revisit later and reopen when ready.

PR #2708, opened 2025-03-06
Use node 20, require vscode 1.95

alexweininger:


bwateratmsft (commented):
We should update the required VSCode engine version at the same time.

bwateratmsft (approved):

PR #2723, opened 2025-04-01
Add `domainNameLabelScope` and `siteLocation` to deployment telemetry

MicroFish91:


src/commands/deploy/deploy.ts
```diff
@@ -63,6 +64,8 @@ export async function deploy(actionContext: IActionContext, arg1?: vscode.Uri |
 
     const correlationId: string = getRandomHexString();
     context.telemetry.properties.correlationId = correlationId;
+    context.telemetry.properties.siteLocation = node.site.location;
+    context.telemetry.properties.domainNameLabelScope = await getDomainNameLabelScope(Object.assign(context, node.subscription), node.site.resourceGroup, node.site.siteName);
```
nturinski, line 68: Can you wrap in a try/catch? I don't want deployment to fail if this GET call fails.

nturinski (approved):

src/commands/deploy/deploy.ts
MicroFish91: Oh there's a try/catch in the func itself

Issue #2735, opened 2025-04-25
Add activity children to P0 and P1 missing commands, and migrate to v2 for exisiting


motm32:
### Command List

- [ ] Creating Web Apps
- [ ] Deploy to Web App
- [ ] Deploy to Slot

PR #2736, opened 2025-04-25
Add new activity children and convert existing to v2

motm32:
Refer to #2735 for which commands are being converted/added.

Notes:
- [x] need to add logic for deploy 
- [x] update packages once `ResourceGroupCreateStep` `AppInsightsCreateStep` and `AppServiceCreateStep` have proper changes released


src/commands/StartingResourcesLogStep.ts
```diff
@@ -0,0 +1,85 @@
...
+
+type StartingResourcesLogContext = IActionContext & Partial<ExecuteActivityContext> & ILocationWizardContext & {
+    resourceGroup?: ResourceGroup,
+    plan?: AppServicePlan,
+    appInsightsComponent?: ApplicationInsightsComponent,
+};
+
+const startingResourcesContext: string = 'startingResourcesLogStepItem';
+
+export class StartingResourcesLogStep<T extends StartingResourcesLogContext> extends AzureWizardPromptStep<T> {
```
nturinski, line 23: This seems like something that belongs in the azure package. I made a similar comment to Matt in his Storage PR as well.

It almost feels like rather than being a step that's injected, it's something that can happen in the wizard itself-- maybe we have a list of keys or something that it checks for before starting to execute steps.

src/utils/activityUtils.ts
```diff
@@ -3,13 +3,35 @@
...
+        activityChildren: options?.withChildren ? [] : undefined,
     };
 }
+
+/**
+ * Adds a new activity child after the last info child in the `activityChildren` array.
+ * If no info child already exists, the new child is prepended to the front of the array.
+ * (This utility function is useful for keeping the info children grouped at the front of the list)
+ */
+export function prependOrInsertAfterLastInfoChild(context: Partial<ExecuteActivityContext>, infoChild: ActivityChildItemBase): void {
```
nturinski, line 23: I think we should make a universal `activityUtils` in our package by this point.

src/commands/StartingResourcesLogStep.ts
motm32: Yah i just followed the conventions set from the Storage PR. I agree that it makes more sense for it to happens in the wizard itself. Thoughts @MicroFish91?

MicroFish91: Yeah that could work.  What if we ship this way for now and see what each client extension needs and then investigate replacing with a more generic shared solution after build?

src/utils/activityUtils.ts
MicroFish91: Agreed, I was putting off adding this one because I wasn't sure how often it would be used

MicroFish91 (approved):
I didn't actually test run any commands but I assume you've got that covered.  The code itself LGTM though!

PR #2743, opened 2025-05-21
Remove duplicate "Created.." messages in Output window 

### LINKED ISSUE #2742: There are two "Created.." messages in Output window after creating a web app
**OS:** All
**Build Version:** 20250516.1
**Regression:** Previous Release

**Repro Steps:**
1. Create a web app in VS Code.
2. Check whether there is no duplicate messages in Output window.

**Expect:**
There is no duplicate messages in Output window.

**Actual:**
There are two `Created..` messages in Output window.
Image

**More Info:**
1. This issue does not reproduce on the current released extension v0.26.1.
Image
2. There is no period in `Created..` messages in the output window after creating a web app.

motm32:
Fixes #2742.


nturinski (commented):
Could you mention in your PR description that the change to showSlotMessage is due to the fact that the slot isn't displayed in the activity log? I can see myself not remembering that at some point.

nturinski (approved):

PR #2751, opened 2025-07-02
Update appservice package for deploy log copilot and deployment tree item perf enhancements

nturinski:


nturinski:
Yeah, it just upgrades appservice to include deployment logs. I still need to actually build out the metadata in the deploy command in the extension itself though.

Issue #2754, opened 2025-07-11
There is no "Azure" activity log when creating a slot for a web app


v-ruizh:
**OS:** All
**Build Version:** 20250704.1
**Regression:** Not a Regression

**Repro Steps:**
1. Create a web app.
2. Create a slot for the above web app.
3. Check whether there is an "Azure" activity log when creating a slot.

**Expect:**
There is an "Azure" activity log when creating a slot.

**Actual:**
There is no "Azure" activity log when creating a slot.

**More Info:**
There is an "Azure" activity log when deleting a slot.
<img width="511" height="57" alt="Image" src="https://github.com/user-attachments/assets/6d9a84b4-f375-4990-a29f-48f0735ca23c" />

Issue #2753, opened 2025-07-11
An error occurs when executing "Deploy to Slot..." command


v-ruizh:
**OS:** All
**Build Version:** 20250704.1

**Repro Steps:**
1. Create a slot for one web app.
2. Right click the slot -> Execute "Deploy to Slot..." action.
3. Check whether succeeds to execute "Deploy to Slot..." action.

**Expect:**
Succeed to execute "Deploy to Slot..." action without any error.

**Actual:**
An error occurs when executing "Deploy to Slot..." action.
<img width="450" height="74" alt="Image" src="https://github.com/user-attachments/assets/af08e9ee-71c1-45a9-aa47-4001eeae4caf" />

**More Info:**
This issue also reproduces when executing "Deploy to Slot..." command from Command Palette.

v-ruizh:
This bug has been fixed on 20250716.2.

PR #2756, opened 2025-07-11
Fixes for deployment slots

nturinski:
Fixes https://github.com/microsoft/vscode-azureappservice/issues/2753
Fixes https://github.com/microsoft/vscode-azureappservice/issues/2754


nturinski:
Note, the create slot looks pretty janky because it populates before the prompt (since there are no prompt steps in createSlot) but I figured that it happens rarely enough that we can just deal with it for now.

alexweininger (approved):

PR #2811, opened 2026-02-03
Update utils to get new telemetry changes

motm32:


bwateratmsft (commented):
Something still seems off; now all the optional esbuild packages are gone (except the one you installed with on Darwin ARM64). Same for VSCE stuff...

PR #2888, opened 2026-06-08
Audit cross-origin redirect usage after @azure/core-rest-pipeline 1.23.0 breaking change

### LINKED ISSUE #2877: Audit cross-origin redirect usage after @azure/core-rest-pipeline 1.23.0 breaking change
`@azure/core-rest-pipeline@1.23.0` ([Azure/azure-sdk-for-js#37384](https://github.com/Azure/azure-sdk-for-js/pull/37384)) restricts the default `redirectPolicy` to same-origin redirects. Cross-origin redirects (different scheme/host/port) are now dropped unless callers opt in via `redirectOptions: { allowCrossOriginRedirects: true }`.

This repo currently pins `@azure/core-rest-pipeline` at `^1.11.0`, so it is **not yet broken**, but will break the moment the transitive/direct dep resolves >= 1.23.0.

### Call sites to audit
Highest risk (likely cross-origin redirects):
- `src/commands/createWebApp/stacks/getStackPicks.ts` — same `aka.ms` → NuGet stacks pattern that is already confirmed broken in vscode-azurefunctions
- `src/commands/postDeploy/validateWebSite.ts` — pings deployed `*.azurewebsites.net`

ARM-only (lower risk, worth a sanity check):
- `src/commands/postDeploy/getLinuxDetectorError.ts`
- `src/commands/deployImage/CreateWebhookStep.ts`

Test-only:
- `test/nightly/deploy.test.ts`

### Suggested fix
Pass `redirectOptions: { allowCrossOriginRedirects: true }` when constructing the request/client for any call that is known to redirect cross-origin. This likely requires plumbing the option through `createGenericClient` in `@microsoft/vscode-azext-azureutils` (`vscode-azuretools/azure`).

cc @bwateratmsft

bwateratmsft:
Fixes #2877

## Background

`@azure/core-rest-pipeline@1.23.0` ([Azure/azure-sdk-for-js#37384](https://github.com/Azure/azure-sdk-for-js/pull/37384)) restricts the default redirect policy to **same-origin** redirects. Cross-origin redirects (different scheme/host/port) are now dropped unless the caller opts in via `redirectOptions: { allowCrossOriginRedirects: true }`. This repo pinned `@azure/core-rest-pipeline` at `^1.11.0`, so it was not yet broken — but would break the moment the dependency resolved `>= 1.23.0`.

## Mechanism

The opt-in is applied at **client construction** (the redirect policy reads the option when the pipeline is built), and `createGenericClient`'s 3rd `options` argument already carries `redirectOptions` — so the fix lives entirely in this repo, no upstream `vscode-azuretools` plumbing required. Two dependency bumps were needed to make it effective:

- **`@azure/core-rest-pipeline` `^1.11.0` → `^1.23.0`** — the `allowCrossOriginRedirects` property only exists on `RedirectPolicyOptions` in `>= 1.23.0`; this also makes the new same-origin-only behavior active so the fix is exercised rather than latent.
- **`@microsoft/vscode-azext-azureutils` `^4.1.1` → `^4.2.0`** — `4.1.1` neither typed `IGenericClientOptions` to accept `redirectOptions` nor forwarded `options` into the `ServiceClient` at runtime, so the option would have been silently dropped. `4.2.0` fixes both (`IGenericClientOptions extends CommonClientOptions` and `new ServiceClient({ ...options, ... })`).

## Changes

Added `redirectOptions: { allowCrossOriginRedirects: true }` to the `createGenericClient` calls at the genuinely cross-origin call sites:

- `src/commands/createWebApp/stacks/getStackPicks.ts` — the ARM stacks provider 302-redirects to a backend/CDN (the same pattern already confirmed broken in vscode-azurefunctions).
- `src/commands/postDeploy/validateWebSite.ts` — pings the deployed `*.azurewebsites.net` host, which commonly redirects cross-origin (Easy Auth / AAD, custom domain, http→https).
- `test/nightly/deploy.test.ts` — shares the `*.azurewebsites.net` ping pattern.

Left **unchanged** (plain ARM control-plane, no cross-origin redirect expected): `src/commands/postDeploy/getLinuxDetectorError.ts` and `src/commands/deployImage/CreateWebhookStep.ts`.

## Verification

- `npm run build` (esbuild + `tsc --noEmit`) ✅
- `npm run lint` (eslint `--max-warnings 0`) ✅


nturinski (approved):
I tried it out in the extension, the stacks load and the app creates without issue.

